PR #13, opened 2018-03-30
minor change copy/editing

shillasaebi:
changed it's to its 
it's = it is and didn't make sense


jranson:
How dare you. I have perfect grammar! ;-)

PR #14, opened 2018-03-30
update kubectl, minikube, helm to latest supported versions

LimitlessEarth:
Update:
Kubectl version to latest supported by minikube
Helm + tiller version to latest

Tested working locally


jranson:
LGTM

PR #15, opened 2018-03-30
Migrating from GoDep to dep

WillGunther:
This is just a suggestion. Based on the GoDep wiki. I noticed that even the Kubernetes project uses GoDep, and Prometheus appears to use govendor.
```markdown
Godep - Archived
Please use [dep](https://github.com/golang/dep) or another tool instead.

The rest of this readme is preserved for those that may still need its contents.
```
As we know, dep will also be depreciated in favor of  [vgo](https://github.com/golang/go/issues/24301)


crandles:
dep seems like a fine alternative to GoDep.

PR #32, opened 2018-04-10
Add code license header and increment to 0.0.12

jranson:


cache.go
```diff
@@ -1,5 +1,17 @@
 package main
 
+/**
```
crandles: can you move the comments above the package declaration? It will read better in godoc

PR #44, opened 2018-05-10
Expose more config params to be set through helm values.

ethanocentricity:
Templatized more of the config map so that users aren't restricted to default configurations.
Also upped the default docker image version and chart version.

### DIFF
--- a/deploy/helm/trickster/values.yaml
+++ b/deploy/helm/trickster/values.yaml
@@ -3,12 +3,13 @@
 # Default trickster originURL, references a source Prometheus instance
 # Ref: https://github.com/Comcast/trickster/blob/master/docs/configuring.md
 originURL: http://prometheus:9090
-cacheType: memory
-redisCache:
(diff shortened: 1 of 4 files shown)


deploy/helm/trickster/values.yaml
```diff
@@ -3,14 +3,28 @@
 # Default trickster originURL, references a source Prometheus instance
 # Ref: https://github.com/Comcast/trickster/blob/master/docs/configuring.md
 originURL: http://prometheus:9090
+cacheType: memory
```
crandles: I'm thinking this might be a better way to structure these cache related variables
```
cache:
  type: memory
  redis:
    protocol: tcp
    endpoint: redis:6379
  filesystem:
    path: /tmp/trickster
```

ethanocentricity: Sure, that's fine with me.  Let me change it quickly.

PR #47, opened 2018-06-04
fix wrong timeunit in trickster_proxy_duration_seconds

123Haynes:
The duration was measured in milliseconds, but the metric is for seconds.

closes https://github.com/Comcast/trickster/issues/37

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -246,7 +246,7 @@ func (t *TricksterHandler) getURL(method string, uri string, params url.Values, 
 		return []byte{}, resp, 0
 	}
 
-	duration := int64(time.Now().Sub(startTime).Nanoseconds() / 1000000)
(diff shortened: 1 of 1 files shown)


jranson:
@123Haynes I approved this, however it will need to be adjusted further. The value is currently represented as an int64 instead of float64, so the value is always rounded down to the nearest whole second. This is a problem that existed before your PR, as a previous submitter adjusted the metric name from milliseconds (where integers made sense) to seconds, but did not fully address the calculation logic. If you'd like to make that change, that would be great! Otherwise I can handle it in the Influx support fork.

123Haynes:
@jranson https://github.com/Comcast/trickster/pull/50 :wink:

FUSAKLA:
Thank you for fixing those issues @123Haynes I didn't find the time to solve them myself :+1:

PR #48, opened 2018-06-04
change default cache path to /tmp/trickster

123Haynes:
This PR changes the default cache path  from `/tmp` to `/tmp/trickster` so it aligns with the docs and the sample config file https://github.com/Comcast/trickster/blob/master/conf/example.conf#L45

closes https://github.com/Comcast/trickster/issues/38


jranson:
@123Haynes Thank you! I had caught this and fixed it in my Influx fork but we're happy to merge your change now.

PR #58, opened 2018-06-15
Enforce strict error handling and early aborts

juliusv:
This fixes a number of crashes that currently happen when either the
origin Prometheus cannot be reached or when a user request is faulty.

It also explicitly hard-errors to the user whenever a partial error
happens somewhere in Trickster (due to unreachable origin, faulty JSON
marshaling, or else). There were many cases where an error was logged,
but not handled, causing undefined/crashy behavior. These errors are now
handled and cause the entire request to fail. This seems better than
returning wrong results or crashing.

I also added tests for the obvious crash cases and check whether they
are now returning a user-exposed HTTP error status code instead.

A TODO for the future would be to clean up the metrics handling, since
it currently happens before or after an error return, and errors aren't
properly tracked in the metrics.

Fixes https://github.com/Comcast/trickster/issues/49

### DIFF
--- a/constants.go
+++ b/constants.go
@@ -92,11 +92,9 @@ const (
 
 // Log Fields
 const (
-	lfEvent      = "event"
-	lfDetail     = "detail"
-	lfParamName  = "paramName"
-	lfParamValue = "paramValue"
-	lfCacheKey   = "cacheKey"
+	lfEvent    = "event"
+	lfDetail   = "detail"
+	lfCacheKey = "cacheKey"
 )
 
 // http methods
--- a/handlers.go
+++ b/handlers.go
@@ -56,7 +56,12 @@ func (t *TricksterHandler) promHealthCheckHandler(w http.ResponseWriter, r *http
 	path := "/api/v1/" + mnLabels
 
 	originURL := t.getOrigin(r).OriginURL + strings.Replace(path, "//", "/", 1)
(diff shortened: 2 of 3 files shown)


SuperQ:
FYI, this is now live on [monitor.gitlab.net](https://monitor.gitlab.net/dashboard/db/trickster).

juliusv:
Could someone retrigger the Travis build? It failed on a transient error:

```
package gopkg.in/natefinch/lumberjack.v2: unrecognized import path "gopkg.in/natefinch/lumberjack.v2" (https fetch: Get https://gopkg.in/natefinch/lumberjack.v2?go-get=1: unexpected EOF)
make: *** [deps] Error 1
```

SuperQ:
It looks like something strange happens occasionally:

image

The cache has no data for the tail of the query, but it's front-filling normally.

juliusv:
@SuperQ Do you have a direct indication that this was caused by this PR? Just reading my changes again I'm not sure how that could happen as a result of them, but who knows.

juliusv:
@SuperQ 

> I have no idea. I would have to figure out what queries cause what cache objects to be created and see what's in the logs. If I can come up with a more reproducible test, I'll let you know.

Maybe if you're not using the in-memory cache, but the FS one, you could switch back to the `master` branch version once you see this problem again and see if it persists with that.

> For sure, this change seems to be much more stable.

Great!

PR #62, opened 2018-06-24
Require "step" URL parameter in range queries

juliusv:
Prometheus's API requires the "step" parameter to be present in range
queries, so Trickster can do the same instead of providing a default
value.

This also ensures that the step is positive, as otherwise a
divide-by-zero exception can happen further on in
buildRequestContext().

### DIFF
--- a/config.go
+++ b/config.go
@@ -75,7 +75,6 @@ type FilesystemCacheConfig struct {
 type PrometheusOriginConfig struct {
 	OriginURL           string `toml:"origin_url"`
 	APIPath             string `toml:"api_path"`
-	DefaultStep         int    `toml:"default_step"`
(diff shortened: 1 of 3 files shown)


juliusv:
Note: this is a breaking change as it removes the `DefaultStep` config file parameter, but I hope breaking changes are still ok at the 0.0.x stage.

jranson:
@juliusv Yep - should be fine to break the default step convenience. I assume 100% of folks are using w/ Grafana which does the right thing anyway. Thank you!

juliusv:
Thanks! Yep, any valid Prometheus client has to include the step anyway
because otherwise Prometheus (without Trickster at least) will return an
error.

PR #68, opened 2018-06-30
Don't cache generically-handled paths

jranson:


jranson:
Apologies for all the apparent changes on the Test file, it looks like my Textmate formatter is different than what Phil uses.  The only functional change to tests is the removal of TestTricksterHandler_promAPIProxyHandler since that func was subsumed by promFullProxyHandler.

handlers_test.go
```diff
@@ -14,25 +14,25 @@
 package main
 
 import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"github.com/go-kit/kit/log"
```
SuperQ: We usually sort external imports from stdlib imports by putting them at the end with a newline between.

SuperQ:
@jranson I think the issue is missing `go fmt`.  If you want, I can clean that up in a separate PR.

SuperQ:
https://github.com/Comcast/trickster/pull/69 should cleanup the code issues.

handlers_test.go
jranson: Looks like you picked this one up in #69. I will merge #68 first once it gets the thumbs up.

PR #69, opened 2018-06-30
Cleanup style

SuperQ:
* Add a `style` make target.
* Run tests in travis.
* `go fmt`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,6 @@ go:
   - "1.9"
   - "1.10.x"
   - master
+
+script:
+- make style build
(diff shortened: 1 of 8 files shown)


jranson:
Looks like we are getting test errors after these updates. I am re-running now to make sure it is not environmental.

SuperQ:
From what I see in the test logs, it could be a flaky test.

jranson:
I think there might be two separate issues. I can reproduce the filesystem cache test failure locally, but very intermittently (perhaps one out of every 30 runs). The tests are failing reliably on Travis, but only on the go:master tests. The two versioned go tests seem to work every time, except when they encounter the filesystem cache race condition. I'll take a look at the cache issue, but not sure what to make of the go:master issue.

SuperQ:
I've removed the tests from Travis for now. Once the test issues are fixed, we can re-enable them in the test pipeline.

PR #70, opened 2018-07-01
BoltDb Cache Support

### LINKED ISSUE #55: Feature: BoltDB caching backend
As a possibly more optimal local filesystem storage, use a BoltDB database to store the cache key/values.

Specifically, we may want to use the "official" fork, [coreos/bbolt](https://github.com/coreos/bbolt).

jranson:
Fixes #55

### DIFF
--- a/boltdb.go
+++ b/boltdb.go
@@ -24,21 +24,21 @@ import (
 	"github.com/go-kit/kit/log/level"
 )
 
-// BoltDbCache describes a BoltDb Cache
-type BoltDbCache struct {
+// BoltDBCache describes a BoltDB Cache
+type BoltDBCache struct {
 	T      *TricksterHandler
-	Config BoltDbCacheConfig
+	Config BoltDBCacheConfig
 	dbh    *bolt.DB
 }
 
-// Connect instantiates the BoltDbCache mutex map and starts the Expired Entry Reaper goroutine
-func (c *BoltDbCache) Connect() error {
+// Connect instantiates the BoltDBCache mutex map and starts the Expired Entry Reaper goroutine
+func (c *BoltDBCache) Connect() error {
 
 	fullPath := path.Join(c.Config.CachePath, c.Config.Filename)
 
 	level.Info(c.T.Logger).Log("event", "boltdb cache setup", "cachePath", fullPath)
 
-	err := mustMakeDirectory(c.Config.CachePath)
+	err := makeDirectory(c.Config.CachePath)
 	if err != nil {
 		return err
 	}
@@ -64,7 +64,7 @@ func (c *BoltDbCache) Connect() error {
 }
 
 // Store places an object in the cache using the specified key and ttl
-func (c *BoltDbCache) Store(cacheKey string, data string, ttl int64) error {
+func (c *BoltDBCache) Store(cacheKey string, data string, ttl int64) error {
 
 	expKey, dataKey := c.getKeyNames(cacheKey)
 	expiration := []byte(strconv.FormatInt(time.Now().Unix()+ttl, 10))
@@ -95,7 +95,7 @@ func (c *BoltDbCache) Store(cacheKey string, data string, ttl int64) error {
 }
 
 // Retrieve looks for an object in cache and returns it (or an error if not found)
-func (c *BoltDbCache) Retrieve(cacheKey string) (string, error) {
+func (c *BoltDBCache) Retrieve(cacheKey string) (string, error) {
 
 	level.Debug(c.T.Logger).Log("event", "boltdb cache retrieve", "key", cacheKey)
 
@@ -107,7 +107,7 @@ func (c *BoltDbCache) Retrieve(cacheKey string) (string, error) {
 }
 
 // retrieve looks for an object in cache and returns it (or an error if not found)
-func (c *BoltDbCache) retrieve(cacheKey string) (string, error) {
+func (c *BoltDBCache) retrieve(cacheKey string) (string, error) {
 
 	content := ""
 
@@ -129,7 +129,7 @@ func (c *BoltDbCache) retrieve(cacheKey string) (string, error) {
 }
 
 // checkExpiration verifies that a cacheKey is not expired
-func (c *BoltDbCache) checkExpiration(cacheKey string) {
+func (c *BoltDBCache) checkExpiration(cacheKey string) {
 
 	expKey, _ := c.getKeyNames(cacheKey)
 
@@ -144,7 +144,7 @@ func (c *BoltDbCache) checkExpiration(cacheKey string) {
 }
 
 // Delete removes an object in cache, if present
-func (c *BoltDbCache) Delete(cacheKey string) error {
+func (c *BoltDBCache) Delete(cacheKey string) error {
 
 	level.Debug(c.T.Logger).Log("event", "boltdb cache delete", "key", cacheKey)
 
@@ -189,7 +189,7 @@ func (c *BoltDbCache) Delete(cacheKey string) error {
 }
 
 // Reap continually iterates through the cache to find expired elements and removes them
-func (c *BoltDbCache) Reap() {
+func (c *BoltDBCache) Reap() {
 
 	for {
 		c.ReapOnce()
@@ -198,7 +198,7 @@ func (c *BoltDbCache) Reap() {
 
 }
 
-func (c *BoltDbCache) ReapOnce() {
+func (c *BoltDBCache) ReapOnce() {
 
 	now := time.Now().Unix()
 	expiredKeys := make([]string, 0)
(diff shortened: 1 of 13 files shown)


conf/example.conf
```diff
@@ -45,6 +45,21 @@
             # default is '/tmp/trickster'
             #cache_path = '/tmp/trickster'
 
+        # Configuration options when using a BoltDb Cache
```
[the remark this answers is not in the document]
jranson: Looks like you are addressing these items for the conf file with PR #71 so I will disregard. Others are coming up momentarily.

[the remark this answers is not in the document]
jranson: Check that. I will rebase with #71 and clean this up.

boltdb.go
```diff
@@ -0,0 +1,248 @@
...
+}
+
+// Connect instantiates the BoltDbCache mutex map and starts the Expired Entry Reaper goroutine
+func (c *BoltDbCache) Connect() error {
+
+	fullPath := path.Join(c.Config.CachePath, c.Config.Filename)
+
+	level.Info(c.T.Logger).Log("event", "boltdb cache setup", "cachePath", fullPath)
+
+	err := mustMakeDirectory(c.Config.CachePath)
```
SuperQ: I'd like to avoid having Trickster create the database directory, this is the responsibility of the config management.

boltdb.go
```diff
@@ -0,0 +1,248 @@
...
+	"strconv"
+	"strings"
+	"time"
+
+	bolt "github.com/coreos/bbolt"
+	"github.com/go-kit/kit/log/level"
+)
+
+// BoltDbCache describes a BoltDb Cache
+type BoltDbCache struct {
```
juliusv: If you want to follow common Go naming guidelines (https://github.com/golang/go/wiki/CodeReviewComments#initialisms) as well as the name that the DB gives itself, `BoltDb` should be `BoltDB` everywhere.

filesystem.go
```diff
@@ -161,7 +161,7 @@ func writeable(path string) bool {
 func mustMakeDirectory(path string) error {
```
juliusv: The `must*` prefix in Go is by convention used for functions that panic if they fail. If they return an `error`, the convention would just be `makeDirectory`.

boltdb.go
juliusv: I think that's fine, Prometheus does it too, right?

boltdb.go
```diff
@@ -0,0 +1,248 @@
...
+		return nil
+	})
+
+	// Iterate through the expired keys so we can delete them
+	for _, cacheKey := range expiredKeys {
+
+		c.Delete(cacheKey)
+	}
+
+	return
```
juliusv: Remove unneeded return

boltdb.go
```diff
@@ -0,0 +1,246 @@
...
+
+		c.Delete(cacheKey)
+	}
+
+}
+
+// Close closes the BoltDBCache
+func (c *BoltDBCache) Close() error {
+	c.dbh.Close()
+	return nil
```
juliusv: Since `c.dbh.Close()` can return an error, would `return c.dbh.Close()` make more sense here?

boltdb.go
```diff
@@ -0,0 +1,245 @@
...
+		if err != nil {
+			return err
+		}
+
+		err = b.Put([]byte(expKey), expiration)
+		if err != nil {
+			return err
+		}
+
+		return nil
```
juliusv: The above 6 lines can be condensed to:

```go
return b.Put([]byte(expKey), expiration)
```

conf/example.conf
```diff
@@ -40,6 +40,21 @@ cache_type = 'memory'
...
 
+    # Configuration options when using a BoltDb Cache
+    #[cache.boltdb]
+
+    # cache_path defines the directory location under which the Trickster cache will be maintained
+    # default is '/tmp/trickster'
+    #cache_path = '/tmp/trickster'
+
+    # filename defines the name of the BoltDb database file written to cache_path
+    # default is 'bolt.db'
```
juliusv: Defaulting to `trickster.db` might make more sense to name it after the application rather than the database file type?

juliusv: Though if it's in a trickster directory already, who cares...

config.go
```diff
@@ -50,20 +50,31 @@ type CachingConfig struct {
...
-type RedisConfig struct {
+// RedisCacheConfig is a collection of Configurations for Connecting to Redis
+type RedisCacheConfig struct {
 	// Protocol represents the connection method (e.g., "tcp", "unix", etc.)
 	Protocol string `toml:"protocol"`
 	// Endpoint represents FQDN:port or IPAddress:Port of the Redis server
 	Endpoint string `toml:"endpoint"`
 }
 
+// BotlDbCacheConfig is a collection of Configurations for storing cached data on the Filesystem
```
juliusv: BotlDb :)

jranson: Damn you!!! :-)     No wonder my `sed` didn't pick that one up

docs/caches.md
```diff
@@ -22,13 +22,17 @@ The Filesystem Cache is a popular option when you have larger dashboard setup (e
 
 The default Filesystem Cache path is `/tmp/trickster`. The sample configuration demonstrates how to specify a custom cache path. Ensure that the user account running Trickster has read/write access to the custom directory or the application will exit on startup upon testing filesystem access. All users generally have access to /tmp so there is no concern about permissions in the default case.
 
+## BoltDB Cache
+
```
juliusv: "is filesystem-based solution" -> "is a filesystem-based solution"

juliusv:
General high-level style comment: a bit of a whitespace (empty line) inflation in some parts. But hey, the creator can determine the style I guess :)

boltdb.go
SuperQ: Yea, it does, but Prometheus database is a lot more complicated than a single boltdb file. :-)

conf/example.conf
SuperQ: Yes, I think having the file named `trickster.db`, as a relative path rather than absolute, would be better.

SuperQ:
I think for the database file, a single file path with `trickster.db` would be simplest.  Config management can either set the `CWD`, or provide a full path.

jranson:
On the pathing stuff, my general concerns are that we would want to do the same thing for the filesystem cache, and users who are already using Trickster with filesystem would have to react to this change as part of the update and both manually remove the older orphaned files and adjust their conf. Then there is logging to consider - the lumberjack package automatically makes the directories as needed when file-based logging is enabled. I think users should be able to easily and explicitly tell you where to put files, and the application should make a best effort to create the enclosing directory when missing. That tenet ensures all aspects of the application that create files will work predictably.

SuperQ:
It's fine to make directories for things that need it, like the filesystem cache option. IMO the botldb option doesn't need it, it only needs to create the single file.

Cleaning up orphaned files is out of scope for the code here, that's up to configuration management / packaging.

SuperQ:
While I object to the mkdir, you could still implement it by having a single `filename` config option and use [`filepath.Split()`](https://golang.org/pkg/path/filepath/#Split) to extract and create any directory from the config.

config.go
```diff
@@ -98,14 +107,21 @@ type LoggingConfig struct {
 
 // NewConfig returns a Config initialized with default values.
 func NewConfig() *Config {
+
+	defaultCachePath := "/tmp/trickster"
```
SuperQ: This should be just `trickster.db`, and below we don't want to include an additional filename now.

jranson: The defaultCachePath is used to instantiate both Filesystem (defaultCachePath) and BoltDb cache (defaultCachePath+filename) types, depending upon the supplied configuration. This change is as per your recommendation above to use a single configuration string to indicate the full path and use `filepath.Split()` to make the directory for a BoltDB cache.

I understand there is a philosophical difference here about whether the app should handle pathing and directory creation, but that disagreement is holding up the release. 

Here is how I propose we proceed. If, other than the disagreement about how Trickster should work with the filesystem, there are no issues with the PR, let's merge it. I would rather maintain consistency across all of the areas in the app where we touch the filesystem than introduce a new feature that works differently. That is what I am doing here.

If we really want to change the pathing philosophy across the entire application, let's work on that in a separate PR that addresses all of the filesystem-dependent areas of the app. That way, across any given release, how Trickster works with the filesystem is identical between features.

SuperQ: I guess that's fine, I have a specific design idea I can propose.

conf/example.conf
SuperQ: This documentation needs to be updated because the code does not allow you to change the file name.

SuperQ:
Ping? This is mostly ready for merge, just have to fix the docs to match.

jranson:
Hi Ben, I've been on holiday for the past couple weeks but will finish it
up as soon as I get back to the PC later this week! Sorry for the delay!

SuperQ:
No problem! Hope you had/are having a good holiday.

PR #72, opened 2018-07-02
Add support for POST API endpoints

jacksontj:
Prometheus upstream added support for POST API calls in https://github.com/prometheus/prometheus/pull/3322

This changes the buildRequestContext method to check both the query
params and the body by using ParseForm().


main.go
```diff
@@ -79,16 +79,16 @@ func main() {
 	router.HandleFunc("/"+mnHealth, t.promHealthCheckHandler).Methods("GET")
 
 	// Path-based  multi-origin support - no support for full proxy of the prometheus UI, only querying
-	router.HandleFunc("/{originMoniker}"+prometheusAPIv1Path+mnQueryRange, t.promQueryRangeHandler).Methods("GET")
-	router.HandleFunc("/{originMoniker}"+prometheusAPIv1Path+mnQuery, t.promQueryHandler).Methods("GET")
-	router.PathPrefix("/{originMoniker}" + prometheusAPIv1Path).HandlerFunc(t.promFullProxyHandler).Methods("GET")
+	router.HandleFunc("/{originMoniker}"+prometheusAPIv1Path+mnQueryRange, t.promQueryRangeHandler).Methods("GET", "POST")
+	router.HandleFunc("/{originMoniker}"+prometheusAPIv1Path+mnQuery, t.promQueryHandler).Methods("GET", "POST")
+	router.PathPrefix("/{originMoniker}"+prometheusAPIv1Path).HandlerFunc(t.promFullProxyHandler).Methods("GET", "POST")
```
juliusv: Only the `query` and `query_range` endpoints got `POST` support added, so I think the `promFullProxyHandler` doesn't need it?

handlers.go
```diff
@@ -409,7 +409,10 @@ func (t *TricksterHandler) buildRequestContext(w http.ResponseWriter, r *http.Re
 	ctx.Origin.OriginURL += strings.Replace(ctx.Origin.APIPath+"/", "//", "/", 1)
 
 	// Get the params from the User request so we can inspect them and pass on to prometheus
-	ctx.RequestParams = r.URL.Query()
+	if err := r.ParseForm(); err != nil {
+		return nil, err
```
juliusv: All the other errors in this method are `fmt.Errorf()`-wrapped to add context, would be nice to do the same here.

jacksontj: @juliusv I'll do that, but It seems like this is a perfect use-case for https://github.com/pkg/errors :)

handlers.go
```diff
@@ -410,7 +411,7 @@ func (t *TricksterHandler) buildRequestContext(w http.ResponseWriter, r *http.Re
 
 	// Get the params from the User request so we can inspect them and pass on to prometheus
 	if err := r.ParseForm(); err != nil {
-		return nil, fmt.Errorf("Unable to parse form: %v", err)
+		return nil, errors.Wrap(err, "Unable to parse form")
```
juliusv: start error strings lower-cased (https://github.com/golang/go/wiki/CodeReviewComments#error-strings)

handlers.go
```diff
@@ -421,7 +422,7 @@ func (t *TricksterHandler) buildRequestContext(w http.ResponseWriter, r *http.Re
 	ctx.StepParam = ctx.RequestParams[upStep][0]
 	step, err := strconv.ParseInt(ctx.StepParam, 10, 64)
 	if err != nil {
-		return nil, fmt.Errorf("failed to parse parameter %q with value %q: %v", upStep, ctx.StepParam, err)
+		return nil, errors.Wrap(err, fmt.Sprintf("failed to parse parameter %q with value %q: %v", upStep, ctx.StepParam))
```
juliusv: You'll need to remove the `%v` here and in the instances below where you've removed `err` from the `fmt.Sprintf()` args.

juliusv:
Generally :+1: from my side

jacksontj:
@juliusv thanks for the review, updated the last comments you had-- so we should be set.

jacksontj:
@jranson great, thanks for the quick response!

PR #76, opened 2018-07-07
Allow non-int formats for "step" value

juliusv:
Fixes https://github.com/Comcast/trickster/issues/74


juliusv:
@jranson Could you merge? I don't have write access it seems.

PR #78, opened 2018-07-09
Add test to build pipeline

SuperQ:
* Run all tests in Travis.
* Use `dep ensure` to install deps.
* Update deps.
* Misc Makefile cleanup.

Closes: https://github.com/Comcast/trickster/issues/75

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,7 @@ GO           ?= go
 GOFMT        ?= $(GO)fmt
 FIRST_GOPATH := $(firstword $(subst :, ,$(shell $(GO) env GOPATH)))
 DEP          := $(FIRST_GOPATH)/bin/dep
+TRICKSTER    := $(FIRST_GOPATH)/bin/trickster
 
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -41,16 +51,16 @@ docker-release:
...
-	go get github.com/alicebob/miniredis
-	go test -run '' -o ${GOPATH}/bin/trickster -v
+.PHONY: test
+test: deps
+	go test -o ${GOPATH}/bin/trickster -v ./...
 
-test-cover:
-	go get github.com/alicebob/miniredis
-	go test -run '' -o ${GOPATH}/bin/trickster -coverprofile=cover.out
+.PHONY: test-cover
```
juliusv: Why define some `.PHONY`s above the target and others in a list at the bottom? Should be consistent IMO unless there's a reason.

SuperQ: Yea, I was trying to move towards having it be with the targets.

juliusv: Then IMO do it all at once...

SuperQ: Done

juliusv:
:+1:

PR #79, opened 2018-07-09
Treat errors unmarshaling from cache as cache miss

### LINKED ISSUE #39: Error on unmarshaling cached data
From time to time Trickster encounters error on unmarshalling data stored in cache.
It does not crash just logs those errors. Also I'm not sure if that means it does not server those data at all or queries the Prometheus to fix this error data?

It happens on for both, the filesystem and memory cache

In log I can see
```
time=2018-04-22T12:19:14.029382922Z app=trickster caller=trickster/handlers.go:445 level=error event="could not unmarshal cached data for this key" key=5afe9f97855959db71d477df683d9348.c837a21ba2b43ebcb924c5c0b23265ed data="{\ufffd{\"status\":\"success\",\"data\":{\"resultType\":\"matrix\",\"\t\u0016\ufffd\":[{\"metric\":{},\"values\":[[1524399120,\"12\"],[\r\u00124420,\"12\"]]}]}}"

```
This is dump of the impacted cache file 
[5afe9f97855959db71d477df683d9348.c837a21ba2b43ebcb924c5c0b23265ed.data.tar.gz](https://github.com/Comcast/trickster/files/1935566/5afe9f97855959db71d477df683d9348.c837a21ba2b43ebcb924c5c0b23265ed.data.tar.gz)

I'm running Trickster `0.0.12` in Kubernetes as proxy for Prometheus 2.2.1 also running in Kubernetes

jacksontj:
Prior to this patch any error unmarshaling from the cache would cause
the request to fail until the item dropped out of cache. This change
simply treats a bad entry as a miss, so it will re-fetch from origin

Fixes #39

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -505,8 +505,11 @@ func (t *TricksterHandler) buildRequestContext(w http.ResponseWriter, r *http.Re
 
 		// Marshall the cache payload into a PrometheusMatrixEnvelope struct
 		err = json.Unmarshal([]byte(cachedBody), &ctx.Matrix)
+		// If there is an error unmarshaling the cache we should treat it as a cache miss
+		// and re-fetch from origin
 		if err != nil {
-			return nil, errors.Wrap(err, fmt.Sprintf("error unmarshalling cached data for key %q with content %q", ctx.CacheKey, cachedBody))
+			ctx.CacheLookupResult = crRangeMiss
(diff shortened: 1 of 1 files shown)


FUSAKLA:
Thanks for fixing this!
Will test it asap

PR #81, opened 2018-07-10
Post

jacksontj:
While testing some more POST usage, I found 2 problems. (1) point-in-time `/query` calls weren't parsing the form (2) the method to origin was always GET regardless of client method (now it mirrors client method).


jacksontj:
cc @jranson

Issue #82, opened 2018-07-10
Unable to handle scalar responses


jacksontj:
If I send a simple scalar query such as `/api/v1/query?query=5` trickster now returns errors. I bisected this and it seems that the error was introduced in ec4eff34d5532b1907723eeaabe620f02dd25b32 . The basic problem here is that trickster assumes the response type from prometheus as a vector, when in reality it can be (1) scalar (2) vector or (3) string. The way this worked before was that the unmarshal error was ignored and the response handed back (I have a [patch](https://github.com/jacksontj/trickster/commit/aa6255d9a17276dccd9c01cca8f47510bdb56ba5) that puts it back to that behavior.

From what I see that looks like caching won't work on those scalar values -- as it wasn't able to unmarshal. Alternatively we could change the model to use `model.Value` and then add some type-switching. But since that is a potentially large change I figured I'd get others' input first.

SuperQ:
The main reason we changed to more strict handling was that there were a bunch of cases where talking to Prometheus is interrupted (server restart, crash, etc) and Trickster would cache bad data.

We should probably type switch, like you said.

/cc @juliusv

jacksontj:
If we want to move to type switching I'd suggest consolidating the response and envelope/data to a single set of structs, something like: 

``` go
type PrometheusResponse struct {
	Status string               `json:"status"`
	Data   PrometheusVectorData `json:"data"`
}

type PrometheusDataEnvelope struct {
	ResultType string       `json:"resultType"`
	Result     model.Value `json:"result"`
}
```

Then we can change the callers to handle the switching. If you are okay with that plan I'm okay with doing the implementation, I just want to make sure thats something we want before spending the time.

jacksontj:
Something along the lines of https://github.com/Comcast/trickster/compare/master...jacksontj:upstream_type

jacksontj:
Sadly the upstream structs [still aren't public](https://github.com/prometheus/prometheus/pull/3615), so if we want to attempt to have *less* code I do have library methods for this in the [promxy](https://github.com/jacksontj/promxy/tree/master/promhttputil) project, just not sure what feelings are for dependency-- so this approach in my branch might be sufficient.

juliusv:
I haven't looked at anything here yet as I'm still half asleep from a boat trip, but FWIW this is how our API client library does it: https://github.com/prometheus/client_golang/blob/master/api/prometheus/v1/api.go#L160-L200

SuperQ:
[...]

PR #83, opened 2018-07-11
Workaround to handle scalar responses

jacksontj:
This is a short-term workaround for #82 as the "correct" fix will take some time, but this fixes it *now*.

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -317,7 +317,11 @@ func (t *TricksterHandler) getVectorFromPrometheus(url string, params url.Values
 	// Unmarshal the prometheus data into another PrometheusMatrixEnvelope
 	err = json.Unmarshal(body, &pe)
 	if err != nil {
-		return pe, nil, fmt.Errorf("Prometheus vector unmarshaling error for URL %q: %v", url, err)
+		// If we get a scalar response, we just want to return the resp without an error
+		// this will allow the upper layers to just use the raw response
+		if pe.Data.ResultType != "scalar" {
+			return pe, nil, fmt.Errorf("Prometheus vector unmarshaling error for URL %q: %v", url, err)
+		}
 	}
 
 	return pe, resp, nil


handlers.go
```diff
@@ -317,7 +317,11 @@ func (t *TricksterHandler) getVectorFromPrometheus(url string, params url.Values
 	// Unmarshal the prometheus data into another PrometheusMatrixEnvelope
 	err = json.Unmarshal(body, &pe)
 	if err != nil {
-		return pe, nil, fmt.Errorf("Prometheus vector unmarshaling error for URL %q: %v", url, err)
+		// If we get a scalar response, we just want to return the resp without an error
+		// this will allow the upper layers to just use the raw response
```
juliusv: Can you point me at the upper levels that still use the response if the JSON parsing failed?

For callers of `getVectorFromPrometheus()` I only found these:

- https://github.com/Comcast/trickster/blob/d7191ce7bc4cf222b1ddc5423d10981b3deb09d3/handlers.go#L615
- https://github.com/Comcast/trickster/blob/d7191ce7bc4cf222b1ddc5423d10981b3deb09d3/handlers.go#L782

In both cases, as far as I can see, only the HTTP status code of the response is ever used, not any scalar contents it might contain.

jacksontj: @juliusv from my testing I see that everything besides that fast-forward request works. Its still not ideal as its missing a point, but this PR does put it back to the previous behavior (which was definitely less broken). I'm planning on focusing on #82 as the longer-term fix for this, since the old behavior was "good enough" (I didn't even notice it wasn't 100% right until working on this issue ;) )

juliusv: I guess it depends on whether you consider silently ignoring partial errors as "less broken" :) But fine, as long as it gets fixed properly later.

jranson: Yup, we're keeping #82 open until the permanent fix is in

Issue #56, opened 2018-06-07
Question: Performance of trickster and memory usage


FUSAKLA:
Hi, we discussed this at the slack channel some time ago when I was deploying Trickster in our company and had some issues regarding memory usage and this leading to frequent OOM kills.

Since than I did not see anyone mentioning any bigger setups but now @SuperQ mentioned deploying it in GitLab so I'm wondering how is Trickster working for you? 

Would it be possible to share any performance data? 

Possibly we can take this to the [slack channel](https://gophers.slack.com/messages/C9VDJLYA2) 

Thank's for any info on this

SuperQ:
Yet another slack channel, happy to talk about it on the official [#prometheus IRC/Riot channel](https://prometheus.io/community/).

FUSAKLA:
Well Trickster performance is not Prometheus related actually, but ok will take it there.

Thanks

PR #85, opened 2018-07-13
Release request context data after handling

juliusv:
Every range query handler goroutine (one for each cache key, of which
there can be hundreds or thousands) currently holds on to a reference to
the last request context that it handled. This is since the channel
range loop variable (which holds the request context) is only
overwritten once a subsequent request is received from the channel. This
change explicitly sets the finished request context to nil so that Go
can release the associated memory.

In GitLab, this has fixed OOM problems in Trickster, and memory usage is
now significantly lower.

Signed-off-by: Julius Volz <julius.volz@gmail.com>

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -882,6 +882,9 @@ func (t *TricksterHandler) originRangeProxyHandler(cacheKey string, originRangeR
 			writeResponse(r.Writer, body, resp)
 			r.WaitGroup.Done()
 		}
+		// Explicitly release the request context so that the underlying memory can be
+		// freed before the next request is received via the channel, which overwrites "r".
+		r = nil
 	}
 }
 


jranson:
@FUSAKLA, this may address #56 for you.

SuperQ:
Pics or it didn't happen: :grin: 
trickster-memory

You can see the fix deploy at 14:00 UTC. And a 10m pause in queries at 14:45 (turned off grafana just to see).

SuperQ:
Just for context, here's how many gouroutines are in use. With the patch we see a bunch more, but we don't OOM anymore.
trickster-goroutines

juliusv:
Note that in the screenshot above I had only set `r = nil` for non-cached requests and not for cached yet. So hopefully behavior with this PR will be even better.

juliusv:
> Just for context, here's how many gouroutines are in use. With the patch we see a bunch more, but we don't OOM anymore.

Yeah, that's just because now we can actually fit all the goroutines that want to be there anyway into memory :)

juliusv:
This is now deployed at GitLab and it seems it has cut the peak `go_memstats_heap_inuse_bytes` from ~35GB down to ~6GB.

juliusv:
Spoke too soon, now it's up to 10GB, but still a >3x gain (and we don't know how much larger the old version would have grown, since it OOMed) :)

FUSAKLA:
This looks great I'll definatelly try out.
Thanks for the heads up @jranson!

PR #89, opened 2018-08-15
Add option to authenticate with redis servers

stewrutledgebonnier:
We have all of our redis instances setup with authentication which makes using them as a backend impossible.

This change just adds AUTH support to the redis client, based on the presence of password in the configuration file.


jranson:
@juliusv Any concerns about this one (e.g., storing passwords in plaintext)?

SuperQ:
Storing credentials in a file is the preferred thing these days. LGTM.

SuperQ:
ENV vars are considered unsafe compared to files.

stewrutledgebonnier:
Not implying otherwise, I just mean that I used the existing configuration convention instead of attempting to implement any new ones.

PR #103, opened 2018-12-05
Close handler goroutines if their channel is empty

SuperQ:
We found that with a large amount of traffic, closing idle handlers helps a ton with memory/go routine leakage.


juliusv:
@jranson It's honestly a bit of a hack, but I didn't see a better way to do this without major restructuring at the time :) Would be good to get independent confirmation that this leads to much lower memory usage (at least in setups with many different cache keys)!

FUSAKLA:
I'll give it a try today and will report back if it helped somehow significantly.

PR #108, opened 2018-12-07
Test + Fix for Issue 105

jacksontj:


jacksontj:
@jranson they are not intended, thanks for the catch. I've just updated the branch

PR #111, opened 2018-12-15
Added RPM spec and build target.  Minor Makfile clean up.

JC000:
Plus others tweaks to make a working RPM.


JC000:
Not a problem at all.  Just a note on the "minor" changes.  I changed the `default` target to build a local binary, and added the `install` target to install under `bin`.  It was just weird for a `clean` target to delete stuff out of the `bin` directory.

PR #114, opened 2018-12-17
Validate ordering assumption in mergeMatrix

### LINKED ISSUE #92: Trickster provokes grafana display quirks
I noticed a few grafana display bugs in my graphs which at first I blamed on Thanos but forgot I had a trickster in front of it.

Expected graph
grafana-prometheus

`Thanos + Trickster` graph
grafana-thanos-2

The data are the same but only come from different origin/path.

This only happens when I set the end of range to `now` in grafana. If fix the end of range to a fixed time I can't reproduce.

Taking a look at the data coming from `Prometheus` and `Thanos + Trickster` I got the following diff

screenshot_15

As you can see the last timestamp is not aligned with the `step` in the `Thanos + Trickster` case and it confuses grafana's output.

Bypassing Trickster fixes the problem.

jacksontj:
the current implementation of mergeMatrix assumes that `pe2` is "before"
`pe`-- but does not thing to actually ensure or enforce that this is the
case. This means if there is overlap we can get dupe points. For
example (with the test case in the PR):

```
--- FAIL: TestTricksterHandler_mergeMatrix (0.00s)
    --- FAIL: TestTricksterHandler_mergeMatrix/0 (0.00s)
    	handlers_test.go:542: Mismatch
    		actual={success {matrix a =>
    		1.5 @[0.001]
    		1.5 @[0.002]
    		1.5 @[0.001]}}
    		expected={success {matrix a =>
    		1.5 @[0.001]
    		1.5 @[0.002]}}
```

This is likely the cause for a variety of "dupe points" or other
out-of-order points problems.

This change still assumes that the 2 series are "in order", if we wanted to not rely on that at all we'd need a more complete merging/sorting implementation like that in [promxy](https://github.com/jacksontj/promxy/blob/master/promhttputil/merge.go)

Fix #92

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -993,12 +993,23 @@ func (t *TricksterHandler) mergeMatrix(pe PrometheusMatrixEnvelope, pe2 Promethe
 	for i := range pe2.Data.Result {
 		metricSetFound := false
 		result2 := pe2.Data.Result[i]
+	METRIC_MERGE:
 		for j := range pe.Data.Result {
 			result1 := pe.Data.Result[j]
 			if result2.Metric.Equal(result1.Metric) {
 				metricSetFound = true
-				pe.Data.Result[j].Values = append(pe2.Data.Result[i].Values, pe.Data.Result[j].Values...)
-				break
+				// Ensure that we don't duplicate datapoints or put points out-of-order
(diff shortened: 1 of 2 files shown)


jranson:
Nice!  Thanks for tracking this down, and helping expand the test cases. I really appreciate it!

PR #117, opened 2019-01-08
Fix ordering assumption enforcement for mergeMatrix

### LINKED ISSUE #116: query_range no longer returns the full merged data set
Using local build from master (646440f678b8e33b23412f8645faf58fe8a6ff46).

1. setup new trickster with copy of `example.conf` with an appropriate prometheus url
1. query prometheus graph via proxy web interface to fill up its cache
   1. first fetch returns a full set of data (~250 data points)
1. wait for scraping to make more data and execute again
   1. trickster only returns points newer than what was on the graph (~3 data points)
   1. graph is empty except for the last couple points

When using a build from the `0.1.5` tag, it works as expected and all data is returned. As far as I can tell, it seems #114 would be the cause since not much else has changed in master since `0.1.5`.

Thanks and I hope this helps!

jacksontj:
Fix for 9c6aecb53443134537847a5b46b351a5f689b694

The original commit correctly stated that the assumption was that pe2
was before pe, but it actually enforced the exact opposite. This fixes
the enforcement and updates the tests

Fixes #116

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -1001,10 +1001,11 @@ func (t *TricksterHandler) mergeMatrix(pe PrometheusMatrixEnvelope, pe2 Promethe
 				// Ensure that we don't duplicate datapoints or put points out-of-order
 				// This method assumes that `pe2` is "before" `pe`, we need to actually
 				// check and enforce that assumption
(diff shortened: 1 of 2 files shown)


jacksontj:
cc @jranson

Issue #91, opened 2018-08-23
High memory usage with redis cache backend


stewrutledgebonnier:
We're using the redis cache backend in our instance of trickster and we're seeing suprisingly high memory usage.

We are running trickster in kubernetes, with memory limited to 4GB. Trickster only needs to be up for half an hour before it's killed for using more than 4GB memory (OOM).

Has anyone seen any similar behavior? Any idea what could be going on? We've tested increasing the memory limits, they were originally set to 2GB, but the problem persists.

FUSAKLA:
Hi, what version do you use? 

I experienced the same behavior but since Julius fixe this
https://github.com/Comcast/trickster/pull/85

it's way more stable and works like a charm for me I have to say.

stewrutledgebonnier:
I'm using version 0.1.1 (latest)

stewrutledgebonnier:
Could it potentially be related to https://github.com/go-redis/redis/issues/832?

FUSAKLA:
Ok, possibly try to use the BoltDB which could sort out the redis. Also what version of redis do you use?

stewrutledgebonnier:
So far using the in memory cache has less of a memory footprint than using redis.

SuperQ:
Here's a profile on BoltDB
[heap.svg.gz](https://github.com/Comcast/trickster/files/2511537/heap.svg.gz)

SuperQ:
@juliusv and I worked up a patch that closes the cache goroutines after they've been idle for 10 seconds. This seems to help greatly with our setup.

https://github.com/juliusv/trickster/commit/b2b82c862932282a32300b4ba279d62500642250

https://dashboards.gitlab.com/dashboard/snapshot/pvDNQsfFnzc9pxE6uuTZ0FNUG2BpjA7j

We went from several hundred thousand open FDs and goroutines and GB of memory to under 1k goroutines and a few hundred MB of memory.

SuperQ:
Yup, I'll get the new version into testing soon.

SuperQ:
I've started 0.1.4 in production. Will see how it behaves over the next day or so.

SuperQ:
Getting a panic:

```
panic: runtime error: index out of range
goroutine 1815 [running]:
main.(*PrometheusMatrixEnvelope).cropToRange(0xc000253238, 0x5c07e3b2, 0x0)
#011/Users/jranso000/code/go/src/github.com/Comcast/trickster/handlers.go:1035 +0x4ee
[...]

PR #118, opened 2019-01-26
Refactor channel reap to use Bulk Get call(Issue #4)

arsenakishev:
The original redis cache GET call within the redis ReapOnce function has been modified to get all the keys with a single bulk call using the redis client MGET method.

Also, I noticed the reaper goroutine is **not started** within the redis.Connect function. This type of goroutine call for the reaper exists both in the memory.go file and the boltdb.go file. I have added this call to the redis.Connect function.

### DIFF
--- a/redis.go
+++ b/redis.go
@@ -39,6 +39,7 @@ func (r *RedisCache) Connect() error {
 	if r.Config.Password != "" {
 		r.client.Options().Password = r.Config.Password
 	}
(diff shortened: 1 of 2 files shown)


jranson:
@arsenakishev Thanks Arsen, I will review and merge this weekend. Regarding the defect you found with the reaper not starting for Redis, correcting that may address #91.

PR #119, opened 2019-01-30
fix max_value_age_secs config doesn't work

qwertwwwe:
The max_value_age_secs config in origin doesn't work, Trickster will not drop any old points based on retention policy.

I noticed that in handlers.go:860, in the code use ctx.Matrix.cropToRange(ctx.Time-ctx.Origin.MaxValueAgeSecs, 0), the start=ctx.Time-ctx.Origin.MaxValueAgeSecs use seconds, but the timestamp in metrics from Prometheus use milliseconds, which cause the ts always >= start in cropToRange function, so old data will never be dropped.

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -857,7 +857,7 @@ func (t *TricksterHandler) originRangeProxyHandler(cacheKey string, originRangeR
 			}
 
 			// Prune any old points based on retention policy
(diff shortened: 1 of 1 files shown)


jranson:
@qwertwwwe Thank you for finding and fixing this! Some number of months back, we switched from Secs to Msecs for the timestamps, and it looks like this one just slipped through the cracks as part of the refactor. Appreciate the help!

PR #120, opened 2019-02-01
fix the max_value_age_secs works on data return to user.

qwertwwwe:
In the use, I find the max_value_age_secs works both on cache and data returned to user. I set the max_value_age_secs to 86400(1day), and I use Grafana to query for 7 days' data, Trickster will only return 1 day's data to Grafana. I think maybe the max_value_age_secs should only works on cached data, not crop the data return to user, so I push a pull merge request, I don't know is this ok for this question.


arnitolog:
I'm also hit the same issue. Trickster v0.1.7 returns data for the last 24h only. v0.1.6 works as expected and returns whole requested range

PR #133, opened 2019-02-26
fix fastforward time ms bug

qwertwwwe:
In the testing, I found in the FastForward compare, ctx.RequestExtents.End < ctx.Time-ctx.StepMS, ctx.Time use seconds and others use milliseconds, this merge request fix this bug and the corresponding unit test.

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -800,7 +800,7 @@ func (t *TricksterHandler) originRangeProxyHandler(cacheKey string, originRangeR
 				}()
 			}
 
-			if !ctx.Origin.FastForwardDisable && !(ctx.RequestExtents.End < ctx.Time-ctx.StepMS) {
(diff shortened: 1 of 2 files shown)


handlers.go
```diff
@@ -800,7 +800,7 @@ func (t *TricksterHandler) originRangeProxyHandler(cacheKey string, originRangeR
 				}()
 			}
 
-			if !ctx.Origin.FastForwardDisable && !(ctx.RequestExtents.End < ctx.Time-ctx.StepMS) {
+			if !ctx.Origin.FastForwardDisable && !(ctx.RequestExtents.End < ctx.Time*1000-ctx.StepMS) {
```
sylr: Nit: use Time.Millisecond instead of 1000 for better understanding.

PR #147, opened 2019-03-11
Relocate example conf

jranson:


jranson:
The Makefile is going to need lots of work for 1.0 and i have an issue separate for that (#148). Update on the doc is incoming. Thanks!

PR #149, opened 2019-03-11
Restore Snappy Compression for Cache

### LINKED ISSUE #137: Snappy Compression on Cache Read/Write

jranson:
Add snappy compression support back into cache. Fixes #137


internal/proxy/cache.go
```diff
@@ -32,6 +37,18 @@ func QueryCache(c cache.Cache, key string) (*HTTPDocument, error) {
 	if err != nil {
 		return d, err
 	}
+
+	// Decompress if it's not JSON. Because the Compression configuration may have changed between the time
```
SuperQ: Maybe check for the existence of the [snappy header magic string](https://github.com/google/snappy/blob/master/framing_format.txt). `sNaPpY`

jranson: Awesome, thanks Ben. I had checked the Wikipedia (which is supposed to know everything) Article for Snappy and it did not provide a magic string. I'll change that out in just a bit.

SuperQ: Yea, the individual frames don't have a magic string, but supposedly the "stream" does. I guess it depends on how the data is written out. Probably good to test.

jranson: In testing and checking the source for [snappy.Encode()](https://github.com/golang/snappy/blob/master/encode.go#L18), the Magic String does not appear to be used used for block encodes. I tried prepending it to encoded blob, but it would not properly decompress without first removing that magic string. I think I am ok with that approach (adding `sNaPpY` on store and checking/removing on retrieve) to positively check for compression. Any thoughts?

SuperQ: Apparently that library only implements snappy block format, not stream format.

Another way to deal with this is to change the filenames, and check for the existence of `item.json` or `item.sz`. This would be more explicit.

PR #155, opened 2019-03-13
Update helm and kube config maps to follow new format, update versions.

LimitlessEarth:
### Changes of note
- Helm 2.8.2 -> 2.9.1
- kubectl 1.9.4 -> 1.13.4
- Minikube 0.25.0 -> 0.35.0
- Kube and Helm config maps updated to new format.
- Helm config map is now templated to range over origins and caches in values.

A thought I had while doing this is that we could put the comments of what each thing is in [values.yaml](https://github.com/helm/charts/blob/master/stable/prometheus/values.yaml) in stead of the [configmap](https://github.com/helm/charts/blob/master/stable/prometheus/templates/server-configmap.yaml) like prometheus does in their chart. Thoughts?

### DIFF
--- a/deploy/helm/trickster/templates/configmap.yaml
+++ b/deploy/helm/trickster/templates/configmap.yaml
@@ -10,105 +10,191 @@ metadata:
 data:
   trickster.conf: |-
     [main]
-
-        # instance_id allows you to run multiple trickster processes on the same host and log to separate files
-        # Useful for baremetal, not so much for elastic deployments, so only uncomment if you really need it
-        #instance_id = 1
+    # instance_id allows you to run multiple trickster processes on the same host and log to separate files
+    # Useful for baremetal, not so much for elastic deployments, so only uncomment if you really need it
+    #instance_id = 1
 
     # Configuration options for the Proxy Server
     [proxy_server]
+    # listen_port defines the port on which Trickster's Proxy server listens.
+    # The default proxy is for Prometheus, so we use 9090 by default, just like Prometheus does
+    # listen_port = 9090
+    listen_port = {{ .Values.service.port }}
+    # listen_address defines the ip on which Trickster's Proxy server listens.
+    # empty by default, listening on all interfaces
+    # listen_address =
+    {{ if .Values.service.metricsAddress }}
+    listen_address = {{ .Values.service.address }}
+    {{ end }}
 
-        # listen_port defines the port on which Trickster's Proxy server listens.
-        # since this is a proxy for Prometheus, we use 9090 by default, just like Prometheus does
-        listen_port = {{ .Values.service.port }}
 
-    [cache]
+    [caches]
+
+        # [caches.default]
         # cache_type defines what kind of cache Trickster uses
-        # options are 'boltdb', 'filesystem', 'memory', and 'redis'.  'memory' is the default
-        cache_type = {{ .Values.cache.type | quote }}
+        # options are 'bbolt', 'filesystem', 'memory', and 'redis'.
+        # The default is 'memory'.
+        # type = 'memory'
 
-        # record_ttl_secs defines the relative expiration of cached queries. default is 6 hours (21600 seconds)
-        record_ttl_secs = {{ .Values.recordTTLSecs }}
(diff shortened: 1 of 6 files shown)


deploy/helm/trickster/templates/configmap.yaml
```diff
@@ -10,105 +10,191 @@ metadata:
...
     [proxy_server]
+    # listen_port defines the port on which Trickster's Proxy server listens.
+    # The default proxy is for Prometheus, so we use 9090 by default, just like Prometheus does
+    # listen_port = 9090
+    listen_port = {{ .Values.service.port }}
+    # listen_address defines the ip on which Trickster's Proxy server listens.
+    # empty by default, listening on all interfaces
+    # listen_address =
+    {{ if .Values.service.metricsAddress }}
+    listen_address = {{ .Values.service.address }}
```
crandles: The naming of these two variables is confusing: `service.metricsAddress` and `service.address` -- I think you may have intended for them to be the same variable?

deploy/helm/trickster/templates/configmap.yaml
```diff
@@ -10,105 +10,191 @@ metadata:
...
-            origin_url = {{ .Values.originURL | quote }}
-
-            # api path defines the path of the Prometheus API (usually '/api/v1')
-            api_path = '/api/v1'
+            # bucket = 'trickster'
+
+      {{- range .Values.caches }}
+        {{ printf "[caches.%s]" .name }}
+        type = {{ .type | quote }}
+        record_ttl_secs = {{ .recordTTLSecs }}
```
crandles: you may want to pass this through the `default` function:
` .recordTTLSecs | default(21600)`

Since this is an object, it is possible someone forgot this key.

deploy/helm/trickster/values.yaml
```diff
@@ -36,7 +54,9 @@ image:
 service:
   type: ClusterIP
   port: 9090
+  #address = "localhost"
```
crandles: These (address/metricsAddress) aren't used to define the service, but for configuring the `listen` address for trickster. 

I'm not sure how relevant they are to the helm chart deployment -- do you see this changing at all? I don't see why it would be overridden.

I'm thinking you might want to:
- specify the ClusterIP here instead (that could get referenced in the service ~~and in the config instead of `address`~~)
- or, at deployment time you should get the pod IP in an env variable, if trickster could use that via the cli you could listen to a single IP
- or, not expose this? I'm not sure its useful if you can't specify the clusterIP.

crandles: Thinking about this further -- I don't think you'd want to listen on the service address, but I could be mistaken. The pod IP could make sense.

crandles:
Comments in the values.yaml are preferred over the configmap, I think. You can always link to trickster docs for anything too complex.

deploy/helm/trickster/templates/configmap.yaml
LimitlessEarth: There are two separate configurable values:
https://github.com/Comcast/trickster/blob/next/cmd/trickster/conf/example.conf#L13
https://github.com/Comcast/trickster/blob/next/cmd/trickster/conf/example.conf#L131

That are used differently in the code:
https://github.com/Comcast/trickster/blob/next/internal/config/config.go#L107
https://github.com/Comcast/trickster/blob/next/internal/config/config.go#L123

deploy/helm/trickster/values.yaml
LimitlessEarth: No, I think you are right. I don't think you would ever want to change the addr in the pod itself, just the clusterIP. Even at that I don't think I usually ever want to.

deploy/helm/trickster/templates/configmap.yaml
LimitlessEarth: This does remain true that these values are separate but this conversation is not relevant given our service discussion below.

crandles:
Two more things:
1. You'll want to update the version/appVersion in the chart.yaml
2. Not sure if you want to update docker image version now

PR #156, opened 2019-03-14
Labweek trickster latest

srijeet0406:


internal/proxy/influxdb/influxdb.go
```diff
@@ -70,23 +106,209 @@ func (c Client) OriginName() string {
...
+	numValue, err := (strconv.ParseInt(number[0], 10, 64))
+	if (err != nil) {
+		panic(err)
+	}
+	timeValue := numValue * multiplier.Nanoseconds()
+	return timeValue
+}
+
+func getTimeValueForQueriesWithNow(timeParsed []string) (int64, string) {
+	suffix := strings.SplitAfterN(timeParsed[0], "now()", 2)
```
jranson: I would recommend putting a check here for the length of timeParsed and abort the method (e.g., by returning a new error param) if the array is empty. Currently if timeParsed is empty for any reason, this method would panic the app.

internal/proxy/influxdb/influxdb.go
```diff
@@ -70,23 +106,209 @@ func (c Client) OriginName() string {
...
+	case 'd':
+		multiplier = 24 * time.Hour
+	case 'h':
+		multiplier = time.Hour
+	case 'm':
+		multiplier = time.Minute
+	case 's':
+		multiplier = time.Second
+	default:
+		multiplier = time.Nanosecond
```
jranson: since the value is already set to time.Nanosecond at instantiation, this switch will unnecessarily set the value of `multiplier` a second time. Changing the instantiation to be `var multiplier time.Duration` will eliminate an unnecessary double-set of the variable.

PR #162, opened 2019-03-23
1.0 fastforward

jranson:


internal/proxy/deltaproxycache.go
```diff
@@ -77,80 +105,149 @@ func DeltaProxyCacheRequest(r *Request, w http.ResponseWriter, client Client, ca
...
 
-	// TODO: Fast Forward Here, Another wg.Add(1), go func, defer wg.Done(), fetch, Matrix from Vector, lock, defer unlock, append to mts
+	var ffts timeseries.Timeseries
+	var ffex timeseries.Extent
+	// Only fast forward if configured and the user request is for the absolute latest datapoint
+	if (!cfg.FastForwardDisable) && (trq.Extent.End == normalizedNow.Extent.End) {
+		ffURL, err := client.FastForwardURL(r)
+		if ffURL.Scheme != "" && err == nil {
+			wg.Add(1)
+			go func() {
```
LimitlessEarth: Is there any chance this will block forever? Does having a timeout on this make sense?

internal/proxy/deltaproxycache.go
```diff
@@ -77,80 +105,149 @@ func DeltaProxyCacheRequest(r *Request, w http.ResponseWriter, client Client, ca
...
+	if !cfg.FastForwardDisable && ffts != nil && ffex.End.After(trq.Extent.End) {
+		rts.Merge(false, ffts)
+	}
+	rdata, err := client.MarshalTimeseries(rts)
+
+	// Don't write the cache unless it has chnaged
+	if cacheStatus != crHit {
+		wg.Add(1)
+		// Write the newly-merged object back to the cache
+		go func() {
```
LimitlessEarth: Is there any chance this will block forever? Does having a timeout on this make sense?

internal/proxy/deltaproxycache.go
jranson: Should be set now on that, the client timeout is a feature in 0.1.x that i neglected to port over, so thank you :)  It is now passed all the way through to proxy.Fetch() on a per-origin basis.

Issue #142, opened 2019-03-11
Support Max Cache Size setting


jranson:
The community has requested a size-based retention policy setting. Setting should be represented in bytes, where 0 = infinite size. This is addition to existing Time-based cache retention and the cache will maintain the cache using both setting in tandem.

jranson:
Completed in #164

PR #164, opened 2019-03-24
1.0 cache stats

jranson:
Provides Cache-based metrics (#136) and size-based cache retention (#142)


internal/cache/bbolt/bbolt.go
```diff
@@ -58,182 +58,109 @@ func (c *Cache) Connect() error {
...
 
 // Store places an object in the cache using the specified key and ttl
 func (c *Cache) Store(cacheKey string, data []byte, ttl int64) error {
+	return c.store(cacheKey, data, ttl, true)
+}
 
-	expKey, dataKey := c.getKeyNames(cacheKey)
-	expiration := []byte(strconv.FormatInt(time.Now().Unix()+ttl, 10))
+func (c *Cache) storeNoIndex(cacheKey string, data []byte) {
+	c.store(cacheKey, data, 31536000, false)
```
LimitlessEarth: Ignore the error here?

internal/cache/filesystem/filesystem.go
```diff
@@ -44,95 +45,105 @@ func (c *Cache) Configuration() *config.CachingConfig {
...
-
-	if err1 != nil {
-		return err1
-	} else if err2 != nil {
-		return err2
+	return c.store(cacheKey, data, ttl, true)
+}
+
+func (c *Cache) storeNoIndex(cacheKey string, data []byte) {
+	c.store(cacheKey, data, 31536000, false)
```
LimitlessEarth: Ignore the error?

internal/cache/index.go
```diff
@@ -0,0 +1,336 @@
...
+		defer indexLock.Unlock()
+	}
+	if o, ok := idx.Objects[key]; ok {
+		idx.CacheSize -= o.Size
+		idx.ObjectCount--
+
+		ObserveCacheOperation(idx.name, idx.cacheType, "del", "none", float64(o.Size))
+
+		delete(idx.Objects, key)
+		ObserveCacheSizeChange(idx.name, idx.cacheType, idx.CacheSize, idx.ObjectCount, idx.config.MaxSizeObjects, idx.config.MaxSizeBytes)
```
LimitlessEarth: Do you ever expect this to change after a cache has been initialized? You could move this to another metric to avoid setting a static value again and again. Then again I don't know how Prometheus storage works, it could very well be the use a common copy.

cmd/trickster/conf/example.conf
```diff
@@ -112,7 +154,7 @@
 
     # [origins.foo]
     # type = 'influxdb'
-    # cache_name = default
+    # cache_name = 'bbolt'
```
LimitlessEarth: origin name of bbolt?

internal/util/metrics/metrics.go
```diff
@@ -62,13 +83,76 @@ func Init() {
 		[]string{"origin", "origin_type", "method", "status", "http_status", "path"},
 	)
 
+	CacheObjectOperations = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "trickster_cache_operations_objects",
+			Help: "Count of operations performed on a Trickster cache.",
+		},
+		[]string{"cache", "type", "operation", "status"},
```
SuperQ: Let's avoid `type` as a label, it's a bit generic and can cause conflicts with discovery. How about `cache_name`, and `cache_type` as the first two labels on these metrics.

internal/util/metrics/metrics.go
```diff
@@ -62,13 +83,76 @@ func Init() {
 		[]string{"origin", "origin_type", "method", "status", "http_status", "path"},
 	)
 
+	CacheObjectOperations = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "trickster_cache_operations_objects",
```
SuperQ: All `Counter` type metrics should end in `_total`. How about `trickster_cache_object_operations_total`.

I might also define a couple of consts in the package. One for the Namespace, and one for the subsystem.

```go
const (
  metricNamespace = "trickster"
  metricSubsystem = "cache"
)
...

prometheus.CounterOpts{
  Namespace: metricNamespace,
  Subsystem: metricSubsystem,
  Name: "object_operations_total",
...
```

internal/cache/index.go
jranson: Anyone can adjust the max cache size in their config and bounce Trickster, so we need to continually monitor it in order to properly report cache utilization / capacity.

internal/cache/bbolt/bbolt.go
jranson: updated to handle the error. thanks!

internal/cache/filesystem/filesystem.go
jranson: updated

internal/cache/index.go
LimitlessEarth: The alternative I was think of it setting it in its own separate metric to avoid including this information on every single unique label set for this metric. Just set it once and you are done.

jranson: I got you. The maxes are already their own metric (the observe func just aggregates the calls). I've moved the two max metrics out of observe and over to the NewIndex(), so they'll only be set once. thanks for the greaet catch!

internal/util/metrics/metrics.go
```diff
@@ -25,50 +25,160 @@ import (
...
+			Help:      "Count (in # of objects) of operations performed on a Trickster cache.",
+		},
+		[]string{"cache_name", "cache_type", "operation", "status"},
+	)
+
+	CacheByteOperations = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Namespace: metricNamespace,
+			Subsystem: cacheSubsystem,
+			Name:      "byte_operations_total",
```
SuperQ: If this is a byte counter, it should be `operation_bytes_total`. This makes it sound like it's a counter of operations.

internal/util/metrics/metrics.go
```diff
@@ -25,50 +25,160 @@ import (
...
+		},
+		[]string{"cache_name", "cache_type", "event", "reason"},
+	)
+
+	CacheObjects = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: metricNamespace,
+			Subsystem: cacheSubsystem,
+			Name:      "usage_objects",
+			Help:      "Count of objects in a Trickster cache.",
```
SuperQ: If it's a gauge, I would say "Number of objects" rather than "Count".

internal/cache/cache.go
```diff
@@ -16,67 +16,29 @@ package cache
...
-	Reap()
+	Remove(cacheKey string)
+	BulkRemove(cacheKeys []string, noLock bool)
 	Close() error
 	Configuration() *config.CachingConfig
 }
+
+// ObserveCacheMiss returns a standard Cache Miss response
+func ObserveCacheMiss(cacheKey, cacheName, cacheType string) ([]byte, error) {
+	ObserveCacheOperation(cacheName, cacheType, "get", "miss", 0)
```
SuperQ: Typically I like to use a separate metric for cache misses, it makes it a little easier to do queries like cache miss ratios.

For example:
```
sum(rate(trickster_cache_misses_total[5m])) / sum(rate(trickster_cache_operations_total[5m]))
```

Having hit vs miss means you have to use label selections. But this also means it would probably mean moving the `operation` label to be separate metric names as you still need to select `get` operations for this query to be valid.

PR #166, opened 2019-03-26
Support Env Var and CLI Flag for OriginType

jranson:


internal/config/flags_test.go
```diff
@@ -26,14 +26,19 @@ func TestLoadFlags(t *testing.T) {
 		"9091",
 		"-metrics-port",
 		"9092",
+		"-origin-type",
+		"prometheus",
```
crandles: do you want to test the origin url flag?

jranson: origin test is already present higher up in the test

PR #170, opened 2019-03-28
Redis Cluster Support

jranson:


cmd/trickster/conf/example.conf
```diff
@@ -67,6 +67,15 @@ listen_port = 9090
...
         password = ''
 
+        ### Configuration options when using Redis Cluster
+        [caches.default.redis_cluster]
+        # endpoints defines a list of seed fqdn+ports for the cluster
+        # default is ['redis:6379']
+        endpoints = ['redis:6379']
+        # password provides the redis cluster password
+        # default is empty
+        password = ''
```
LimitlessEarth: Is this also an array?

jranson: For Redis Cluster, an array of endpoints is provided. For Redis (non-cluster) it is a string for a single endpoint.

PR #176, opened 2019-03-29
Go 1.12.1

sylr:


Makefile
```diff
@@ -1,21 +1,31 @@
...
-GO           ?= go
-GOFMT        ?= $(GO)fmt
-FIRST_GOPATH := $(firstword $(subst :, ,$(shell $(GO) env GOPATH)))
+GO             ?= go
+GOFMT          ?= $(GO)fmt
+FIRST_GOPATH   := $(firstword $(subst :, ,$(shell $(GO) env GOPATH)))
 TRICKSTER_MAIN := cmd/trickster
-DEP          := $(FIRST_GOPATH)/bin/dep
-TRICKSTER    := $(FIRST_GOPATH)/bin/trickster
+DEP            := $(FIRST_GOPATH)/bin/dep
```
jranson: Do we still need the `DEP` var since we are using Modules?

Makefile
```diff
@@ -1,21 +1,30 @@
...
-	GO111MODULE=on $(GO) mod vendor
+	$(GO) mod vendor
+
+.PHONY: go-mod-tidy
+go-mod-tidy:
+	$(GO) mod tidy
+
+.PHONY: test-go-mod
+test-go-mod: go-mod-tidy go-mod-vendor
+	@git diff --quiet --exit-code go.mod go.sum || echo "There are changes to go.mod and go.sum which needs to be commited"
```
jranson: TravisCI is failing because of the typo here on the word committed. If you can fix that it should pass all the checks. Thanks!

PR #177, opened 2019-03-29
Add support for redis cluster configuration in kube artifacts and helm chart

LimitlessEarth:
Follows along with #170

### DIFF
--- a/deploy/helm/trickster/templates/configmap.yaml
+++ b/deploy/helm/trickster/templates/configmap.yaml
@@ -131,9 +131,9 @@ data:
             endpoints = {{ .redis.endpoint | quote }}
             password = {{ .redis.password | quote }}
 
-            {{- if eq .type "redis_cluster" }}
+            {{- else if eq .type "redis_cluster" }}
             {{ printf "[caches.%s.redis_cluster]" .name }}
-            endpoint = {{ .redis_cluster.endpoints }}
+            endpoints = {{ .redis_cluster.endpoints }}
             password = {{ .redis_cluster.password | quote }}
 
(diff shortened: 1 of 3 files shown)


deploy/helm/trickster/templates/configmap.yaml
```diff
@@ -119,9 +128,14 @@ data:
             {{- if eq .type "redis" }}
             {{ printf "[caches.%s.redis]" .name }}
             protocol = {{ .redis.protocol | quote }}
-            endpoint = {{ .redis.endpoint | quote }}
+            endpoints = {{ .redis.endpoint | quote }}
             password = {{ .redis.password | quote }}
 
+            {{- if eq .type "redis_cluster" }}
+            {{ printf "[caches.%s.redis_cluster]" .name }}
+            endpoint = {{ .redis_cluster.endpoints }}
```
jranson: you're using `endpoints =` on line 131 but `endpoint =` on 136. Just want to make sure this is intentional, as on the surface it appears to be breaking.

deploy/helm/trickster/templates/configmap.yaml
```diff
@@ -119,9 +128,14 @@ data:
             {{- if eq .type "redis" }}
             {{ printf "[caches.%s.redis]" .name }}
             protocol = {{ .redis.protocol | quote }}
-            endpoint = {{ .redis.endpoint | quote }}
+            endpoints = {{ .redis.endpoint | quote }}
             password = {{ .redis.password | quote }}
 
+            {{- if eq .type "redis_cluster" }}
```
crandles: this should be `else if`

PR #189, opened 2019-04-08
Update helm chart example

dominikhahn:
Update helm chart example


dominikhahn:
@jranson just saw that you're already tuning the helm charts on branch _next_.
this pr is tested against the current master branch and i could also contrib further on _next_.

LimitlessEarth:
I agree with Chris. I believe `next` will become master sooner than later, your changes will be better off in that branch. So we should either adapt these changes into next or however we want to solve that.

jranson:
Approved for  `master` branch, lets implement in `next` under a separate PR

PR #194, opened 2019-04-10
1.0 - refactor Step as time.Duration throughout

jranson:


jranson:
(i'll address the Testing OOM today and ping you when it's passing)

PR #198, opened 2019-04-11
More conservative data fetching for partial cache hit

### LINKED ISSUE #131: Performance Problem: Some times Trickster request too much time range than needed.
In our use case, some times Trickster will request more time range than needed. For example, I set the MaxValueAgeSecs=86400(24h). The first request time range is 2019.01.29-2019.01.30, the second request time range is 2019.01.01-2019.01.02. In the second request, Trickster will send request to prometheus with time range 2019.01.01-2019.01.29, some times this length of time range will cause timeout in query. If trickster optimize this case, I think this may improve a lot of performance in some query cases.

jacksontj:
The old behavior would query from the end of the requested data until
the beginning of the cached data. So as an example, if you did a query for the last 30m and then did a query for a 30m window a month ago-- that second query would attempt to fetch all data from a month ago until now-- which isn't what we want (and in practice it fails).

This instead changes the end of the query range to be the normalized end
of the query-range

Fixes #131

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -492,7 +492,6 @@ func (t *TricksterHandler) buildRequestContext(w http.ResponseWriter, r *http.Re
 	if err != nil {
 		return nil, errors.Wrap(err, "error aligning step boundary")
 	}
-
 	// setup some variables to determine and track the status of the query vs what's in the cache
 	ctx.Matrix = defaultPrometheusMatrixEnvelope()
 	ctx.CacheLookupResult = crKeyMiss
@@ -500,6 +499,13 @@ func (t *TricksterHandler) buildRequestContext(w http.ResponseWriter, r *http.Re
 	// parameters for filling gap on the upper bound
 	ctx.OriginUpperExtents.Start = ctx.RequestExtents.Start
 	ctx.OriginUpperExtents.End = ctx.RequestExtents.End
+
+	// If the entire request is outside of the MaxValueAgeSecs, then lets not look in the
+	// cache, we won't have it
+	if (ctx.Time*1000 - ctx.RequestExtents.End) > ctx.Origin.MaxValueAgeSecs*1000 {
+		ctx.CacheLookupResult = crRangeMiss
+		return ctx, nil
+	}
 
 	// Get the cached result set if present
 	cachedBody, err := t.Cacher.Retrieve(ctx.CacheKey)
@@ -858,8 +864,12 @@ func (t *TricksterHandler) originRangeProxyHandler(cacheKey string, originRangeR
 				ctx.Matrix = t.mergeMatrix(upperDeltaData, ctx.Matrix)
 			}
 
+			// If the request is entirely outside of the cache window, we don't want to cache it
+			// otherwise we actually *clear* the cache of any data it has in it!
+			skipCache := (ctx.Time*1000 - ctx.RequestExtents.End) > ctx.Origin.MaxValueAgeSecs*1000
+
 			// If it's not a full cache hit, we want to write this back to the cache
-			if ctx.CacheLookupResult != crHit {
+			if ctx.CacheLookupResult != crHit && !skipCache {
 				cacheMatrix := ctx.Matrix.copy()
 
 				// Prune any old points based on retention policy


jacksontj:
cc @jranson IMO this is worth merging in. I'm not sure how people didn't hit this before, but this is actually quite easy to repro :scream:

jranson:
@jacksontj Unfortunately I do not think it is going to be as easy as this commit.

Heres the long explanation: 
The extents of a cached dataset in 0.1.x are calculated by simply looking at the Start and End times of the overall dataset (https://github.com/Comcast/trickster/blob/master/handlers.go#L1087). That is why gaps can't be handled, there's no mechanism to track those gaps, and that's why the upstream proxy request is extended to fill gaps to avoid them completely.

So this basically change inserts untracked gaps into the cached dataset that should result in data inaccuracies, in the form of large cached gaps.

Imagine you are on a 6-hour view in Grafana and have the last 6 hours in cache, then 'rollback' to a day ago, so your patch fetches and inserts into the cache the 6-hour window from yesterday into the cache. So there should be an 18-hour data gap in the middle of the newly-cached section and the originally-cached section. 

What I would predict would happen in this case (which is the exact case you are trying to solve) is that any subsequent requests for a timerange in the new gap area (having the same step as the cached range) will return an empty dataset as a cache hit, instead of fetching the missing data from the origin. The reason is because the extent calculation will indicate the range is cached because checking the start and ends will say so.

jacksontj:
> The extents of a cached dataset in 0.1.x are calculated by simply looking at the Start and End times of the overall dataset

This means the overall dataset it has ever seen? meaning if I query today, a month ago, and a year ago-- an entire year? If so then I'm really not sure how we didn't run into this before :/

jranson:
Unfortunately yes -- this particular usecase seemed like a rare-enough one when compared to the vast majority (dashboards just refreshing most recent data constantly) where Trickster helps, that it took a while for me to get back to address it. I am glad it's fully fixed in 1.0.!

To get the gap capabilities working in 0.1.x, not only would we have to create some ability to track gaps, but we'd also have to augment the upstream proxy client count from functional max of 2 (+1 for FF) to instead equal the number of gaps that are present in the requested dataset. That is effectively what 1.0 does, and I think we are less than a month from its release. So I do not know if it is worth all of that effort to get 0.1.x doing what we promised for and have delivered in 1.0.

Here is a potential alternative fix for the 0.1.x codebase to help: Right now, even if all of the newly-requested data is outside of the retention policy, Trickster 0.1.x will still request the entire block including any gap fill from upstream, despite the fact that none of the requested data will ever be cached. That is certainly painful and avoidable. We can definitely optimize that by checking, _before_ making the upstream request (and only on a "range miss" when this particular condition is triggered), if the requested dataset's timerange is fully outside of the retention policy.  If it is, we can simply proxy the request rather than going through all of the normal delta merging caching cropping hoopla. That quick fix should make the problem _much_ less noticeable, esp when trickster's retention policy and/or the query's step resolution are low.

jacksontj:
@jranson ok, here is a (hopefully) more workable patch.

So assuming the extent has to be the entire `MaxValueAgeSecs` -- now you will *at most* query that entire range. What I've done is made the cache handler check if the request is outside of the window -- if so immediately count it as a cache miss (the cache shouldn't have it, and this avoids all the window changing). Similarly when the response is going to be written to the cache I check if the data is within the window, if not we don't cache it. This way all queries that are older than `MaxValueAgeSecs` will effectively bypass the cache.

jacksontj:
@jranson ok, here is a (hopefully) more workable patch.

So assuming the extent has to be the entire `MaxValueAgeSecs` -- now you will *at most* query that entire range. What I've done is made the cache handler check if the request is outside of the window -- if so immediately count it as a cache miss (the cache shouldn't have it, and this avoids all the window changing). Similarly when the response is going to be written to the cache I check if the data is within the window, if not we don't cache it. This way all queries that are older than `MaxValueAgeSecs` will effectively bypass the cache.

jacksontj:
bump @jranson in case it was missed, I think I have implemented what you described, let me know if that works :)

jranson:
On the Surface this looks good, thanks for taking the time to do this! I will give it a thorough vetting this evening.

jacksontj:
sounds good!

PR #204, opened 2019-04-13
get coveralls to read full coverage profile

jranson:


jranson:
Appears to be working as expected; merging

Issue #208, opened 2019-04-18
Trickster cannot display points in the future


ahurtaud:
Hello,

When using "today" timerange view in Grafana. Trickster cannot display datapoints in the future.

The usecase we have is, when we want to compare last week traffic with current traffic. We use the [offset modifier](https://prometheus.io/docs/prometheus/latest/querying/basics/#offset-modifier) from prometheus. 

When querying prometheus:
![OffsetFutureWithPrometheus](https://user-images.githubusercontent.com/25080995/56343617-89c9e780-61bb-11e9-98c4-9ea6f084e8aa.png)

When querying trickster:
![OffsetFutureWithTrickster](https://user-images.githubusercontent.com/25080995/56343628-92222280-61bb-11e9-8ea3-de7c2dd05622.png)

The query B (in blue) is query A offset 1w. (the issue is valid for any offset value (even 5minutes)
E.g: 
query A: rate(msg_received_total[5m])
query B: rate(msg_received_total[5m] offset 1w)

jranson:
@ahurtaud Thanks to Alban for reporting this issue and also demonstrating this very useful Grafana feature. This issue affects 0.1.x (won't fix) and 1.0 (fix in progress), and occurs where we extract the client-requested time range from a cached dataset. Since the dates of these points can be in the future (due to the way Grafana constructs time ranges for the "Today" view), they are being cropped by Trickster before being returned to the client.

jranson:
Separately, in debugging this issue, I identified a defect in Grafana's "Today" feature where it is incompatible with Prometheus vector requests. [Grafana #16661](https://github.com/grafana/grafana/issues/16661)

PR #210, opened 2019-04-21
allow future times when the query is offset

jranson:


jranson:
(Fixes #208)

internal/cache/redis/redis_test.go
```diff
@@ -1,5 +1,5 @@
 /**
-* Copyright 2018 Comcast Cable Communications Management, LLC
+* Copyright 2018 Corcast Cable Communications Management, LLC
```
LimitlessEarth: Oops

Issue #214, opened 2019-04-23
[question] does trickster supports many instances use single redis cluster as backend?


Wing924:
Hello!
1) To keep HA, I setup many trickster instances with same config. Can I use same redis cluster as backend? If it works, does they share the same cache? (ex. trickster 2 can read the data cached by trickster 1)
```
                    /-> [trickster 1] -\
                   |         \----------|-----------------\
[grafana] --> [LB] -+-->[trickster 2] --+-> [prometheus]  |
                   |         \----------|-----------\     |
                    \-> [trickster 3] -/             |    |
                             |                       |    |
                             \--> [redis cluster] <-/     |
                                         ^                |
                                          \--------------/
```

2) To scale-out prometheus, I setup many trickster and prometheus instances (each trickster connect to different prometheus and each prometheus scrape different targets). Can I use same redis cluster as backend?
```
            /-> [datasource 1] ---> [trickster 1] ---> [prometheus 1]
           |                             \------------------------------\
[grafana] -+--> [datasource 2] ---> [trickster 2] ---> [prometheus 2]    |
           |                             \----------------------------\  |
            \-> [datasource 3] ---> [trickster 3] ---> [prometheus 3]  | |
                                         |                             | |
                                          \->  [redis cluster] <------/  |
                                                     ^                   |
                                                      \-----------------/
```

jranson:
Hi Wei, thanks for trying out Trickster and for asking about this. The answer is *yes* to both questions; in fact, the reason for supporting Redis is to facilitate these exact scenarios. Trickster uses the upstream prometheus hostname as part of the cache/redis key name, so each prometheus's data is cached separately and will encounter collisions. As I write this, I checked and noticed the hostname component did not make it into the Trickster 1.0 branch, I will add a PR for it right away.

PR #215, opened 2019-04-23
use upstream hostname in cache key name

jranson:


jranson:
Fixes the issue i mentioned in #214

PR #196, opened 2019-04-10
Add cors support for queries like series

Dudesons:
Hi,

This pr fixes a problem on grafana when you want to have dashboard with parameters.
The request send by grafana for that is something like that: `https://trickster.veepeez.io/api/v1/series?match[]=probe_success&start=1554912065&end=1554913865`
but grafana can't display the result.

in the browser console I have the following message:
Screenshot 2019-04-10 at 18 57 05

@trickster-reviewers

Issue #216, opened 2019-04-24
Content-Type: application/x-gzip


mozai:
Strange behaviour by Trickster, I can reproduce it where I am, and co-workers see it too, but not sure how to recreate it in a lab.

When using Trickster 1.0.4 as a data source with Grafana, Trickster is responding with strange http headers:  "Content-Encoding: gzip" and "Content-Type: application/x-gzip".  Shouldn't that be "Content-Type: application/json" ?

```
GET /api/v1/query_range?query=node_load5%7Brole%3D%22tls_alerts%22%2Cip%3D~%22.*%22%7D&start=1556133300&end=1556136960&step=60&timeout=1m HTTP/1.1
Host: gis-trickster-mog.[redacted]:9090
User-Agent: Grafana/5.4.3
Accept: application/json, text/plain, */*
Accept-Encoding: gzip, deflate, br
Accept-Language: en-CA,en-US;q=0.7,en;q=0.3
X-Forwarded-For: 10.137.16.75, 127.0.0.1, 127.0.0.1
X-Forwarded-Server: picasso.[redacted]
X-Grafana-Org-Id: 1

20:14:16.451737 IP gis-trickster-mog.[redacted].9090 > tls-oms-tsg.[redacted].33792: Flags [P.], seq 7261:8083, ack 4684, win 82, options [nop,nop,TS val 17771021 ecr 3234733124], length 822
E..j..@.@...
l..
k.'#.......2.0....R w.....
..*....DHTTP/1.1 200 OK
Access-Control-Allow-Origin: *
Content-Encoding: gzip
Vary: Accept-Encoding
X-Accelerator: trickster 1.0.4
Date: Wed, 24 Apr 2019 20:14:16 GMT
Content-Length: 600
Content-Type: application/x-gzip
```

Bizarrely, I'm unable to reproduce it using curl and "-H" parameters to duplicate the HTTP request sent by Grafana, I even get extra headers in the response. When I use curl I get "Content-Type: application/json" but when I tcpdump the request from Grafana I see "Content-Type: application/x-gzip".  The latter should only happen if it was a gzipped file on-disk that was compressed a second time (encoded) by gzip again.

```
GET /api/v1/query_range?query=node_load5%7Brole%3D%22tls_alerts%22%2Cip%3D~%22.*%22%7D&start=1556133300&end=1556136960&step=60&timeout=1m HTTP/1.1
Host: gis-trickster-mog.[redacted]:9090
[...]

PR #221, opened 2019-05-02
Properly set Content-Type

jacksontj:
Fixes the regression added in #196 

cc @jranson once this is merged, can we cut a release of the 0.x series?

### DIFF
--- a/handlers.go
+++ b/handlers.go
@@ -291,7 +291,12 @@ func (t *TricksterHandler) getURL(o PrometheusOriginConfig, method string, uri s
 	}
 
 	startTime := time.Now()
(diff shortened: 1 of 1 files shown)


hatemosphere:
@jranson Any chances to port this fix for 1.x?

jranson:
Hi @hatemosphere, this fix was actually in 1.x first; this PR is the back-port to 0.1.x

hatemosphere:
@jranson so then i guess this https://github.com/Comcast/trickster/issues/216 is not related, right?

jranson:
Correct, this issue here was around proxy-only requests and not assuming everything will be `application/json`, while the issue in #216 affects objects that are cached by Trickster and how upstream and downstream compression plays into automatic header creation for those objects at the time they are served. We will get 216 worked out today; sorry for the delay on that one!

PR #224, opened 2019-05-09
1.0 fix various bugs and tests

jranson:


internal/config/loader_test.go
```diff
@@ -39,3 +39,475 @@ func TestLoadConfiguration(t *testing.T) {
...
+		t.Errorf("expected 57821, got %d", ProxyServer.ListenPort)
+	}
+
+	if ProxyServer.ListenAddress != "test" {
+		t.Errorf("expected test, got %s", ProxyServer.ListenAddress)
+	}
+
+	// Test Metrics Server
+	if Metrics.ListenPort != 57822 {
+		t.Errorf("expected 57821, got %d", Metrics.ListenPort)
```
LimitlessEarth: 57822?

internal/config/loader_test.go
```diff
@@ -39,3 +39,475 @@ func TestLoadConfiguration(t *testing.T) {
...
+	}
+
+	if Metrics.ListenAddress != "metrics_test" {
+		t.Errorf("expected test, got %s", Metrics.ListenAddress)
```
LimitlessEarth: test -> metrics_test?

internal/config/loader_test.go
LimitlessEarth: Oh you set constants in the next Commit. NVM.

internal/config/loader_test.go
LimitlessEarth: Oh you set constants in the next Commit. NVM.

internal/config/loader_test.go
jranson: Actually it's not the constants that makes this work, but rather the configuration files in `testdata/test.full.conf` and `testdata/test.empty.conf`. This test loads those files and expects them to resolve specific values.

PR #249, opened 2019-07-12
add endpoint for printing running config

jranson:


crandles:
Should you remove passwords from the config before encoding?

jranson:
Nahhhh we don't need that! ;)    Good catch, incoming!

PR #250, opened 2019-07-16
Support Multi-Origin in the Helm Chart

lazypower:
I'm not positive of the consequences of how i've implemented this. We're going
to be using a trickster multi-origin deployment, and the chart was the fastest
way to get this up and running in our use-case. However the chart doesn't seem
to support multi-origin out of the box.

I had the idea to just provide an override configmap name, to eschew any
unintended side-effects of the template hacking i've done, but decided it might
be best as a first iteration to shop a "pass-through" config and see how it would
be received, as it's then all provided by the helm chart and not a one-off manifest
that an operator would have to deal with. But i can see negatives to both approaches.

with what i'm proposing, to configure multi-origin would look similar to the following:

```
rawOrigins: |                                                                   
 [origins]                                                                      
    [origins.default]                                                           
        origin_url = 'http://preview-prometheus-server.management-api-preview.svc.cluster.local'
        api_path = '/api/v1'                                                    
        default_step = 300                                                      
        ignore_no_cache_header = false                                          
        max_value_age_secs = 86400                                              
    [origins.preview]                                                           
        origin_url = 'http://preview-prometheus-server.management-api-preview.svc.cluster.local'
        api_path = '/api/v1'                                                    
        default_step = 300                                                      
        ignore_no_cache_header = false                                          
        max_value_age_secs = 86400                                              
    [origins.dev]                                                               
        origin_url = 'http://dev-prometheus-server.management-api-dev.svc.cluster.local'
        api_path = '/api/v1'                                                    
        default_step = 300                                                      
        ignore_no_cache_header = false                                          
        max_value_age_secs = 86400   
```

I'm happy to refactor with guidance on what's a desireable outcome.

### DIFF
--- a/deploy/helm/trickster/README.md
+++ b/deploy/helm/trickster/README.md
@@ -46,6 +46,7 @@ Parameter | Description | Default
 `persistentVolume.size` | trickster data Persistent Volume size | `15Gi`
 `persistentVolume.storageClass` | trickster data Persistent Volume Storage Class | `unset`
 `podAnnotations` | annotations to be added to trickster pods | `{}`
+`rawOrigins` | manual origin TOML segment to provide the trickster configmap | `""`
 `replicaCount` | desired number of trickster pods | `1`
 `statefulSet.enabled` | If true, use a statefulset instead of a deployment for pod management | `false`
 `priorityClassName` | trickster priorityClassName | `nil`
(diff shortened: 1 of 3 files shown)


crandles:
I think we may have enabled support for this on the `next` branch chart, which uses the 1.x image.

https://github.com/Comcast/trickster/tree/next/deploy/helm/trickster

lazypower:
ah yeah, and that's more along what I initially wanted to do but it felt very invasive, though i see you split apart the multi-home origins, and still keep the default. Interesting.

Ok i'll pull the next branch and kick it around a bit. Thank you for letting me know! :bowing_man:

Issue #239, opened 2019-06-15
Any plans for tickster as a library?


jacksontj:
I see in the new 1.x branch that the code is no longer in the root of the repo (yay!), but I do see that most of the code is under `internal` which means that trickster wouldn't be usable as a library anywhere-- are there any plans to allow/enable this type of usage?

jranson:
That definitely sounds interesting! We've been intentionally placing things deemed worthy of import by other projects into `pkg`, with the criteria of being simple, useful, portable and unique; and keeping the currently-less-sharable stuff in `internal`. With 1.0, the intent is to make things as generic, yet customizable, as possible for the existing users and developers. So those exposed packages fit that theme. I would say when we get `next` merged into `master`, we can start working on getting more of the core trickster packages into `pkg` so long as they are refactored to be a little more generic and less intertwined with each other. And with that, help is definitely welcomed :)

jranson:
@jacksontj there is some interest in giving this a look by a few other stakeholders from the Trickster community. @dmitsh and @alfred-landrum from Sysdig demonstrated to me a prom-specific working POC of an importable Trickster package; it looks like a really good starting point for this feature. 

One of the things I am concerned about getting right is ensuring that we make make the package solution simple, easy-to-use and future proof. As we are planning to expose Trickster's HTTP Reverse Proxy/Cache abilities as an origin type itself (outside of time series accelerators) in 1.0, a single generic library may or may not be the best idea. So I think the options are, 1) come up with a single package that is designed well enough to flexibly support all origin types (including non-timeseries caching), or 2) expose origin-type-specific libraries (e.g., `/pkg/trickster/prometheus`, `/pkg/trickster/hrpc`) that would be scoped for the features for the specific origin type.

I see benefits to both approaches. Everyone who has inquired about this is developing a solution for Prometheus, so it's possible that Option 2 is more desirable. I invite everyone with an opinion to chime in on this issue so that we can make the right choices. Based on the consensus here, we can come up with the final design and get it shipped. Thanks to everyone for their interest and feedback!

dmitsh:
I've posted a PR #262 with our proposed approach. This PR incorporates @jranson idea of origin-type-specific libraries.

jacksontj:
My initial thinking is that approach #2 would be beneficial -- as each system has their own client types etc -- so it would be nice if they mirrored the upstreams. Then trickster the binary could use those same libraries itself (meaning the http -> library translation is done presumably in *another* library).

[...]

PR #262, opened 2019-08-16
use trickster as a library (PoC, do not merge)

dmitsh:
- add Prometheus support

Signed-off-by: Dmitry Shmulevich <dmitry.shmulevich@sysdig.com>


dmitsh:
A proposed solution for issue #239 (cc @jranson @alfred-landrum @jacksontj)

pkg/trickster/prometheus/client.go
```diff
@@ -0,0 +1,55 @@
...
+		apiPrefixMap: map[string]processFunc{
+			originConfig.PathPrefix + originConfig.APIPath + "/query_range?": processQueryRange,
+			originConfig.PathPrefix + originConfig.APIPath + "/query?":       processQuery,
+			originConfig.PathPrefix + originConfig.APIPath + "/series?":      processSeries,
+		},
+	}
+	return t, nil
+}
+
+func (t *TricksterClient) Process(w http.ResponseWriter, r *http.Request) {
```
jacksontj: Personally I'd prefer if this was more structured, the prometheus clients for example have Query/QueryRange etc. -- which is what clients using this as a library would expect.

dmitsh: Agreed and implemented. That makes the code even cleaner.

dmitsh: I've added another commit to this PR with 2 changes:
First, I implemented @jacksontj recommendations regarding explicit API calls.
Second, I attempted to make the library more independent by passing logger to the library as opposed to using the static logger.
The code is not completed and won't compile, but the overall idea is there.
I would like to get the feedback before proceeding with this rather a massive change.
Thanks!

dmitsh:
In the latest commit I have completed the change to remove the static logger, and updated all the tests. Now the code compiles, and the tests are successfully passing.
I would appreciate your feedback guys (cc @jranson @jacksontj @alfred-landrum)

dmitsh:
I noticed that TestDF4SeriesEnvelopeMerge has failed. It is weird. I haven't changed that part of the code, and that test passes successfully on my local machine. Any clue?
```
$ go test -run TestDF4SeriesEnvelopeMerge
PASS
ok  	github.com/Comcast/trickster/internal/proxy/origins/irondb	0.006s
```

pkg/trickster/prometheus/client.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		return nil, err
+	}
+	t := &PrometheusClient{
+		client: prometheus.NewClient(originConfig.Type, originConfig, c, logger),
+	}
+	return t, nil
+}
+
+func (t *PrometheusClient) Query(w http.ResponseWriter, r *http.Request) {
```
jacksontj: I was referring to a more parsed-out specific handler type e.g. https://godoc.org/github.com/prometheus/client_golang/api/prometheus/v1#API


This is an improvement, but this would mean that all library users would have to convert their parsed out types into http request/response objects just to use this library.

jranson:
@dmitsh @jacksontj we have moved from `master` to `main` and unfortunately that has closed this PoC PR and I am unable to open it. But we have the corresponding issue still open and expect to have this fully worked out in the 2.0 release.

PR #264, opened 2019-08-20
Adds support to de-couple cache from origins

lazypower:
- Adds an 'is_default' parameter during the template generator. the if
  guards prevent the key from being rendered on non-declared origins.
- Adds a cacheName parameter during generation to remove the tightly
  coupled cache/origin assumption in the chart. Defaults to the existing
  behavior of desiring a cacheName that matches the origin monicker.

Note: Still needs a README update, but it didn't look like all the
origin values are documented yet. The values are optional, but i declared them in the base values.yaml to illustrate their usage. I'm happy to circle back and document the origin datapoints in the table or as a blurb. Just advise on what would be a more useful contribution.


lazypower:
looks like this was already in progress and i was a little early :) :+1:

PR #268, opened 2019-08-25
Observe frontend requests

pcarranza:
This PR provides a set of metrics for return codes and durations from the front end perspective of the cache.

These metrics are:
1. trickster_frontend_requests_total *counter*
2. trickster_frontend_requests_duration_seconds *histogram*
3. trickster_frontend_written_bytes_total *counter*

All metrics have the following labels:
1. http_status which follows the haproxy style of status code grouping them by hundreds (2xx, 3xx, etc)
2. method - GET/POST
3. origin_name
4. origin_type
5. path which only provides the general handler, for example: proxy, query, query_range, status, etc.

Still pending implement the same for other backends besides prometheus.

### DIFF
--- a/internal/proxy/origins/prometheus/routes.go
+++ b/internal/proxy/origins/prometheus/routes.go
@@ -14,61 +14,68 @@
 package prometheus
 
 import (
+	"net/http"
+
 	"github.com/Comcast/trickster/internal/config"
 	"github.com/Comcast/trickster/internal/routing"
 	"github.com/Comcast/trickster/internal/util/log"
+	"github.com/Comcast/trickster/internal/util/middleware"
 )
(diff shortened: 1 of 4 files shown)


pcarranza:
Hey @jranson. Thanks!

For the record, there are 2 tests that seem to be failing intermittently. 

You can see one [here](https://github.com/Comcast/trickster/pull/268/commits/10ab8ede49439a943881ed7b207bcdc2b88118bf) and other one is TestBboltCache_StoreNoIndex.

These 2 tests are on the backend side of the app, thus I assumed that my changes were not a possible source of them failing.

In any case, there's that, seems like some form of race condition or dependency between tests.

PR #292, opened 2019-10-02
makefile: feature: added commit id and build time

Flamingo09:


Flamingo09:
> I don't think this code bakes the build time and commit id directly into the binary. It seems to echo it to terminal, I'm not sure if that's what the maintainer wants.

Got your point.

jranson:
@Flamingo09 just wanted to check on the progress with this PR and incorporating @pavan-kalyan's feedback.

jranson:
This feature was added in #295

PR #295, opened 2019-10-11
added build time and commit ID to the binary

pavan-kalyan:
I have added the build time using `date -u %FT%T%z` which gives sample timestamp 2019-10-11T07:46:1707:46:17+0000 which seems to be the standard according to [here](https://stackoverflow.com/questions/11354518/application-auto-build-versioning) and [here](https://github.com/commissure/go-git-build-vars/blob/master/Makefile). I made 2 global variables in `cmd/trickster/main.go` applicationBuildTime and applicationGitCommitID which are set during build time from the Makefile using `-ldflags` option.
@jranson please let me know if any changes are required/recommended.


pavan-kalyan:
It seems to be failing at make style test build. I will make the required changes and test again.

Issue #285, opened 2019-09-28
frontend tls support


jranson:
as an end user, I would like for trickster to support TLS natively on the frontend (facing clients) without the need for a sidecar. This should include providing trickster, on a per-origin-config basis, with the path to an Intermediate/Chain Certificate Authority file; a Certificate file; a Certificate Key file; and any other necessary configurations to implement this feature.

PR #296, opened 2019-10-11
Support TLS #285

u5surf:
#285 
I added TLS setting in trickster.config


jranson:
@u5surf Thanks for this great contribution!

PR #298, opened 2019-10-12
1.0 tls support

jranson:
This builds on #296, which provides basic TLS support into `master` thanks to @u5surf.  This PR extends that functionality into `next`, and provides additional functionality. In this port+extension, multiple TLS configs can be provided, and each origin configuration can be mapped to any one of the configured TLS sections. Further, each origin can support http only, http+https, or https only, depending upon the operator's preference. Trickster will only listen on the TLS and/or HTTP ports if origins are configured for either, and origins will only be routed to the TLS and/or HTTP router based on the operating configuration. Test coverage is also included, as well as a few ancillary bug fixes that were identified along the way during testing.


jranson:
this PS is about 92.7% completed at this point, but I forgot to mark the 'Draft' flag on it. i will request reviews when it is fully ready.

jranson:
PR ready for review.

.gitignore
```diff
@@ -31,9 +31,6 @@ OPATH
 ehthumbs.db
 Thumbs.db
 
-# The binary
-trickster
```
LimitlessEarth: Do we want to ignore it in OPATH as well?

cmd/trickster/main.go
```diff
@@ -72,12 +73,47 @@ func main() {
...
+					config.ProxyServer.ConnectionsLimit,
+					tlsConfig)
+				if err == nil {
+					err = http.Serve(l, handlers.CompressHandler(routing.TLSRouter))
+				}
+			}
+			log.Error("exiting", log.Pairs{"err": err})
+			wg.Done()
+		}()
+	}
```
LimitlessEarth: Do we want to log here in the else condition? Want to avoid failing silently.

.gitignore
jranson: OPATH is already in .gitignore so we're good

cmd/trickster/main.go
jranson: I think we are OK here, in that it should not be possible to fail silently. The condition block starting L66 will fatal out before getting to this point if no listeners are configured, and the configuration loader L48 will fail out earlier than that if TLS is misconfigured; therefore, at this point of the execution, at least one of the listener blocks must execute and they are both logging errors.

PR #315, opened 2019-10-25
1.0 wc method path

jranson:
allows a user to


jranson:
Well that was awkward.   It allows a user to smack down smart-aleck reviewers? ;-)  heh  hopefully this functionality is actually self-explanatory!

PR #320, opened 2019-10-29
Flatten path config to follow same pattern as regular config

LimitlessEarth:
This changes the path config to follow the same pattern as the main config.


LimitlessEarth:
Nvm. This was needed.

PR #336, opened 2019-11-19
Update configmap.yaml

hieultan:
I was confused with this configmaps, because if I deploy and check at path: `localhost:9090/trickster/config` to see the configure at runtime, it will be load with 
```
[origins]
  [origins.default]
    type = "prometheus"
    scheme = "http"
    host = "prometheus:9090"
```
and 
```
[caches]
  [caches.default]
    type = "memory"
```
So, let me check at source code, it must be 
```
[origins]
  [origins.default]
    origin_type = "prometheus"
    scheme = "http"
    origin_url = "prometheus:9090"
```
and 
```
[caches]
  [caches.default]
    cache_type = "memory"

```
Refer: https://github.com/Comcast/trickster/blob/next/internal/config/config.go#L103

But, i changed some configure at helm chart on this PR, it's work
What wrong with me?

### DIFF
--- a/deploy/helm/trickster/templates/configmap.yaml
+++ b/deploy/helm/trickster/templates/configmap.yaml
@@ -41,7 +41,7 @@ data:
 
       {{ printf "[caches.%s]" .name }}
         {{- if .cacheType }}
-      cache_type = {{ .cacheType | quote }}
+      type = {{ .cacheType | quote }}
         {{- end }}
         {{- if .compression }}
       compression = {{ .compression }}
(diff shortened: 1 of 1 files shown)


jranson:
@hieultan the `type` fields are part of previous Trickster releases. They are changed to `cache_type` and `origin_type` for clarity, which will be in the next release (1.0 Beta 10) which is forthcoming.

hieultan:
@jranson I use image `tricksterio/trickster:1.0-beta` and it works with `type` fields ;) Hopefully helm chart and release 1.0 Beta 10 is compatible

jranson:
Ah, yes, the `next` branch is currently just ahead of the latest release. You can get the helm chart for the latest beta @ https://github.com/Comcast/trickster/tree/v1.09-beta9/deploy/helm However, it is not as comprehensive as the version in `next`. So I think if changing those two fields works for you locally, that is great and hope you'll try out the next beta where they'll need to be changed back. Thanks for using and contributing to Trickster!

Issue #12, opened 2018-03-30
Support dynamic config reloading


alkalinecoffee:
It would be nice if Trickster supported dynamic reloading of the config file (similar to Prometheus' `POST /-/reload` endpoint or via kill signal).  This is useful when using consul-template for pushing configs dynamically.

pcarranza:
I'm thinking that it would be enough with just reloading the origins from the configuration file on SIGHUP as a first step.

Reloading ports would include bouncing the server, and that may be too much.

Would it make sense as a first approach?

jranson:
I am working on this at https://github.com/jranson/trickster/tree/config-reload and have it working but not tested beyond confirming a kill -HUP <pid> will trigger the config reload, and the reloaded origins are respected. I would appreciate any help testing out to ensure it is otherwise safe!

sylr:
You can also implement fsnotify so you don't need external action to reload config.

https://github.com/jimmidyson/configmap-reload/blob/master/configmap-reload.go#L46-L75

SuperQ:
I would prefer the standard convention of active reloading. Automatic reloading can be touchy if the config is reloaded in the middle of a file write. So you end up having to always do atomic writes, which means if you're just using a text editor to tweak a config, it now becomes much more complicated.

jranson:
All, at long last, v1.1 is released and includes the config reload feature, which you can read about [here](https://github.com/tricksterproxy/trickster/blob/master/docs/configuring.md#reloading-the-configuration). Enjoy!

PR #342, opened 2019-12-05
remove config package-scoped variables

crandles:
As part of supporting config reloading (#12), this PR aims to remove package-scoped global variables that are utilized throughout trickster. These are generally hard to maintain if they require post-init updates, and would all require some sort of locking or atomic operations for coordinating usage.

This will not attempt to add config reloading functionality, but should prepare the codebase for it 🤞.

https://peter.bourgon.org/blog/2017/06/09/theory-of-modern-go.html
https://dave.cheney.net/2017/06/11/go-without-package-scoped-variables

TO DO:
- [ ] Remove `log` singleton
- [ ] Remove global state required for binding flags
- [x] fix tests
- [ ] remove comments

Test failures don't seem to be returned through travis; I may look at that as well.

### DIFF
--- a/internal/config/loader_test.go
+++ b/internal/config/loader_test.go
@@ -23,21 +23,21 @@ import (
 func TestLoadConfiguration(t *testing.T) {
 	a := []string{"-origin-type", "testing", "-origin-url", "http://prometheus:9090/test/path"}
 	// it should not error if config path is not set
-	err := Load("trickster-test", "0", a)
+	conf, _, err := Load("trickster-test", "0", a)
 	if err != nil {
-		t.Error(err)
+		t.Fatal(err)
 	}
 
-	if Origins["default"].TimeseriesRetention != 1024 {
-		t.Errorf("expected 1024, got %d", Origins["default"].TimeseriesRetention)
+	if conf.Origins["default"].TimeseriesRetention != 1024 {
+		t.Errorf("expected 1024, got %d", conf.Origins["default"].TimeseriesRetention)
 	}
 
-	if Origins["default"].FastForwardTTL != time.Duration(15)*time.Second {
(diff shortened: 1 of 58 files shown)


jranson:
re: the test failures, if I had just one guess, the cause of the test return code is likely because the `make test` is being piped to a grep that strips log line entries from the output. I would guess that the grep returns exit code 0 even if the make test returns a nonzero exit code, so we may need to pull that out.

internal/config/loader_test.go
```diff
@@ -92,63 +92,64 @@ func TestLoadConfigurationFileFailures(t *testing.T) {
...
-	a := []string{"-origin-type", "testing"}
-	err := Load("trickster-test", "0", a)
-	if err == nil {
-		t.Errorf("expected error `%s` got nothing", expected)
-	} else if err.Error() != expected {
-		t.Errorf("expected error `%s` got `%s`", expected, err.Error())
-	}
-}
+// TODO: this fails.. the default config has a url set, unsure why this would error
+// func TestLoadConfigurationMissingOriginURL(t *testing.T) {
```
crandles: this fails for me; there is a default that gets added by the Loader. I think this fails on `next` too.

jranson:
@crandles We've merged `next` into `master`, so I rebased your PR target here and `next` is no more. Currently all tests are passing on the `master`, so if you can rebase locally and fix up any compile failures as a result of the obscene amount of changes, it should be good to go. This is looking really good, thanks! Also, the fix is in for unit tests output and exit code working like it should, while also excluding generated code from the coverage report. I accomplished this by moving the exclusion sed from the Makefile target to the .travisci configuration, so you should be able to see all of the things now locally and in travis. Sorry for the frustrations with that!

PR #343, opened 2019-12-06
Release with tar.gz format instead of only gz format (support Ansible unarchive module)

xluffy:
I try to deploy trickster with Ansible, but Ansible doesn't support `.gz` release file. [From Ansible  unarchive docs](https://docs.ansible.com/ansible/latest/modules/unarchive_module.html):

- Can handle .zip files using unzip as well as .tar, .tar.gz, .tar.bz2 and .tar.xz files using gtar.
- Does not handle .gz files, .bz2 files or .xz files that do not contain a .tar archive.

So I recommend trickster releases with `tar.gz` format, please consider this request. Example Ansible code

image

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,8 +37,8 @@ release: build release-artifacts docker docker-release
 
 .PHONY: release-artifacts
 release-artifacts:
-	GOOS=darwin GOARCH=amd64 go build -o ./OPATH/trickster-$(PROGVER).darwin-amd64 && gzip -f ./OPATH/trickster-$(PROGVER).darwin-amd64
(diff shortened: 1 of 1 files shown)


jranson:
Hi xluffy, thanks for this contribution. I think it makes good sense, as it allows us, at some point in the future, to add additional files to the binary distribution (e.g., docs or example configs subdirectories). We are currently doing most of our work against the `next` branch (which we hope to merge into master soon), so while I'll merge this PR, I would ask if you wouldn't mind submitting a separate PR with the same update on `next` too. That way, the change is included in the very next beta release (and all going forward); as an added bonus you'll be credit with another contribution to the project on your github profile. Thanks again!

PR #344, opened 2020-01-16
1.0 overhaul

jranson:


internal/proxy/engines/caching_policy_gen.go
```diff
@@ -0,0 +1,409 @@
...
+// Code generated by github.com/tinylib/msgp DO NOT EDIT.
+
+import (
+	"github.com/tinylib/msgp/msgp"
+)
+
+// DecodeMsg implements msgp.Decodable
+func (z *CachingPolicy) DecodeMsg(dc *msgp.Reader) (err error) {
+	var field []byte
+	_ = field
```
LimitlessEarth: What is the purpose of this?

jranson: This is an auto-generated code file created by `go generate` and is used internally by the msgp unmarshaler. That is true for any file ending in `_gen.go` or `_gen_test.go`

internal/cache/memory/memory.go
```diff
@@ -48,35 +49,73 @@ func (c *Cache) Connect() error {
...
+
 // Retrieve looks for an object in cache and returns it (or an error if not found)
-func (c *Cache) Retrieve(cacheKey string, allowExpired bool) ([]byte, error) {
-	return c.retrieve(cacheKey, allowExpired, true)
+func (c *Cache) Retrieve(cacheKey string, allowExpired bool) ([]byte, status.LookupStatus, error) {
+	o, s, err := c.retrieve(cacheKey, allowExpired, true)
+	if err != nil {
+		return nil, s, err
+	}
+	var bytes []byte
```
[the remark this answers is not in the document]
jranson: looks like after distilling down the function somewhat with search/replace, this one slipped through. updated to operate identically to RetreiveReference. Thanks!

internal/proxy/engines/caching_policy_gen.go
LimitlessEarth: Lol. Gotcha. It probably knows something we don't.

Issue #36, opened 2018-04-16
Feature: OpenTracing support


mattbostock:
If Trickster supported tracing, it would be make it easier to see where time is spent in queries across all services involved in serving a query.

Grafana and [Thanos](https://github.com/improbable-eng/thanos) both have some support for tracing:
https://github.com/grafana/grafana/pull/9213

jranson:
@WillGunther for you :) Thanks!!

jranson:
We will be bumping this feature back to v1.1, since the OpenTelemetry libraries are currently barely emergent. We could adopt one of the now-sunsetting solutions from which OpenTelemetry is based and switch later, but that would be a lot of work to introduce a new feature. Targeting late Q3.

jranson:
Revisiting this one. Not sure why it didn't occur to me already, but tracing should certainly be something we implement as an interface, or some other mechanism that makes the underlying tracing implementation (if any) configurable by the Trickster operator, similar to how we already do for Cache type selection.

guygrigsby:
@jranson Is this the one you want me to take a look at? I'll get started right away.

jranson:
:-D Yes please!

guygrigsby:
I have been doing some research and playing with Open Telemetry in Trickster and here are my thoughts.

#### Interfacing
You mentioned above that we create our own interface and provide a reference for wrapping others, ex OpenTelemetry, within our interface. I suggest that we should use the OpenTelemetry API spec directly. Rather than adapting the OpenTelemetry API to our interface, we should adapt other tracing interfaces to OpenTelemetry. This is for the following reasons:

1. Open Telemetry _is_ the interface. Described [here](https://www.cncf.io/blog/2019/05/21/a-brief-history-of-opentelemetry-so-far/) 
> [...] as the next major version of both OpenTracing and OpenCensus

2. An OpenTracing bridge already exists as part of the [Go OpenTelemetry repo](https://github.com/open-telemetry/opentelemetry-go/tree/master/bridge/opentracing).

3. An OpenCensus bridge is on the [release roadmap for OpenTelemetry](https://github.com/open-telemetry/opentelemetry-go#release-schedule), though a date is not given.

#### Implementations

The OpenTelemetry Go client currently has support for 
 - Jaeger and 
 - Stackdriver 

[...]

PR #350, opened 2020-01-18
basic tracing support draft

guygrigsby:
Here we have a draft PR for Open Telemetry support. It includes span propagation from the client through the proxy and adds the events below. While initially I intended to implement all of them, time restraints have limited what I can do. That being said, I wanted to get this out there so Trickster has minimal tracing support for 1.0 and we can build it up as we go.

The requested support for initial tracing is as follows: (pulled from #36 )

- [x] upstream request time
- [x] fast forward request time
- [x] upstream revalidation request time
- [x] cache query time (also recording the query result in the span)
- [x]  cache write time (recording # bytes written into the span)

~~As I was testing the tracing, I noticed that some traces are happening twice. The more I dug into it, the more I had a suspicion that it wasn't the tracing, but the tracing was exposing that we are making requests twice. This may be expected behavior that I don't understand, but I thought it was worth pointing out.~~ Resolved.


jranson:
What a Guy! Thank you, sir. Do you happen to know which spans are doubling up?  In the delta proxy cache, this would be possible (or probable?), depending upon the spans and how they are observed. For example, if Trickster needs to make parallel requests to the upstream origin, for fast forward, or for reconstituting disparate ranges in the cache to fulfill the client request, that may register the spans instrumented during the upstream request process multiple times.

guygrigsby:
The initial trace starts in the middleware and it looks that's where the duplication starts. It's basically two full traces. I have been using the example calls found in Promsim,  I am adding a middleware function to all generated handlers. 

## Example 
I make this call to a local Trickster backed by PromSim:
```
curl -i 'localhost:8080/test/api/v1/query_range?query=my_test_query{random_label="57",series_count="1"}&start=2&end=40&step=1'
```
it runs the delta routine two times and we get two traces. (screenshots below)
### Log:
```
time=2020-01-18T16:37:41.636963Z app=trickster caller=proxy/proxy.go:126 level=debug event="starting proxy listener" scheme=http address=localhost port=8080 connectionsLimit=0
time=2020-01-18T16:37:43.304852Z app=trickster caller=proxy/engines/access_logs.go:23 level=debug event="upstream request" originName=test handlerName=query_range method=GET uri="http://localhost:9090/api/v1/query_range?end=40&query=my_test_queryrandom_label%3D%2257%22&start=2&step=1" userAgent=curl/7.64.1 code=200 size=481 durationMS=1 originType=prometheus
time=2020-01-18T16:37:43.305299Z app=trickster caller=proxy/engines/deltaproxycache.go:364 level=debug event="delta routine completed" cacheKey=localhost:9090.bd1f28939c6c25a3b54c3098e6ec75fd cacheStatus=kmiss reqStart=2 reqEnd=40
time=2020-01-18T16:37:43.305509Z app=trickster caller=runtime/asm_amd64.s:1357 level=debug event="memorycache cache store" is_direct=true cacheKey=localhost:9090.bd1f28939c6c25a3b54c3098e6ec75fd length=0 ttl=6h0m0s
time=2020-01-18T16:37:43.306858Z app=trickster caller=proxy/engines/access_logs.go:23 level=debug event="upstream request" code=200 originName=test method=GET userAgent=curl/7.64.1 size=481 durationMS=0 originType=prometheus handlerName=query_range uri="http://localhost:9090/api/v1/query_range?end=40&query=my_test_queryseries_count%3D%221%22&start=2&step=1"
time=2020-01-18T16:37:43.307016Z app=trickster caller=proxy/engines/deltaproxycache.go:364 level=debug event="delta routine completed" reqStart=2 reqEnd=40 cacheKey=localhost:9090.0b9bf1721fa89fecae272d494ef73847 cacheStatus=kmiss
time=2020-01-18T16:37:43.307232Z app=trickster caller=runtime/asm_amd64.s:1357 level=debug event="memorycache cache store" cacheKey=localhost:9090.0b9bf1721fa89fecae272d494ef73847 length=0 ttl=6h0m0s is_direct=true
```
### First Trace:
image

### Second Trace:
image

## Other example
If I call
```
curl -i 'localhost:8080/test/api/v1/query_range?query=my_test_query{random_label="57",series_count="2"}&start=2&end=4&step=2
```
when the data is comletely cached, I still get two traces. The cache keys are different in the logs. 

```
time=2020-01-18T16:34:20.460968Z app=trickster caller=cache/memory/memory.go:128 level=debug event="memory cache retrieve" cacheKey=localhost:9090.759d69494f85a30c5dba81ab0cae191a
time=2020-01-18T16:34:20.461067Z app=trickster caller=proxy/engines/deltaproxycache.go:364 level=debug event="delta routine completed" cacheStatus=hit reqStart=2 reqEnd=4 cacheKey=localhost:9090.759d69494f85a30c5dba81ab0cae191a
time=2020-01-18T16:34:20.461464Z app=trickster caller=cache/memory/memory.go:128 level=debug event="memory cache retrieve" cacheKey=localhost:9090.54e1ed9896bcc071b671cd6a6f7c3419
time=2020-01-18T16:34:20.461522Z app=trickster caller=proxy/engines/deltaproxycache.go:364 level=debug event="delta routine completed" cacheStatus=hit reqStart=2 reqEnd=4 cacheKey=localhost:9090.54e1ed9896bcc071b671cd6a6f7c3419
```

I have been testing with a limited number of http calls to Trickster, mostly examples. The number of traces seems to equal the number of calls to delta so I am thinking it has to do with the fact that I am adding the trace middleware handler to _all_ handlers. It seems for every delta run, we get a distinct trace. Any other ideas?

Even if this is expected behavior, the trace should be enclosing all those calls into a single trace.

guygrigsby:
I am an idiot. I was sending multiple queries. Disregard duplicate comment.

jranson:
One general change I would prefer is that we scope a TracingConfig to an Origin, rather than to the Trickster process, as there is a possibility different Origins will need different TracingConfigs. I think we can easily change this out for the same map[string]*__Config approach we use for mapping caches and negative caches to an origin. There may be implications with the context to address as a result, but should be minimal effort.

internal/util/tracing/registration/registration.go
```diff
@@ -0,0 +1,109 @@
...
+		},
+	)
+
+	tracer, fl, _, err = tracing.SetTracer(
+		tracing.TracerImplementations[cfg.Implementation],
+		tracing.TraceExporters[cfg.Exporter],
+		cfg.CollectorEndpoint,
+		cfg.SampleRate,
+	)
+	flusher = fl
```
LimitlessEarth: Any reason we don't just return fl?

cmd/trickster/main.go
```diff
@@ -84,6 +85,19 @@ func main() {
 	}
 
 	metrics.Init()
+
+	// Register Tracing Configurations
+	tracerFlushers, err := tr.RegisterAll(config.Config)
+	if err != nil {
+		log.Fatal(1, "tracing registration failed", log.Pairs{"detail": err.Error()})
```
LimitlessEarth: Do we want this to be fatal? I feel like we probably want to whine about it but fail open to not block the main functionality.

`					return nil, fmt.Errorf("invalid tracing implementation [%s] for tracing config [%s]", tc.Implementation, oc.TracingConfigName)`

like if someone goofs the config value do we want to block trickster as a whole?

internal/config/tracing.go
```diff
@@ -0,0 +1,57 @@
...
+
+// Clone returns an exact copy of a tracing config
+func (tc *TracingConfig) Clone() *TracingConfig {
+	return &TracingConfig{
+		Name:              tc.Name,
+		Implementation:    tc.Implementation,
+		Exporter:          tc.Exporter,
+		SampleRate:        tc.SampleRate,
+		CollectorEndpoint: tc.CollectorEndpoint,
+		Tracer:            tc.Tracer,
```
LimitlessEarth: Since this is an interface we are just going to copy the pointer to the struct. Is that the expected behavior?

internal/proxy/engines/objectproxycache.go
```diff
@@ -131,6 +133,20 @@ func handleCacheRangeMiss(pr *proxyRequest) error {
...
+
+	ctx, span := tracing.NewChildSpan(pr.Request.Context(), oc.TracingConfig.Tracer, "CacheRevalidation")
+	defer func() {
+		reval := revalidationStatusValues[pr.revalidation]
+		span.AddEvent(
+			ctx,
+			"Complete",
+			core.Key("Result").String(reval),
+		)
+		defer span.End()
```
LimitlessEarth: We may not need to defer in this defered function.

internal/util/testing/testing.go
```diff
@@ -107,6 +108,10 @@ func NewTestInstance(
 	oc := config.Origins["default"]
 	p := NewTestPathConfig(oc, DefaultPathConfigs, urlPath)
 
+	tracer, _, _ := tr.Init(oc.TracingConfig)
+	// TODO worry about running closures for cleanup once the test is complete
```
LimitlessEarth: Is still to do?

internal/util/tracing/registration/registration.go
guygrigsby: Good call. I think I was trying to readability, but it just seems to confuse it.

cmd/trickster/main.go
guygrigsby: It's a valid question. I have always been of the mind that if any of our observability tools fail, we fail, but that's more from a SAAS point of view. For a product like this, I think it's viable either way. Need some advice from dad, @jranson ?

internal/config/tracing.go
guygrigsby: Yeah, great point. I don't think that there is a guarantee that the impl _is_ a pointer.

cmd/trickster/main.go
jranson: Any detectable configuration issue is that is due to human error should cause the application to exit fatally during startup, regardless of how insignificant the issue seems. If we start to make exceptions here and there, then we'd have to maintain the list of which features can be fatally configured and which ones cannot. So the all or nothing approach will win here. (resolving)

internal/config/tracing.go
jranson: unfortunately, the Tracer interface does not provide a Clone() method that we can call, so this might have to do, short of re-registering the Tracer altogether during Clone. But since the configuration around the already-instantiated tracer would be identical to a new one that is registered, it should be fine to to re-use the existing Tracer during clone. If the user subsequently edits the cloned config, they would be on the hook for re-registering.

internal/util/testing/testing.go
jranson: This is a to-do we'll leave for now since it's non-consequential to the running app or tests, but would be good to put in for completeness. My goal will be to get back to it some in the distant future but not in this PR.

PR #351, opened 2020-01-18
Update vendor

jranson:


jranson:
(the mysterious James Ranson is me, new laptop and did not set the git flags yet; merging)

Issue #337, opened 2019-11-19
Possible to read metrics in outside max_value_age_sec?


ninjaslothx:
Is it possible to have trickster read in metrics further back in time than the cache will allow...but not cache?

Example we have metrics going back several years. 
90% of our dashboard usage is for 1week~. Trickster cache is great for this and is set to this size. 
However for us to query data outside this period we have to setup a datasource variable in grafana to change to "long term" bypassing trickster and going direct to thanos-query. 
Is it possible to have trickster pull in what it can from the cache and otherwise run a non cached query if it exceeds the cache age?

jranson:
hi @ninjaslothx what version of Trickster do you use? The latest stable release (0.1.x) as well as the 1.0 beta (and several releases prior) work the way you are asking without needing additional data sources.

ninjasloth22:
Currently on 0.1.10
Is there a setting required to enable this? I am still getting only metrics up to the cache age.

![image](https://user-images.githubusercontent.com/14971594/70667287-581b0900-1cd5-11ea-8ecd-48b9d698b1e9.png)

My config looks like the following:
`[main]

[proxy_server]

[cache]
cache_type = 'filesystem'
    [cache.filesystem]
    cache_path = '/opt/data/trickster/cache'

[origins]
    [origins.default]
    origin_url = 'https://thanosurl'
    api_path = '/api/v1'
    max_value_age_secs = 432000

[metrics]
listen_port = 8082

[profiler]

[logging]
log_level = 'info'
log_file = '/var/log/trickster.log'`

jranson:
Hi @ninjaslothx, would you mind trying this on the 1.0 Beta 10 that was just released (or beta 9 if you're not feeling adventurous). As we are primarily focused on 1.0 in advance of the GA release (hopefully next month!), and as of today the 1.0.x codebase is merged to `master`, so the older 0.1.x codebase can no longer be maintained.

ninjaslothx:
Hey @jranson, I upgraded to Beta 10. This has indeed resolved the issue! 
I have however run into a new issue where after the first query or two are run it fails to run any new queries. 
The only fix is to stop the service and delete the cache. However the same scenario happens next run. 
If I try to restart the service without deleting the cache I get this.

`panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x9182e3]
goroutine 1 [running]:
[...]

PR #353, opened 2020-01-22
fix startup panic for bbolt and filesystem cache types

jranson:
this addresses a bug wherein a filesystem or bbolt cache type will cause trickster to panic on subsequent starts of the trickster process, because the filesystem cache or bbolt cache had already been created on a previous run of the trickster process. The panic is due to the cache.Retrieve() function, which is called to load the index from the filesystem or bbolt cache at startup, being dependent upon the Index being loaded already so as to check if the object has expired before serving (the expiration metadata is in the Index only). This creates a chicken-or-the-egg situation with the cache and its index, and is resolved by allowing the cache to simply return a cache hit in the event that the Index is not loaded (because we're trying to load it), thus bypassing the Index-dependent checks on the object before confirming a cache hit. This did not affect the first run of the application (e.g., unit tests) because it requires a cache hit of the index object in order to trigger the panic condition. The tests now address this.


jranson:
(p.s., fixes #337)

PR #370, opened 2020-02-12
Helm Updates

LimitlessEarth:
# General
Fixes:
1. label selectors not matching labels
2. missing config mount path
3. unquoted `1.0` image tag truncates to `1`, which is not the same image tag.
4. change mounting path pattern now that we have multiple configs
5. fix ingress controller templating

# Details
### 1
The first issue relates to [label selector updates](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates)

### 2
Not sure how the second issue came about. No matter what was changed my config values were not reflected -> missing volume mount. The helm templated trickster config was never being exposed in the pod.

### 3
Last issue is just a simple fix due to how helm does numbers.

### 4
This is probably the beefiest section of the rework.
The new scheme supports multiple many to one and many to many relationships between filesystem reliant caches and the underlying volumes present in Kubernetes.

Each cache now has a `volumeName` section that maps to the corresponding name present in the `volumes` section of the Values file. Multiple caches may map to a single volume. This is facilitated by `subDir`s on the volumeMount.

We may need to think this through harder as more use cases arise but for now the filesystem related caching option have simplified because of this mapping as the details of where to spin up a cache are handled in templating. 

Any given filesystem dependent cache will appear at the value generated by `.Values.volumes.[].mountPath/.cache.Name/`

### 5
just some mis-matched templating.

_Side Node:_ There could be a way better solution to these. I haven't touched helm in a while. If anyone has suggestions please say so.

### DIFF
--- a/deploy/helm/trickster/templates/deployment.yaml
+++ b/deploy/helm/trickster/templates/deployment.yaml
@@ -25,8 +25,7 @@ spec:
 {{ toYaml .Values.podAnnotations | indent 8 }}
     {{- end }}    
       labels:
-        app: {{ template "trickster.name" . }}
-        release: {{ .Release.Name }}
+        {{- include "trickster.labels" . | nindent 8 }}
       annotations:
         checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
     spec:
@@ -58,6 +57,8 @@ spec:
           {{- end }}
         {{- end }}
           volumeMounts:
+            - name: cfg-volume
+              mountPath: /etc/trickster
           {{- range .Values.caches }}
           {{- if eq .cacheType "filesystem" }}
             - name: cfg-volume-{{ .name }}
--- a/deploy/helm/trickster/templates/deployment.yaml
+++ b/deploy/helm/trickster/templates/deployment.yaml
@@ -59,14 +59,13 @@ spec:
           volumeMounts:
             - name: cfg-volume
               mountPath: /etc/trickster
-          {{- range .Values.caches }}
-          {{- if eq .cacheType "filesystem" }}
-            - name: cfg-volume-{{ .name }}
-              {{- if eq .path "" }}
-              mountPath: /tmp/trickster
-              {{- else }}
-              mountPath: {{ .path }}
-              {{- end }}
+          {{- range $key, $value := .Values.volumes }}
+          {{- range $.Values.caches }}
+          {{- if eq .volumeName $key }}
+            - name: cache-volume-{{ $key }}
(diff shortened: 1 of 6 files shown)


jranson:
can you also increment chart version to 1.4.1?

LimitlessEarth:
Yes sir.

deploy/helm/trickster/templates/deployment.yaml
```diff
@@ -58,6 +57,8 @@ spec:
           {{- end }}
         {{- end }}
           volumeMounts:
+            - name: cfg-volume
```
jranson: This looks suspect. If the user has several filesystem caches defined, the iterator below should populate them, so I am not sure if one should also be hardcoded?  Should we

LimitlessEarth: I think this just got removed by accident. Those filesystem caches 

https://github.com/Comcast/trickster/blob/v0.1.10/deploy/helm/trickster/templates/deployment.yaml#L57

This is actually pretty confusing. What is going on in this file...

If the path is `""`, which shouldn't happen ideally, we may have a bunch of volumes mounting over each other? I also don't see where the matching config volume is generated that this maps with is even generated.

LimitlessEarth: It is also kind of funny that unless you mount a pvc/pv in there the "filesystem" caches will just be in the docker "filesystem" which is just node memory...

LimitlessEarth: Yeah, having multiple of the same path try to mount over each other doesn't even work. D:

LimitlessEarth: It looks like it used to be that if your cache was a file system we would, and still do, a volume named `cache-volume`, then provide that in the volume mounts.

https://github.com/Comcast/trickster/pull/189/files#diff-9fa4494b678a0fd83a220b037f2b3e8aR60

That is no longer a thing.

LimitlessEarth:
This has a major rework on some things that became broken as we merged next into master. I apologize that things have gotten here; we had a TODO in the last PR and never completed the TODO. 

So here it is. Tested and working as best I can.

deploy/helm/trickster/templates/deployment.yaml
LimitlessEarth: James and I have discussed offline what my plan was to rectify these discrepancies. 

The TL;DR is to rework how volumes are mapped given we support multiple caches now instead of one.

I'll leave a more details comment.

PR #380, opened 2020-02-22
remove http listener global state from metrics

jranson:
In this PR, we move the metrics HTTP Listener orchestration from the metrics package to the main package, and standardize the way we start up http listeners for tls, http, and metrics (and move this functionality to its own file in main). This allows us to do basic singleton of the Metrics variables in the metrics package, and removes a lot of orchestration (`metrics.Init(conf)`) in all of the other test packages, and allows us to more easily control the metrics listener in case it needs to be turned off or restarted on a different port due to a config reload. I've also moved the usage/version print methods into their own files, and added some tests for them.

### DIFF
--- a/cmd/trickster/listeners.go
+++ b/cmd/trickster/listeners.go
@@ -45,6 +45,10 @@ func startListener(listenerName, address string, port int, connectionsLimit int,
 	err = http.Serve(l, handlers.CompressHandler(router))
 	if err != nil {
 		log.Error("listener stopping", tl.Pairs{"name": listenerName, "detail": err})
+		if exitOnError {
+			// TODO: don't exit when it's a graceful stop
+			os.Exit(1)
+		}
 	}
 	return err
 }
--- a/cmd/trickster/listeners.go
+++ b/cmd/trickster/listeners.go
@@ -20,38 +20,40 @@ import (
 	"sync"
 
 	"github.com/Comcast/trickster/internal/proxy"
-	"github.com/Comcast/trickster/internal/util/log"
+	tl "github.com/Comcast/trickster/internal/util/log"
+
 	"github.com/gorilla/handlers"
 )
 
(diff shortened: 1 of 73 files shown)


jranson:
The addition of tests for `main` (now @ 7% coverage) has resulted in it being included in the coveralls report, and overall coverage is reduced to 96.7% in v1.1.x branch.  I think I'm OK with that for now, knowing that we will get it back above 98% (by continuing to make main more testable) before we cut this release and merge into master.

jranson:
Also, w/ the changes to logging in #378, this patch eventually fail to build without some slight refactoring. So let's prioritize merges of 378, 379, and 380, in ascending order to fix up any problems that might arise along the way after rebasing between merges.

cmd/trickster/listeners.go
```diff
@@ -0,0 +1,57 @@
...
+			os.Exit(1)
+		}
+		return err
+	}
+	log.Info("proxy listener starting",
+		log.Pairs{"name": listenerName, "port": port, "address": address})
+
+	err = http.Serve(l, handlers.CompressHandler(router))
+	if err != nil {
+		log.Error("listener stopping", log.Pairs{"name": listenerName, "detail": err})
```
crandles: should this trigger an `os.Exit(1)` like the above error?

jranson: Serve always returns a non-nil error once it stops serving (see https://golang.org/pkg/net/http/#Serve). So if we were to Exit on err != nil, the Trickster process would exit whenever a config reload (once implemented) required stopping a listener due to port or address change. For now, I'll add the os.Exit back in, but when we implement config reloading, we'll want to make sure to account for that condition and bypass the Exit.

jranson: (added back in with a TODO)

jranson:
#378 and #379 are merged in, and all conflicts resolved. ready for final review!

PR #382, opened 2020-03-02
Use options for tracing exporters

guygrigsby:
Attempts to address#381.

This allows the modification of each tracer exporter by way of an options pattern. In addition, I have added an option for the Jaeger exporter to allow for use of a Jaeger agent _or_ a jaeger collector. The underlying otel SDK allows for one or the other. There is currently no enforcement of only one option and if both are present, the agent will take precedence. 

This has *not* been tested yet against a local jaeger agent. @d-ulyanov is this what you had in mind?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,4 @@ out.log
 
 vendor
 trickster
+tags
--- a/.gitignore
+++ b/.gitignore
@@ -43,5 +43,4 @@ cacheKey.expiration
 out.log
 
 vendor
-trickster
 tags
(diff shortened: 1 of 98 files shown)


d-ulyanov:
Hi, @guygrigsby 

Not exactly what I've meant, but it will be enough for us.

I've meant that it would be nice to have separate structures in code for tracer-specific configuration. 
For example, Jaeger allows us to use remote-strategy sampler, agents for trace data delivery, but such things are totally specific for Jaeger. 
From my point of view it's wrong to add such specific settings to common "ExporterOptions" struct.

Anyway, I'll be happy to see Jaeger agent support in the next release in any implementation :)

PR #387, opened 2020-03-07
use default ports 8480-8483

jranson:


jranson:
Yep, when we get closer to the 1.1 launch, i'll tackle the accompanying helm charts update in the new project (there will be several new config options to incorporate across the various feature enhancements still to come), and we'll have a separate PR on this branch to remove them here.

PR #396, opened 2020-03-15
Helm service ports

Matiasmct:
Istio supports proxying all TCP traffic by default, but in order to provide additional capabilities, such as routing and rich metrics, the protocol must be determined.

https://istio.io/docs/ops/configuration/traffic-management/protocol-selection/

### DIFF
--- a/deploy/helm/trickster/templates/service.yaml
+++ b/deploy/helm/trickster/templates/service.yaml
@@ -44,7 +44,7 @@ spec:
     {{- if .Values.service.nodePort }}
       nodePort: {{ .Values.service.nodePort }}
     {{- end }}
-    - name: metrics
+    - name: http-metrics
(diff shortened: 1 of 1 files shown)


deploy/helm/trickster/templates/service.yaml
```diff
@@ -1,12 +1,17 @@
 apiVersion: v1
 kind: Service
 metadata:
-{{- if .Values.service.annotations }}
+{{- if or .Values.service.annotations .Values.prometheusScrape }}
```
LimitlessEarth: Did you also want to include this set of changes in this PR as well since it was in #395 already?

jranson: should be able to manage this merge as is; noting that the only effective change in this pr is the renaming of `metrics` to `http-metrics`

Issue #397, opened 2020-03-20
Unexpected upstream response


chicknsoup:
I have to restart trickster once every week or so in order to normally query from Grafana. Checking the log file, there are only unexpected upstream response errors, nothing else.
Prometheus queries still ok, restarting trickster fixes  the problem.
```
time=2020-03-20T05:14:12.615544864Z app=trickster caller=proxy/engines/deltaproxycache.go:379 level=error event="unexpected upstream response" statusCode=400
time=2020-03-20T05:14:12.620210271Z app=trickster caller=proxy/engines/deltaproxycache.go:379 level=error event="unexpected upstream response" statusCode=400
time=2020-03-20T05:14:12.624673167Z app=trickster caller=proxy/engines/deltaproxycache.go:379 level=error event="unexpected upstream response" statusCode=400
time=2020-03-20T05:15:12.277004078Z app=trickster caller=proxy/engines/deltaproxycache.go:379 level=error event="unexpected upstream response" statusCode=400
time=2020-03-20T05:15:12.281477037Z app=trickster caller=proxy/engines/deltaproxycache.go:379 level=error event="unexpected upstream response" statusCode=400
time=2020-03-20T05:15:12.28644297Z app=trickster caller=proxy/engines/deltaproxycache.go:379 level=error event="unexpected upstream response" statusCode=400
time=2020-03-20T05:15:12.614298131Z app=trickster caller=proxy/engines/deltaproxycache.go:379 level=error event="unexpected upstream response" statusCode=400
time=2020-03-20T05:15:12.620147753Z app=trickster caller=proxy/engines/deltaproxycache.go:379 level=error event="unexpected upstream response" statusCode=400
time=2020-03-20T05:15:12.62552363Z app=trickster caller=proxy/engines/deltaproxycache.go:379 level=error event="unexpected upstream response" statusCode=400
time=2020-03-20T05:16:12.276595385Z app=trickster caller=proxy/engines/deltaproxycache.go:379 level=error event="unexpected upstream response" statusCode=400
time=2020-03-20T05:16:12.28187261Z app=trickster caller=proxy/engines/deltaproxycache.go:379 level=error event="unexpected upstream response" statusCode=400
time=2020-03-20T05:16:12.286772955Z app=trickster caller=proxy/engines/deltaproxycache.go:379 level=error event="unexpected upstream response" statusCode=400
[...]

Issue #391, opened 2020-03-13
How to avoid max open files limit


DirkFries:
Hi @all,

tonight trickster ran into an open files limit, which was configured to 32k files.

2020-03-13T01:16:10.910147+01:00 host trickster-1.0.1.linux-amd64[1416]: 2020/03/13 01:16:10 http: Accept error: accept tcp [::]:9092: accept4: too many open files; retrying in 1s

Of course one way to fix this is raising the open files limit, but I am wondering which configuration option of trickster affects the number of files the daemon opens.

Does anyone have an idea?

Thanks a lot !

Bye, Dirk

jranson:
Thanks for the report! Can you let us know what type of cache you are using in your Trickster config?

DirkFries:
I am using

 cache_type = 'memory'

jwshieldsGetty:
Hello,

We are using Trickster at my organization;  
Running version 1.0.1 - we've configured our instance of Trickster to use an in memory cache, and have given it a limit of 64k files

Lately, we've been running into this issue, where our instance soft-crashes and floods our logs with errors of "too many open files"

Is there any idea where this may be coming from or what can be done to mitigate/fix it?

jranson:
Thanks for the additional info. Since this is happening on a memory cache, it would point to network connections causing the file descriptors to be exhausted, and the default (32k) should work just fine, so there is a possibility something is leaking. We'll take a look at the pprof diagnostics to try and isolate this and issue a patch.

DirkFries:
Hi @all,

had the same error tonight and collected some infos before restarting

lsof showed 1028 opened files like this:

trickster 29131 trickster 1023u     IPv6           68885247      0t0      TCP host:XmlIpcRegSvc->nagios:35762 (ESTABLISHED)

So the assumption of  jranson  that the problem are not really open files but open network connections is true.

jranson:
@jwshieldsGetty @DirkFries Thanks for all the information on this - we have identified the mutex deadlock that is the root cause of this (causing subsequent inbound requests to hang while waiting on lock acquisition and exhausting the file descriptors), and will be issuing a patch in the next 24 hours.

jranson:
All we have cut release v1.0.3, which is the patch release to address this issue. We would really appreciate if you can try it out and report back if the issue has resolved for yo. Thanks again for the report and your patience while we got it worked out.

jwshieldsGetty:
[...]

PR #410, opened 2020-04-16
V1.0.3 bugfix

jranson:
This addresses 2 separate issues:
- Mitigate mutex deadlocks leading to too many open files #391, #397 
- Fix "context canceled" and false flag "502 Bad Gateway" errors when downstream clients cancel a request.

### DIFF
--- a/internal/cache/index/index.go
+++ b/internal/cache/index/index.go
@@ -45,7 +45,7 @@ type Index struct {
 	name           string                             `msg:"-"`
 	cacheType      string                             `msg:"-"`
 	config         config.CacheIndexConfig            `msg:"-"`
-	bulkRemoveFunc func([]string, bool)               `msg:"-"`
+	bulkRemoveFunc func([]string)                     `msg:"-"`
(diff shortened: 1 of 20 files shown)


internal/cache/index/index.go
```diff
@@ -183,25 +183,37 @@ func (idx *Index) UpdateObject(obj *Object) {
...
-
 		delete(idx.Objects, key)
 		cache.ObserveCacheSizeChange(idx.name, idx.cacheType, idx.CacheSize, idx.ObjectCount)
 	}
+	indexLock.Unlock()
+}
+
+// RemoveObjects removes a list of Objects' Metadata from the Index
+func (idx *Index) RemoveObjects(keys []string, noLock bool) {
+	if !noLock {
```
rockyhardy, line 201: Question:  It appears RemoveObjects is not used without the lock.   Is this logic needed?

jranson: It was intentional in case we need to use the function elsewhere in a later release that does not already have a lock on the index.

PR #411, opened 2020-04-21
support config reload

jranson:
This PR adds full configuration reload capability, via an HTTP reload endpoint (listening on 127.0.0.1 by default) and via SIGHUP. Additionally, this PR adds a `-validate-config` CLI flag, which will validate the configuration and exit.  Example config and Kube deploy files are updated, however, Helm Charts will be published to the separate helm project and not in this project, as the helm directory will be removed in a separate PR.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -23,7 +23,10 @@ import (
 	"bytes"
 	"fmt"
 	"net/http"
+	"os"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/tricksterproxy/trickster/pkg/cache/evictionmethods"
 	cache "github.com/tricksterproxy/trickster/pkg/cache/options"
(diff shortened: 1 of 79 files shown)


pkg/config/config.go
```diff
@@ -684,7 +733,31 @@ func (c *TricksterConfig) Clone() *TricksterConfig {
...
-func (c *TricksterConfig) String() string {
+// IsStale returns true if the running config is stale versus the
+func (c *Config) IsStale() bool {
+
+	if c.Main == nil || c.Main.configFilePath == "" ||
+		time.Now().Before(c.Main.configRateLimitTime) {
+		return false
+	}
+
+	c.Main.stalenessCheckLock.Lock()
```
LimitlessEarth, line 744: Should this lock get moved up to include the read as well?

PR #421, opened 2020-05-11
V1.1.0 beta2

jranson:


pkg/proxy/engines/proxy_request.go
```diff
@@ -237,8 +266,9 @@ func (pr *proxyRequest) prepareUpstreamRequests() {
 		pr.originRequests = make([]*http.Request, 0, l)
 	}
 
-	// TODO: This looks suspicious
-	rsc.OriginConfig.DearticulateUpstreamRanges = true
+	// // TODO: This looks suspicious
+	// rsc.OriginConfig.DearticulateUpstreamRanges = true
```
LimitlessEarth, line 270: ?

pkg/proxy/engines/proxy_request.go
```diff
@@ -139,8 +153,15 @@ func (pr *proxyRequest) Fetch() ([]byte, *http.Response, time.Duration) {
 		handlerName = pc.HandlerName
 	}
 
+	// span := trace.SpanFromContext(pr.upstreamRequest.Context())
+	// _, span := tspan.NewChildSpan(pr.Context(), rsc.Tracer, "Fetch")
+	// if span != nil {
+	// 	pr.upstreamRequest = pr.upstreamRequest.WithContext(trace.ContextWithSpan(pr.upstreamRequest.Context(), span))
+	// 	defer span.End()
+	// }
+
```
LimitlessEarth, line 162: ?

pkg/proxy/engines/proxy_request.go
jranson: I believe this was put in for debugging purposes and forgotten about; so it is commented for Beta2 and will be scrubbed completely in the next release

pkg/proxy/engines/proxy_request.go
jranson: this should be unneeded so as to minimize the span count, but I kept in comments in case it needs to come back, will be scrubbed in next release.

PR #425, opened 2020-05-13
1.1 misc tests

jranson:
this patch adds additional code coverage, cleans up comments/variable names, and corrects a defect when cloning custom path configs where all fields were not included in the clone.


cmd/trickster/config.go
```diff
@@ -219,6 +219,12 @@ func applyCachingConfig(c, oc *config.Config, logger *log.Logger,
 				caches[k] = w
 				continue
 			}
+
+			// if we got to this point, the cache won't be used, so lets close it
+			go func() {
+				time.Sleep(time.Second * time.Duration(c.ReloadConfig.DrainTimeoutSecs))
+				w.Close()
```
LimitlessEarth, line 226: Is this threadsafe?

jranson: In the worst case, it will throw an inconsequential panic on the no-longer-used cache. But point taken and we'll add locking around the entire reload handler in a subsequent PR.

PR #428, opened 2020-05-14
update opentelemetry-go to 0.5.0

jranson:


jranson:
(merging this under the assumption that there is no objection)

PR #429, opened 2020-05-14
only process one reload call at a time

jranson:


pkg/config/config.go
```diff
@@ -101,6 +101,9 @@ type MainConfig struct {
 	// defaults to os.Hostname
 	ServerName string `toml:"server_name"`
 
+	// ReloaderLock is used to lock the config for reloading
+	ReloaderLock *sync.Mutex `toml:"-"`
```
crandles, line 105: Is this optional? Why use a pointer

jranson: all fixed up

PR #445, opened 2020-05-26
Add Cache to proxy Client interface. Enforce origins implement Client

glinton:
This is just some minor tidying that helps ensure the origin clients are implementing the proxy client interface. I assumed/hoped `Cache` was acceptable to add to the interface, but am fine with the possibility it was intentionally left out.


jranson:
@glinton hey greg, if you can fix this typo, the linter should pass and we'll be able to merge. Thanks!

`./pkg/proxy/origins/timeseries_client.go:58: hte ==> the`

PR #454, opened 2020-06-03
V1.1 more post support

jranson:
This patch adds additional post body support for Time Series requests, replaces buffers with readers for representations of those bodies, and uses bitmap comparisons where possible.


pkg/proxy/methods/methods.go
```diff
@@ -20,13 +20,45 @@ package methods
 import "net/http"
 
 const (
+	get uint16 = 1 << iota
```
LimitlessEarth, line 23: This is cool!

pkg/proxy/origins/influxdb/handler_health.go
```diff
@@ -41,11 +44,18 @@ func (c *Client) HealthHandler(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	req, _ := http.NewRequest(c.healthMethod, c.healthURL.String(), nil)
+	fmt.Println(c.healthURL.String())
```
LimitlessEarth, line 47: Rogue println

PR #456, opened 2020-06-03
Clickhouse enhancements

genzgd:
There is some hackiness around BackfillTolerance manipulation, but otherwise this is totally limited to the clickhouse origin package


jranson:
@genzgd i'm going to flip you over to Draft for now, and that will at least silence notifications to anybody watching all activity on the project until we get that test passing!

jranson:
it looks like the timestamps are off by 8 hours expected versus actual, it might be worth checking if there is a timezone -> utc conversion issue.

genzgd:
Yeah I think I have a fix -- I'm so used to every server being in UTC and this is my first work ever in Go time string parsing and formatting I was sloppy with that.

pkg/proxy/origins/clickhouse/clickhouse.go
```diff
@@ -90,39 +90,34 @@ func (c *Client) Router() http.Handler {
...
+		bf = res.OriginConfig.BackfillTolerance
 	}
 
-	var err error
-	trq.Statement, trq.Extent, _, err = getQueryParts(trq.Statement, trq.TimestampFieldName)
-	if err != nil {
+	// Force gzip compression since Brotli is broken on CH 20.3
+	// See https://github.com/ClickHouse/ClickHouse/issues/9969
+	// Clients that don't understand gzip are going to break, but oh well
+	r.Header.Set("Accept-Encoding", "gzip")
```
jranson, line 112: We have a pretty cool way to set headers like this - check out routes.go file in this package, and where we have the main/only path defined in the DefaultPathConfigs(). I think this will work:

```go
// in clickhouse/routes.go

import "github.com/tricksterproxy/trickster/pkg/proxy/headers" // <-- add to imports

paths := map[string]*po.Options{
	"/": {
		Path:           "/",
		HandlerName:    "query",
		Methods:        []string{http.MethodGet, http.MethodPost},
		MatchType:      matching.PathMatchTypePrefix,
		MatchTypeName:  "prefix",
		CacheKeyParams: []string{"query", "database"},
		RequestHeaders: map[string]string{headers.NameAcceptEncoding: "gzip"}, // <-- the new line
	},
}
```

That way we do it in a known place, and it's easily/directly overridden by config.

pkg/proxy/origins/clickhouse/handler_proxy.go
```diff
@@ -24,7 +24,7 @@ import (
 )
 
 // ProxyHandler sends a request through the basic reverse proxy to the origin,
-// and services non-cacheable InfluxDB API calls
+// and services non-cacheable ClickHouse API calls
```
jranson, line 27: 😂

pkg/proxy/origins/clickhouse/parsing.go
```diff
@@ -0,0 +1,342 @@
+/*
+ * Copyright 2020 Comcast Cable Communications Management, LLC
```
jranson, line 2: nit on the 2020 again

pkg/proxy/origins/clickhouse/parsing_test.go
```diff
@@ -0,0 +1,163 @@
+/*
+ * Copyright 2020 Comcast Cable Communications Management, LLC
```
jranson, line 2: ... hugh downs and barbara walters would approve ...

pkg/proxy/origins/clickhouse/routes.go
```diff
@@ -27,7 +27,7 @@ import (
 func (c *Client) registerHandlers() {
 	c.handlersRegistered = true
 	c.handlers = make(map[string]http.Handler)
-	// This is the registry of handlers that Trickster supports for InfluxDB,
+	// This is the registry of handlers that Trickster supports for ClickHouse,
```
jranson, line 30: i appreciate the attention to detail on this!

pkg/proxy/origins/clickhouse/model.go
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 Comcast Cable Communications Management, LLC
+ * Copyright 2020 Comcast Cable Communications Management, LLC
```
jranson, line 2: 2020

pkg/proxy/origins/clickhouse/model_test.go
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 Comcast Cable Communications Management, LLC
+ * Copyright 2020 Comcast Cable Communications Management, LLC
```
jranson, line 2: uno mas

pkg/proxy/origins/clickhouse/series_test.go
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 Comcast Cable Communications Management, LLC
+ * Copyright 2020 Comcast Cable Communications Management, LLC
```
jranson, line 2: broken record

pkg/proxy/origins/clickhouse/clickhouse.go
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 Comcast Cable Communications Management, LLC
+ * Copyright 2020 Comcast Cable Communications Management, LLC
```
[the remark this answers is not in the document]
genzgd: Sure. :). Copyright dates are utterly irrelevant legally anyway (copyright automatically dates from when the work was created), so whatever makes people happy.

pkg/proxy/origins/clickhouse/clickhouse.go
```diff
@@ -90,39 +90,34 @@ func (c *Client) Router() http.Handler {
 
 // ParseTimeRangeQuery parses the key parts of a TimeRangeQuery from the inbound HTTP Request
 func (c *Client) ParseTimeRangeQuery(r *http.Request) (*timeseries.TimeRangeQuery, error) {
-
-	trq := &timeseries.TimeRangeQuery{Extent: timeseries.Extent{}}
-	trq.TemplateURL = urls.Clone(r.URL)
-	qi := trq.TemplateURL.Query()
+	qi := r.URL.Query()
```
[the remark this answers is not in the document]
genzgd: ClickHouse POST queries can get a little messy -- for example, the body can be an actual query with substitutions from query parameters.  Since it's not in our current use cases (Grafana or our data access API) I'd prefer to defer that for now, although it would be nice to make it work eventually.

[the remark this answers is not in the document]
genzgd: However, we still want to proxy them, so I now explicitly send POSTs directly to the origin, and at least it won't be broken.

pkg/proxy/origins/clickhouse/clickhouse.go
genzgd: so I tried this, and Grafana blew up?  I wonder if we were sending two Accept-Encoding headers and ClickHouse got confused and sent back Brotli but said it was sending back gzip (that's kind of what it looked like, Trickster logs got horribly ugly too).  I'm happy to revisit this since using the route definition is much cleaner, but the whole thing is really a hack around broken ClickHouse Brotli and I'm hoping they'll fix the bug and I can remove it completely, so I don't have a lot of motivation to figure out the details.

PR #467, opened 2020-07-31
add base sql select parser and lexer

jranson:


pkg/parsing/lex/run_state.go
```diff
@@ -0,0 +1,173 @@
...
+	for strings.ContainsRune(valid, rs.Next()) {
+		b = true
+	}
+	rs.Backup()
+	return b
+}
+
+// // Drain drains the output so the lexing goroutine will exit.
+// // Called by the parser, not in the lexing goroutine.
+// func (rs *RunState) Drain() {
```
LimitlessEarth, line 104: Ded code?

pkg/parsing/sql/sql.go
```diff
@@ -0,0 +1,156 @@
...
+)
+
+// ErrUnsupportedVerb represents an error of type "unsupported verb"
+var ErrUnsupportedVerb = errors.New("unsupported verb")
+
+// ErrUnsupportedClause represents an error of type "unsupported clause"
+var ErrUnsupportedClause = errors.New("unsupported clause")
+
+// // Parser defines the SQL Parser Interface
+// type Parser interface {
```
LimitlessEarth, line 41: ded?

pkg/parsing/lex/run_state.go
jranson: We're unsure if the Drain func will be needed yet. Once ClickHouse and InfluxDB implement the parser, we might need to run a drain in the event of a lexing error, so that orphan channels don't lie around indefinitely.

pkg/parsing/sql/sql.go
jranson: will remove this one since it was relocated to timeseries/sqlparser

PR #474, opened 2020-08-15
bump imported module versions

jranson:


jranson:
(merging without objection)

PR #475, opened 2020-08-17
2.0 use common ts format

jranson:
this PR implements the common time series format for Prometheus, InfluxDB, ironDB, and ClickHouse. Additionally (and unavoidably), ClickHouse also implements the new sql parser/lexer for query parsing. there are a lot of `TODO` items that will be addressed in subsequent PRs, but for now things function and all tests pass.


pkg/proxy/origins/clickhouse/model/model.go
```diff
@@ -0,0 +1,572 @@
...
+	return UnmarshalTimeseriesReader(buf, trq)
+}
+
+// UnmarshalTimeseriesReader converts a JSON blob into a Timeseries via io.Reader
+func UnmarshalTimeseriesReader(reader io.Reader, trq *timeseries.TimeRangeQuery) (timeseries.Timeseries, error) {
+	return nil, nil
+}
+
+// /*
+// {"meta":[{"name":"t","type":"UInt64"},{"name":"cnt","type":"UInt64"}{"name":"meta1","type":"UInt16"},{"name":"meta2","type":"String"}]"data":[{"cnt":"12648509","meta1":200,"meta2":"value2","t":"1557766080000"},{"cnt":"10260032","meta1":200,"meta2":"value3","t":"1557766680000"},{"cnt":"1","meta1":206,"meta2":"value3","t":"1557767280000"}],"rows":3}
```
LimitlessEarth, line 111: Looks like a ton of ded code?

pkg/proxy/paths/options/options.go
```diff
@@ -188,5 +187,56 @@ func (o *Options) Merge(o2 *Options) {
...
+	o.Custom = strutil.Unique(o.Custom)
 }
+
+var pathMembers = []string{"path", "match_type", "handler", "methods", "cache_key_params",
+	"cache_key_headers", "default_ttl_secs", "request_headers", "response_headers",
+	"response_headers", "response_code", "response_body", "no_metrics", "collapsed_forwarding",
+	"req_rewriter_name",
+}
+
+// func ProcessTOML() {
```
LimitlessEarth, line 199: More ded code?

jranson: this is implemented / cleaned up in a forthcoming PR (see https://github.com/jranson/trickster/blob/2.0-options-cleanup-01/pkg/proxy/paths/options/options.go#L206)

pkg/proxy/origins/clickhouse/model/model.go
jranson: while we wrote the marshalers for clickhouse (the part you noticed is commented out), the final common format interface meant they needed to be refactored. that appears to have been dropped on the floor hence the comments as a placeholder. will fix up in a subsequent PR prior to the first pre-release of 2.0

pkg/proxy/engines/deltaproxycache.go
```diff
@@ -450,38 +457,45 @@ func DeltaProxyCacheRequest(w http.ResponseWriter, r *http.Request) {
 		rts.Merge(false, ffts)
 	}
 	rts.SetExtents(nil) // so they are not included in the client response json
-	rts.SetStep(0)
-	rdata, err := client.MarshalTimeseries(rts)
+	//rts.SetTimeRangeQuery(&timeseries.TimeRangeQuery{})
```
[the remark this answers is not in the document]
jranson: this is ok as-is. the removed lines are not needed after all ts origins are now converted to comment format, and the commented line was needed for some origins during the transition. i've left it here for now just in case it will need to be revived as things evolve in 2.0.

PR #476, opened 2020-08-23
2.0 options cleanup 01

jranson:
unfortunately i could not rebase or squash this prior to PR, but it's a lot of search/replace copy/paste with no new functionality

this renames all NewOptions() functions to just New() (since they're all now relocated to their own options packages), converts all configs defined in seconds to use milliseconds, and does other minor tidying up.


cmd/trickster/conf/example.conf
```diff
@@ -316,19 +316,19 @@ listen_port = 8480
     ## Default list is provided here:
     # compressable_types = [ 'text/javascript', 'text/css', 'text/plain', 'text/xml', 'text/json', 'application/json', 'application/javascript', 'application/xml' ]
 
-    ## timeout_secs defines how many seconds Trickster will wait before aborting and upstream http request. Default: 180s
-    # timeout_secs = 180
+    ## timeout_ms defines how long Trickster will wait before aborting and upstream http request. Default: 180s
```
LimitlessEarth, line 319: Do you want to leave the default comment in s vs ms?

cmd/trickster/conf/example.conf
```diff
@@ -316,19 +316,19 @@ listen_port = 8480
...
+    # timeout_ms = 180000
 
-    ## keep_alive_timeout_secs defines how long Trickster will wait before closing a keep-alive connection due to inactivity
+    ## keep_alive_timeout_ms defines how long Trickster will wait before closing a keep-alive connection due to inactivity
     ## if the origin's keep-alive timeout is shorter than Trickster's, the connect will be closed sooner. Default: 300
```
LimitlessEarth, line 323: same here.

PR #505, opened 2020-10-18
2.0 fix log caller

jranson:
this fixes an issue where the `caller` field in a 2.0 log was never accurate; with this change, it is 100% of the time. the meat of this change is in logging.go (requiring the caller to pass in it's stack.Call for Debug, Info, Warn, etc. All other files changed are solely for conforming to the updated logging function prototypes.


LimitlessEarth (approved):
For sure this is not a change that could be made in the logging package rather than on every line?

jranson:
@LimitlessEarth done and merged in #507

PR #508, opened 2020-10-21
2.0 clickhouse updates

jranson:
this PR re-implements clickhouse via the Common Time Series Format in 2.0


pkg/proxy/engines/deltaproxycache.go
```diff
@@ -67,7 +68,8 @@ func DeltaProxyCacheRequest(w http.ResponseWriter, r *http.Request, modeler *tim
 	trq, rlo, canOPC, err := client.ParseTimeRangeQuery(r)
 	if err != nil {
 		if canOPC {
-			// TODO - custom TTL ?
+			tl.Debug(rsc.Logger, "could not parse time range query, using object proxy cache", tl.Pairs{"error": err.Error()})
+			rsc.AlternateCacheTTL = time.Second * 30 // TODO: make configurable
```
LimitlessEarth, line 72: Fine to leave this as is for now?

LimitlessEarth (approved):

pkg/proxy/engines/deltaproxycache.go
jranson: yep, a subsequent PR will implement the alternate cache ttl as a config option.

PR #528, opened 2021-01-28
always display times in UTC when formatting

jranson:


jranson:
@samuelgmartinez i found just a few more places where we are using Time.Format() without first setting the location to UTC. Just to be on the safe side, I inserted the UTC() location modifier ahead of the formatter. If you would not mind giving this a review, that would be really great!

LimitlessEarth (approved):

samuelgmartinez:
It builds fine and passes all the tests on my end. Also, I look for more instances, just in case there were some of them missed by this change, and I couldn't find any.

LGTM.

PR #530, opened 2021-01-29
[Minor] Correct Compose sample config

ntk148v:


jranson (approved):
Thanks so much for coming across these and submitting the patches. Much appreciated, and sorry that my tired eyes did not catch them when we published first the docker compose!  The provider -> origin_type is good for now, since the docker-compose is still based on the 1.1 codebase, so I appreciate that as well. Once we move the docker-compose to a 2.0 build, I will change them back for compatibility.

jranson:
btw, @ntk148v if you would not mind signing the Contributor License Agreement, which just ensures that you acknowledge you are donating this contribution to a Comcast project, the system will allow me to merge it. Thanks again!

Issue #531, opened 2021-02-03
Trickster's health endpoint infinitely appends to the "Via" header on upstream requests, ultimately causing an HTTP 431 error


tminor:
Hi!

I recently set up Trickster behind a load balancer that checks Trickster's `/trickster/health/default` endpoint every second or so. After a Trickster node has been up for around 5 days, the load balancer marks the node as down. My first instinct was to check via `curl`, which returned HTTP 431. Restarting Trickster solved the problem, but it ultimately resurfaced. I had a look at a Grafana dashboard for `node_exporter` and found this:

![trickster](https://user-images.githubusercontent.com/11579850/106806734-4e213b80-6636-11eb-919b-340f9c4fd5aa.png)

The curious looking pattern is from the loopback/localhost interface and the sudden drops correspond with Trickster restarting; Trickster's default backend is Promxy running on the same host (via localhost). After seeing the above dashboard panel, I used `tcpdump` to examine the traffic between Trickster and Promxy. This revealed giant `Via` headers in the requests from Trickster to Promxy. I restarted Trickster and noted that the value for `Via` was appended to with every request to `/trickster/health/default`. Thinking this was an error on my part, I configured Trickster on my laptop and pointed it directly at a Prometheus server with the simplest possible configuration:

```toml
[caches.default]
cache_type = "memory"

[frontend]
listen_port = 8480

[logging]
log_level = "info"

[backends.default]
provider = "prometheus"
origin_url = "http://prometheus:9090" # Changed to a real value, of course.
```

During the next phase of troubleshooting, I used `dlv` to pinpoint the source of the problem. I'm very far from proficient with Go, but I believe [this line](https://github.com/tricksterproxy/trickster/blob/main/pkg/backends/prometheus/handler_health.go#L48) is the culprit. Based on what I've read, the right side of the assignment is passing a pointer. Subsequent operations (such as in [`SetVia()`](https://github.com/tricksterproxy/trickster/blob/main/pkg/proxy/headers/forwarding.go#L200-L204)) cause unintended side effects. Please correct me if I'm wrong!

I was able to ameliorate the problem with this patch:

```patch
diff --git a/pkg/backends/prometheus/handler_health.go b/pkg/backends/prometheus/handler_health.go
index 59b52a2..a1eaebf 100644
--- a/pkg/backends/prometheus/handler_health.go
+++ b/pkg/backends/prometheus/handler_health.go
@@ -19,6 +19,7 @@ package prometheus
 import (
[...]

PR #532, opened 2021-02-03
Fix health endpoint HTTP 431 bug (#531)

tminor:


jranson:
@tminor  Thanks for this great analysis and root causing, and putting in the fix. We really appreciate the time and contribution.  This PR will fix up `main` which is the nextgen codebase, so will not make it out to the masses immediately. The stable branch is `v1.1.x`, so I will backport this fix and cut a release today. The good news on your question about doing the health check differently in InfluxDB versus the other providers is that I am working on a [unified healthcheck package](https://github.com/jranson/trickster/tree/2.0-alb/pkg/backends/healthcheck) that will eliminate the backend providers needing to implement this handler. Separately, I will add calls to the health check URLs as part of our load test harness, so that we'll be able to catch any issues with calls to health check endpoints in the future.

jranson (approved):

tminor:
Fantastic! Thanks for the explanation!

PR #534, opened 2021-02-10
2.0 base client

jranson:
This patch introduces a base Backend and base TimeseriesBackend that provide boilerplate functionality and accessors to all Backends; these were previously implemented separately in each backend, so they are now implemented just once. A separate PR will implement a common healthcheck for all backends, so for now I have left the provider-specific healthcheck-related implementations alone rather than moving to the base backend.


pkg/proxy/headers/forwarding_test.go
```diff
@@ -46,7 +46,7 @@ var testHops1 = &Hop{
 var testHops2 = &Hop{
 	RemoteAddr: "5.6.7.8",
 	Scheme:     "http",
-	Host:       "foo.com",
+	Host:       "fbo.com",
```
LimitlessEarth, line 49: Heh

LimitlessEarth (approved):

pkg/proxy/headers/forwarding_test.go
jranson: lol search / replace mishap artifact. i'll restore that one pre-merge

PR #535, opened 2021-02-11
2.0 common health checker

jranson:
this patch introduces a common health checker package used by all Backends, and provides a new mechanism to perform periodic health checks by setting an interval value in the backend options.


pkg/backends/irondb/health.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+// DefaultHealthCheckConfig returns the default HealthCheck Config for this backend provider
+func (c *Client) DefaultHealthCheckConfig() *ho.Options {
+	o := ho.New()
+	u := c.BaseUpstreamURL()
+	o.Scheme = u.Scheme
+	o.Host = u.Host
+	o.Path = u.Path
+	// TODO: Provide health check query
```
LimitlessEarth, line 30: Want to leave this in?

LimitlessEarth (approved):

pkg/backends/irondb/health.go
jranson: yep, this one is a carryover and will need vendor imput

PR #550, opened 2021-03-11
2.0 backfill tolerance update

jranson:
This patch changes the backfill tolerance pattern, which is used to prevent preliminary real-time data from being cached until it is old enough to be considered final, from an eager load to a lazy load pattern.

The previous implementation would simply not cache data new enough to fall in the backfill tolerance window. As a result, every single real-time request to Trickster for the given Backend would result in a partial cache hit at best, since some data would always be excluded from the cache and need to be fetched. There was no possibility of a Cache Hit for time series requests on backends with a backfill tolerance set.

The new approach improves the methodology by caching the the preliminary data and serving to subsequent requests, so long as the lookup is otherwise a cache hit. once a request is made that has a cache status of partial hit - meaning some data requested by the client, but in a different time range than the preliminary data, needs to be fetched - then the resulting delta proxy request will also include any preliminary data ranges that were part of the client range, so as to refresh them.

In this way, preliminary real-time data are eventually updated to their final values, so long as subsequent real-time requests continue to come through to create qualifying partial hits. This allows backends with backfill tolerance to have a Cache Hit Rate comparable to Backends without a backfill tolerance set, while ensuring preliminary data still gets updated.

Separately, a new option was added called `BackfillTolerancePoints` (in addition to the existing `BackfillToleranceMS`). The new option will calculate an alternate backfill tolerance of `BackfillTolerancePoints * request.StepValue`, and the delta proxy cache will choose whichever tolerance value is greater, based on the query's step value. This way, no matter if a chart is a 1yr look with 1wk steps, or 10m look with a 5s step, backfill tolerance will work proportionately.

### DIFF
--- a/pkg/timeseries/modeler_test.go
+++ b/pkg/timeseries/modeler_test.go
@@ -20,9 +20,6 @@ import "testing"
 
 func TestNewModeler(t *testing.T) {
 
-	// 	// UnmarshalerFunc describes a function that unmarshals a Timeseries
-	// type UnmarshalerFunc func([]byte, *TimeRangeQuery) (Timeseries, error)
-
 	f := func([]byte, *TimeRangeQuery) (Timeseries, error) {
 		return nil, nil
 	}
--- a/pkg/timeseries/modeler_test.go
+++ b/pkg/timeseries/modeler_test.go
@@ -18,34 +18,23 @@ package timeseries
 
 import "testing"
 
-func TestFieldDefinitionClone(t *testing.T) {
+func TestNewModeler(t *testing.T) {
 
-	fd := FieldDefinition{
-		Name:     "test",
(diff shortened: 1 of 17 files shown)


pkg/timeseries/modeler_test.go
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package timeseries
+
+import "testing"
+
+func TestNewModeler(t *testing.T) {
+
+	// 	// UnmarshalerFunc describes a function that unmarshals a Timeseries
+	// type UnmarshalerFunc func([]byte, *TimeRangeQuery) (Timeseries, error)
```
LimitlessEarth, line 24: Still want?

LimitlessEarth (approved):

PR #554, opened 2021-03-21
2.0 compression

jranson:
this patch replaces the gorilla compression handler with the new Trickster encoding handler. In addition to supporting gzip and deflate, this introduces support for zstd and brotli.  Additionally, this will implement compression between Trickster and the Origin, and replaces the internal cache compression with brotli, unless the object was already compressed in a different encoding.


pkg/encoding/brotli/brotli.go
```diff
@@ -0,0 +1,52 @@
...
+
+// Decode returns the inflated version of the gzip-deflated byte slice
+func Decode(in []byte) ([]byte, error) {
+	br := brotli.NewReader(bytes.NewReader(in))
+	return io.ReadAll(br)
+}
+
+// Encode returns the gzip-deflated version of the byte slice
+func Encode(in []byte) ([]byte, error) {
+	buf := bytes.NewBuffer(nil)
```
LimitlessEarth, line 36: Maybe it's silly but you could save some rummaging around in the buffer code if you change this to
`buf := bytes.NewBuffer(len(in))`. It's not a big deal but you get to skip a couple of lines since you already know the size of the thing you are writing.

pkg/encoding/gzip/gzip.go
```diff
@@ -0,0 +1,60 @@
...
+	gr, err := gzip.NewReader(bytes.NewReader(in))
+	if err != nil {
+		return []byte{}, err
+	}
+	return io.ReadAll(gr)
+}
+
+// Encode returns the gzip-deflated version of the byte slice
+func Encode(in []byte) ([]byte, error) {
+	buf := bytes.NewBuffer(nil)
```
LimitlessEarth, line 39: Ditto here as above.

LimitlessEarth (approved):

Issue #559, opened 2021-03-31
ALB with time series merge. Don't merge result.


bessonovevgen:
Hello!
I am getting strange behavior when using alb with tsm.

On request with unique labels trickster return two metric:
<details>
  <summary>two metric</summary>
```
curl -k 'https://trickster-test.local/api/v1/query_range?query=sum%20by%20(cluster%2Ccontainer)%20(container_cpu_usage_seconds_total%7Bcontainer%3D%22test-container-01%22%2Cnamespace%3D%22test-namespace%22%7D)&start=1617019260&end=1617019260&step=15' |
jq .
{
  "status": "success",
  "data": {
    "resultType": "matrix",
    "result": [
      {
        "metric": {
          "cluster": "cluster-01",
          "container": "test-container-01"
        },
        "values": [
          [
            1617019260,
            "783.069233226"
          ]
        ]
      },
      {
        "metric": {
          "cluster": "cluster-02",
          "container": "test-container-01"
        },
        "values": [
          [
            1617019260,
            "1558.804951745"
          ]
        ]
      }
    ]
  }
}
```
</details>

But if I try to sum this metric I get a one of previous metric instead of sum:
<details>
  <summary>sum of two metric</summary>
```
curl -k 'https://trickster-test.local/api/v1/query_range?query=sum%20by%20(container)%20(sum%20by%20(cluster%2Ccontainer)%20(container_cpu_usage_seconds_total%7Bcontainer%3D%22test-container-01%22%2Cnamespace%3D%22test-namespace%22%7D))&start=1617019260&end=1617019260&step=15' | jq .
{
  "status": "success",
  "data": {
    "resultType": "matrix",
    "result": [
      {
        "metric": {
          "container": "test-container-01"
        },
        "values": [
          [
            1617019260,
            "1558.804951745"
          ]
        ]
      }
    ]
  }
}
```
</details>

Maybe my config for backends is wrong and trickster use backends (vmetrics-zone-01,vmetrics-zone-02) as redundant ?
<details>
  <summary>config file</summary>
frontend:
  listen_port: 8480
backends:
  vmetrics-zone-01:
    provider: prometheus
    origin_url: http://vmetrics-test-1.zone01.local:8428
    prometheus:
      labels:
        region: zone-01

  vmetrics-zone-02:
    provider: prometheus
    origin_url: http://vmetrics-test-1.zone01.local:8428
    prometheus:
      labels:
        region: zone-02

  prom-alb-all:
[...]

PR #563, opened 2021-04-22
2.0 label insertion fixes

jranson:
this patch updates the method used to insert labels via configuration to use interface-driven calls at payload write time, fixing #559


pkg/backends/alb/options/options.go
```diff
@@ -60,7 +59,7 @@ func (o *Options) Clone() *Options {
 	}
 
 	c.Pool = copiers.CopyStrings(o.Pool)
-	c.MergeablePaths = copiers.CopyStrings(o.MergeablePaths)
+	// c.MergeablePaths = copiers.CopyStrings(o.MergeablePaths)
```
LimitlessEarth, line 62: No need?

LimitlessEarth (approved):

PR #573, opened 2021-05-26
Shard Time Series Requests in Trickster 2.0

jranson:
This patch extends the Time Series Delta Proxy Cache with the ability to shard requests to an origin for uncached data into smaller, concurrent requests, for which the results are reconstituted back into a single dataset. The manner in which the shard sizes and time ranges are determined is controlled by 3 new `shard_*` backend configuration values as documented herein.


LimitlessEarth (approved):
This is awesome

nbroyles:
Neat! Could also be useful to take the same approach for a `kmiss` in addition to `rmiss` and `phit`.

jranson:
@nbroyles good news! this does actually handle `kmiss` as well, since the base fetchTimeSeries func [has been updated](https://github.com/jranson/trickster/blob/c754bd7fffbff369cda329af5dfd9727c8977f6c/pkg/proxy/engines/deltaproxycache.go#L526) to perform the extent list Splice, and use the new fetchExtents() method. So prior to this PR, kmiss had one way to grab a full timeseries, and phit/rmiss had another way to do whatever was needed in parallel to grab partial or disparate data. Now the "do whatever needed in parallel" logic is encapsulated into [fetchExtents](https://github.com/jranson/trickster/blob/c754bd7fffbff369cda329af5dfd9727c8977f6c/pkg/proxy/engines/deltaproxycache.go#L580), which is now called in all non-hit scenarios.

PR #591, opened 2021-11-13
Fix docker hub org in docker compose file

tamalsaha:
Signed-off-by: Tamal Saha <tamal@appscode.com>


jranson:
@tamalsaha Hey Tamal - `trickstercache` is actually the correct repo, but it looks like while the I had the new org created, I did not mirror over the images form `tricksterproxy`. Totally my fault, and it's corrected now, if you wouldn't mind giving it a go.

PR #596, opened 2022-01-26
fix: small error in docs

0xflotus:


jranson (approved):

jranson:
@0xflotus thanks very much for catching these!

PR #610, opened 2022-08-22
add jake to contributors

jranson:


jakenichols2719 (changes requested):
Need to update table separators (---) to match new number of columns on lines 12 and 22 to make sure tables render correctly.

jakenichols2719 (approved):
Looks good!

PR #612, opened 2022-08-23
Update pkg versions and  go version

jranson:
this bumps the go version to 1.19, and updates all dependent modules to their latest versions. following package update, the logging package would periodically panic due to concurrent map writes in one of the dependent packages, so I added a mutex to the logger to avoid this scenario.


jakenichols2719 (approved):
Glad to be moving to 1.19. I am a little bummed about the mutex implementation for logging, and would like to keep an eye on that package so that the change can be reverted if/when that issue is fixed--I think looking back on it later, it may be confusing as to why that's required.

PR #614, opened 2022-09-14
Replaced GOARCH and GO_VER ldflags

### LINKED ISSUE #606: `goVersion` & `goArch` ldflags redundant - can be pulled more accurately from `runtime`
[`runtime.Version()`](https://pkg.go.dev/runtime#Version)

[`runtime.GOARCH`](https://pkg.go.dev/runtime#pkg-constants)

[As far as I can tell](https://github.com/trickstercache/trickster/blob/main/Makefile#L28) you're looking for the `running program's architecture target` and that value will match. 

(If you were really looking for the original builder's architecture e.g. if you crosscompiled from an amd64 -> arm64 and wanted it to show amd64 when running you should be using `GOHOSTARCH` rather than `GOARCH` since the latter is the target)

The version is the same for either the build host or the build target

jakenichols2719:
GOARCH is now GOHOSTARCH; references in logging to GOARCH now refer instead to GOHOSTARCH. Runtime functionality that relies on runtime architecture is already using it, so no changes to packages that make decisions based on architecture. GOARCH was used for logging only.

GO_VER is removed; package `runtime` now has `GoVersion` which is set with other runtime variables in main. All references to `applicationGoVersion` now reference `trickster.runtime.GoVersion`.

Fixes #606 

Signed-off-by: jakenichols2719 <jnichols2719@protonmail.com>


jakenichols2719:
The test being failed is regarding the example version print; since the test now pulls the go version on startup, rather than being passed on build, the field for goVersion is not empty. Looking into the best resolution for this.

Issue #609, opened 2022-08-16
Is there any plans to implement automatic service discovery to identify backend targets?


teja-v1:
As mentioned in the subject, does Trickster have plans to implement service discovery, I'm thinking about Kubernetes-based service discovery, Prometheus design can be taken as inspiration (see https://prometheus.io/docs/prometheus/latest/configuration/configuration/#kubernetes_sd_config).

jranson:
hey @tejaswiniVadlamudi, thanks for requesting this feature! @jakenichols2719 is currently working on implementing it, and has submitted a draft-status PR #615 to allow the community to review the approach while development is ongoing. 

Right now, we're supporting kube api queries, but the goal is to eventually support multiple methods of service discovery (e.g., querying etcd, querying a DNS record, etc.) based on the most popular use cases.

Let us know if you have any questions or feedback!

jakenichols2719:
Hello! Just wanted to give an update on how this is going. PR #675 addresses autodiscovery for Kubernetes services (externally) or pods (internally) with specific names or labels. Any discovered backend will be added to the ALB with discovery configured.

petemounce:
Are there plans to support discovery in the absence of a kubernetes cluster? I'm thinking here of equivalents to prometheus' ec2, gce, and consul SDs.

jranson:
reopening this. autodiscovery feature is implemented now for kube, a/aaa/srv, and prom-style file_ds. A follow-on PR will cover ec2, gce, consul and any others that might be useful.

jranson:
autodiscovery now supports ec2, ecs, gce, azure, consul, nomad and docker. [documentation](https://github.com/trickstercache/trickster/blob/main/docs/alb-autodiscovery.md)

PR #615, opened 2022-09-29
Autodiscovery

jakenichols2719:
Addresses #609 

Implementation of automatic service discovery as an interface, with current implementation supporting discovery of kubernetes resources in a cluster pointed to by the local kubeconfig. Draft PR while implementing new discovery methods and more thorough testing--want to have the current work publicly-facing to allow for easier testing and feedback.


jakenichols2719:
A combination of old, unmerged code and forgetting to sign off on a couple of commits has made this PR too unwieldy. I have fixed locally and will make a new PR that is more up to date.

PR #616, opened 2022-10-03
build and report correct target os/arch

jranson:
This patch updates the `initLogger()` and `PrintVersion()` functions to use the correct flags for reporting a build's target os, arch, and go version at runtime from the compiled binary.

Signed-off-by: James Ranson <james@ranson.org>


LimitlessEarth (approved):

jranson:
this finishes out #606

PR #630, opened 2023-01-23
remove gorilla/mux (deprecated)

### LINKED ISSUE #628: Deprecate Gorilla Mux
github.com/gorilla/mux has been archived so we should deprecate it in Trickster.

jranson:
This patch duplicates `gorilla/mux`, which is archived / no longer supported, into a new `router` package; and implements a `Router` interface atop the original gorilla `*Router` struct (now lowercase r `*router`)

fixes #628 

Signed-off-by: James Ranson <james@ranson.org>


jakenichols2719 (approved):

pkg/router/router.go
```diff
@@ -13,38 +11,45 @@ import (
...
-//     }
+//	func init() {
+//	    http.Handle("/", router)
+//	}
 //
 // This will send all incoming requests to the router.
-type Router struct {
+type Router interface {
+	ServeHTTP(http.ResponseWriter, *http.Request)
+	PathPrefix(string) *Route
```
tamalsaha, line 34: It feels weird to see pointer to an interface.

jranson:
hi @tamalsaha - not sure i understand here. While `Router` is an interface, `Route` is actually a struct. Just FYI, we'll be switching to <https://github.com/julienschmidt/httprouter> before we cut the next beta release, since it is much more performant than gorilla/mux.

tamalsaha:
> While Router is an interface, Route is actually a struct.

Ah! Sorry, I missed that `Router` and `Route` thing. I understand now.

> Just FYI, we'll be switching to [julienschmidt/httprouter](https://github.com/julienschmidt/httprouter)

Wow! It must be a major rework of the project.

PR #635, opened 2023-01-25
Flux Support for InfluxDB backends

jakenichols2719:
Adds support for Flux queries. Currently in draft pending additional testing; branch introduces new duration parsing and flux parsing as possible points of failure and/or performance bottlenecks.

If you have an InfluxDB solution with Flux and are able to test this, please let me know if there are any issues, and keep an eye out for any improvements!


pkg/backends/influxdb/flux/parser.go
```diff
@@ -0,0 +1,123 @@
...
+			}
+		}
+	}
+	if start.IsZero() {
+		return timeseries.Extent{}, FluxSemantics("range() expressions require a valid start argument")
+	}
+	return timeseries.Extent{Start: start, End: stop}, nil
+}
+
+func tryParseTimeField(s string) (t time.Time, ok bool) {
```
jranson, line 88: Let's replace the bool return vals in these functions with error. since the 3 `try` subfunctions actually check for errors and only return false in the event of an error, we should just pass the original error all the way back through to the calling function (parseRangeFilter), esp since parseRangeFilter _already_ passes an error back to its calling function (ParseQuery). No need to go from err > bool > custom-created error. Remember, in applications like trickster that can handle tens of thousands of requests per second, minimizing memory churn is important to performance.

Also, since the rest of the codebase does _not_ use named response variables in the prototype, we should maintain that stylistic choice across this new package.

pkg/backends/influxdb/flux/errors.go
```diff
@@ -0,0 +1,33 @@
+package flux
+
+import "fmt"
+
+type ErrFluxSyntax struct {
```
jranson, line 5: the `Err` prefix should be on variables and functions that represent error Objects, not their underlying struct names. This should be `FluxSyntaxError` ...

pkg/backends/influxdb/flux/errors.go
```diff
@@ -0,0 +1,33 @@
...
+	token string
+	rule  string
+}
+
+func FluxSyntax(token, rule string) *ErrFluxSyntax {
```
jranson, line 10: ... and _this_ should be called ErrFluxSyntax since it provides the error Object  :-D

pkg/backends/influxdb/flux/errors.go
```diff
@@ -0,0 +1,33 @@
...
+		token: token,
+		rule:  rule,
+	}
+}
+
+func (err *ErrFluxSyntax) Error() string {
+	return fmt.Sprintf("flux syntax error at '%s': %s", err.token, err.rule)
+}
+
+type ErrFluxSemantics struct {
```
jranson, line 21: Samesies on the naming

pkg/backends/influxdb/flux/parser.go
```diff
@@ -0,0 +1,123 @@
...
+		reader: reader,
+	}
+}
+
+func (p *Parser) ParseQuery() (*Query, error) {
+	valid := false
+	r := bufio.NewReader(p.reader)
+	ln := 0
+	q := &Query{}
+	for line, err := r.ReadString('\n'); ; line, err = r.ReadString('\n') {
```
jranson, line 30: since a flux query is generally very small (almost always under 1k), I'm not sure if it's necessary to iterate over the query lines here. To avoid memory churn and unnecessary checks for the newline character as an interruptor, I think we can just io.ReadAll() and apply the subsequent string.Contains() checks to the whole query string. That would be a bit more performant at scale.

pkg/backends/influxdb/flux/parser.go
```diff
@@ -0,0 +1,123 @@
...
+		}
+		// Check for range
+		if strings.Contains(line, "range") {
+			q.Extent, err = parseRangeFilter(line)
+			if err != nil {
+				return nil, err
+			}
+			valid = true
+		}
+		q.Statement += line + " "
```
jranson, line 52: using the io.ReadAll() approach, be sure to just set the q.Statement = string(b) instead of using incremental appends

pkg/backends/influxdb/flux/parser.go
```diff
@@ -0,0 +1,123 @@
...
+		if err != nil {
+			if errors.Is(err, io.EOF) {
+				break
+			} else {
+				return nil, err
+			}
+		}
+		// Check for line 0 "from"
+		if ln == 0 {
+			if !strings.Contains(line, "from") {
```
jranson, line 40: i would recommend checking for ` from(` instead of `from` (prefix with a space, suffix with left paren) to avoid some possible false positives.

pkg/util/timeconv/duration/duration.go
```diff
@@ -0,0 +1,161 @@
+package duration
```
jranson, line 1: since we already have a `ParseDuration` func with the same prototype in the parent `timeconv` package, which this one perfectly subsumes and extends; we should just replace that original function rather than introducing another package altogether that does the same thing.

pkg/util/timeconv/duration/duration.go
```diff
@@ -0,0 +1,161 @@
...
+type DurationUnit string
+
+const (
+	UnitMonth         = DurationUnit("mo")
+	UnitMillisecond   = DurationUnit("ms")
+	UnitMicrosecond   = DurationUnit("us")
+	UnitMicrosecondB5 = DurationUnit("µs")
+	UnitMicrosecondBC = DurationUnit("μs")
+	UnitNanosecond    = DurationUnit("ns")
+	UnitYear          = DurationUnit('y')
```
jranson, line 17: for clarity, make sure to use double quotes (string) rather than single (char) on all of these Unit consts.

pkg/util/timeconv/duration/duration.go
```diff
@@ -0,0 +1,161 @@
+package duration
+
+import (
+	"strconv"
+	"time"
+)
+
+type DurationUnit string
```
jranson, line 8: for package-level elements, order them by   imports, consts, vars, types, funcs.  So move this defnition down below the consts (even if it's referenced in consts)

pkg/util/timeconv/duration/duration.go
```diff
@@ -0,0 +1,161 @@
...
+func isUnit(s string, u DurationUnit) bool {
+	return s == string(u)
+}
+
+// Determine if rune is a digit, allowing for a sign if true, based on ASCII values
+func isDigit(s rune, allowSign bool) bool {
+	return (allowSign && s == 45) || (s >= 48 && s <= 57)
+}
+
+const (
```
jranson, line 60: move up to be with the other const definitions

pkg/util/timeconv/duration/errors.go
```diff
@@ -0,0 +1,35 @@
+package duration
+
+import "fmt"
+
+type ErrInvalidDurationFormat struct {
```
jranson, line 5: Same feedback on Err struct vs Obj naming

jranson (changes requested):

pkg/backends/influxdb/flux/errors.go
```diff
@@ -0,0 +1,51 @@
...
+		at: absoluteTime,
+		ut: unixTimestamp,
+	}
+}
+
+func (err *InvalidTimeFormatError) Error() string {
+	return fmt.Sprintf("invalid time format; must be relative duration (%s), RFC3999 string (%s), or Unix timestamp (%s)", err.rd, err.at, err.ut)
+}
+
+type FluxSyntaxError struct {
```
jranson, line 23: move up with the other type definition

jranson (approved):

PR #639, opened 2023-02-11
Cache Chunking

jakenichols2719:
WIP pull request for cache chunking functionality. Currently:

- Setting `use_cache_chunking` in cache config allows timeseries data to be split into even-duration chunks to be cached, reducing the overhead of getting small amounts of data from remote caches
- Setting `timeseries_chunk_factor` in cache config will determine the `Extent` that a single chunk encompasses. By default, this is `420 * step`.

This functionality is supported by all caches except memory.

Work remaining:

- [ ] Non-timeseries chunking; determine chunk size and key without alignment (time.Truncate)
- [ ] Documentation update; cache documentation needs a description of the new feature
- [ ] Testing; figure out a way to run current tests on a cache with chunking enabled


jakenichols2719:
Forgot to sign off on a commit. Fixing.

PR #640, opened 2023-02-11
Cache Chunking

jakenichols2719:
Introduced _chunking_ to cache functionality, splitting data into smaller chunks to allow for more efficient access, especially in remote caches like Redis.

- [x] Timeseries data + configuration
- [x] Byte data + configuration
- [ ] Testing
- [x] Documentation


jakenichols2719:
These changes incurred too much of a performance cost to be feasible in Trickster. I've refactored them to run concurrently and use much less memory, which has brought them much more in line with baseline performance. New PR incoming.

PR #654, opened 2023-03-12
Autodiscovery

jakenichols2719:
Functionality for the automatic discovery and creation of backend sources through queries on Clients. Currently supported are etcd clusters and Kubernetes clusters, specifics on how to use can be found in the documentation for ALBs at `/docs/alb.md`.


jakenichols2719:
Autodiscovery has been implemented for Kubernetes and ETCD queries.

This PR introduces a *lot* of lines of code--the k8s modules in particular have a significant number of dependencies, including multiple versions of the same dependencies. The main changes to `trickster` include:

- Introduction of package `autodiscovery` which defines the process for finding and instantiating backends
- Modification of package `alb` to use autodiscovery as configured

pkg/backends/templates/templates.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"fmt"
+	"reflect"
+	"strings"
+
+	"github.com/trickstercache/trickster/v2/pkg/backends/options"
+	ustrings "github.com/trickstercache/trickster/v2/pkg/util/strings"
+)
+
+var templateBackends map[string]*options.Options = make(map[string]*options.Options)
```
jranson, line 12: how does one reset the templateBackends back to an empty map (e.g., when reloading the application config when it has changed)?

pkg/routing/routing.go
```diff
@@ -137,6 +141,22 @@ func RegisterProxyRoutes(conf *config.Config, r router.Router, metricsRouter *ht
 	return clients, nil
 }
 
+// Register a every backend marked as a template under package backend/templates.
```
jranson, line 144: grammar check here.  also, comment should start with the function name

jranson (changes requested):

jakenichols2719:
So, my GPG certificate expired since I started working on this, and also, in retrospect I could have done a much better job on the configuration layout. I fixed those things in a new branch, so I'll be opening a new draft PR soon. Closing this one.

Issue #667, opened 2023-05-16
trickster alb panic


togear:
trickster panic when I test its alb roudrobin function
I want trickster work just like nginx proxy
-------------------------------------------
2023/05/16 10:19:16 http: panic serving 127.0.0.1:55564: unaligned 64-bit atomic operation
goroutine 84 [running]:
net/http.(*conn).serve.func1()
	/usr/local/go/src/net/http/server.go:1854 +0xe0
panic({0x8911800, 0x8b0bfc8})
	/usr/local/go/src/runtime/panic.go:890 +0x220
runtime/internal/atomic.panicUnaligned()
	/usr/local/go/src/runtime/internal/atomic/unaligned.go:8 +0x2d
runtime/internal/atomic.Xadd64(0xa006254, 0x1)
	/usr/local/go/src/runtime/internal/atomic/atomic_386.s:125 +0x11
github.com/trickstercache/trickster/v2/pkg/backends/alb/pool.nextRoundRobin(0xa006230)
	/home/huang/go/src/github.com/trickstercache/trickster/pkg/backends/alb/pool/round_robin.go:31 +0x77
github.com/trickstercache/trickster/v2/pkg/backends/alb/pool.(*pool).Next(0xa006230)
	/home/huang/go/src/github.com/trickstercache/trickster/pkg/backends/alb/pool/pool.go:86 +0x23
github.com/trickstercache/trickster/v2/pkg/backends/alb.(*Client).handleRoundRobin(0x9f20fc0, {0xef74e1d0, 0x9c7f6b0}, 0x9f3a580)
	/home/huang/go/src/github.com/trickstercache/trickster/pkg/backends/alb/round_robin.go:30 +0x2f
net/http.HandlerFunc.ServeHTTP(0x9c127a0, {0xef74e1d0, 0x9c7f6b0}, 0x9f3a580)
	/usr/local/go/src/net/http/server.go:2122 +0x34
github.com/trickstercache/trickster/v2/pkg/encoding/handler.HandleCompression.func1({0x8b0f910, 0x9c28ea0}, 0x9f3a500)
	/home/huang/go/src/github.com/trickstercache/trickster/pkg/encoding/handler/handler.go:55 +0x2b7
net/http.HandlerFunc.ServeHTTP(0x9c10a50, {0x8b0f910, 0x9c28ea0}, 0x9f3a500)
	/usr/local/go/src/net/http/server.go:2122 +0x34
[...]

PR #668, opened 2023-05-16
Replace atomic call in roundrobin

jakenichols2719:
The roundrobin package was using atomic.AddUint64, which appears to be creating an issue for this user: #667. Per documentation linked below, this function is unsafe on some hardware, and the call/error matches up with what would be expected in that case. The pool now uses an `atomic.Uint64` to count its position, and roundrobin uses the suggested `Uint64.Add` function instead.

There are other parts of the code that also use this, but I don't want to go around changing stuff that we haven't heard is broken, so I'm leaving those instances out of this for now and noting it as a potential area of concern if we see similar reports.

### DIFF
--- a/pkg/backends/alb/pool/pool.go
+++ b/pkg/backends/alb/pool/pool.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"net/http"
 	"sync"
+	"sync/atomic"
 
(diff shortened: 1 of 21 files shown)


jakenichols2719:
Fixing the DCO issue and then writing an alias so I don't forget to sign off again. Back in a few.

Issue #671, opened 2023-05-22
Custom paths not effective


morningstart-liu:
Hi，
   I tested paths functionality by referring to the profile of paths.md. But  didn't get the result that I wanted, and then I got the runtime configuration (http://127.0.0.1:8484/trickster/config) and I didn't see the configuration that I had customized, so, can you see where the problem is, thanks you .

   version: Trickster version: 2.0.0-beta2 (linux/amd64)   

  my config.yaml
~~~
request_rewriters:
  example_rewriter:
    instructions:
      - [ 'header', 'set', 'Req', 'www.test.com1111' ]
      - [ 'hostname', 'set', 'www.test.com' ]
      
frontend:
  listen_port: 8480
#  listen_address: ''
#  tls_listen_address: ''
#  tls_listen_port: 0
#  0 by default, unlimited.
#  connections_limit: 0

caches:
  default:
    provider: filesystem
    index:
#     reap_interval_ms defines how long the Cache Index reaper sleeps between reap cycles. Default is 3 (3s)
      reap_interval_ms: 3000
#     flush_interval_ms sets how often the Cache Index saves its metadata to the cache from application memory. Default is 5 (5s)
      flush_interval_ms: 5000
#     max_size_bytes indicates how large the cache can grow in bytes before the Index evicts least-recently-accessed items. default is 512MB
      max_size_bytes: 536870912
#     max_size_backoff_bytes indicates how far below max_size_bytes the cache size must be to complete a byte-size-based eviction exercise. default is 16MB
      max_size_backoff_bytes: 16777216
#     max_size_objects indicates how large the cache can grow in objects before the Index evicts least-recently-accessed items. default is 0 (infinite)
      max_size_objects: 0
#     max_size_backoff_objects indicates how far under max_size_objects the cache size must be to complete object-size-based eviction exercise. default is 100
      max_size_backoff_objects: 100
    filesystem:
#     cache_path defines the directory location under which the Trickster cache will be maintained
#     default is /tmp/trickster
      cache_path: /data/trickster
      
backends:
  default:
    req_rewriter_name: example_rewriter
    origin_url: http://www.test.com:8089
    provider: reverseproxy # use 'reverseproxy' for no caching  reverseproxycache
    path_routing_disabled: true
    cache_name: default
    is_default: true
    hosts: [ www.test.com ]
    max_object_size_bytes: 524288
    max_ttl_ms: 300000
    cache_key_prefix: test-prefix
    tracing_name: default
[...]

PR #672, opened 2023-05-24
Safely copy paths to new *bo.Options

### LINKED ISSUE #655: Custom paths are not appended to config
Hi, it seems that custom paths are not considered while parsing the configuration.

I tried with both 1.x and 2.x. I debugged 2.x and it seems that, while parsing the configuration, setDefaults runs on `o.Paths` but o.Paths is not assigned to `no.Paths` ( https://github.com/trickstercache/trickster/blob/main/pkg/backends/options/options.go#L630 )

jnichols-git:
The `SetDefaults` function now copies over custom paths to the returned options by iterating the "old" routes (after they've had their own SetDefaults called) and setting the value for each key to a clone of the "old" value.

This was brought to my attention in issue #671, and my testing indicates that it resolves that issue; however, due to issues with the posted config (happens with copying over sometimes), my test wasn't an exact replica, so I will wait to hear from the issue creator. #655 was mentioned as a possible cause, and resolving that was what lead to this solution.

Closes #655.


morningstart-liu:
Something is missing，in paths's SetDefaults()   have a line :    p.Custom = make([]string, 0) .   take memory for Custom. your modify will lost Custom.

jakenichols2719:
I'll have to get back to you on that specific chunk, I need to check with the original author to see what's meant to happen there.

jakenichols2719:
> Something is missing，in paths's SetDefaults() have a line : p.Custom = make([]string, 0) . take memory for Custom. your modify will lost Custom.

I meant to get back to you on this earlier, sorry. Slice size is automatically updated when you append to them, so the issue isn't the allocation here. We've discovered that the issue is (probably) that configs are reaching Merge before Custom is assigned to by SetDefault. That's why you're seeing empty Custom fields when Merge happens sometimes. Working on a fix!

PR #674, opened 2023-05-25
fix parsing backends.paths

paragor:
fix parsing backends.paths


jakenichols2719:
This is resolved in #672. Additionally, please make sure to copy over the elements of slices/maps, and not by reference; this can cause some undesired behavior when modifying either copy. Marking as duplicate for now--will leave open for a moment in case you have other thoughts on the issue.

PR #675, opened 2023-05-31
Autodiscovery

jnichols-git:
Unreasonably Large Pull Request 2

- Config makes more sense; reduced unneeded options
- Backends are now properly instantiated and added to the ALB pool

Drafting as there's currently a bug preventing access to the instantiated backends. Working on a fix and will un-draft once ready.


jranson:
This is replaced by #1116

Issue #594, opened 2022-01-10
Trickster v2.0.0-beta2 : Prometheus query_range calls resulting in proxy-only sometimes instead of cache hit


sumitk23:
We are trying to upgrade from Trickster v1.1 to v2.0.0-beta2 and facing the following issue. Hit rate has decreased and some of the calls are now being served as `proxy-only`. More details in logs below.

The same suite which used to give 100% hit rate in v1.1 is giving lesser hit rate in v.2.0 beta version.

**Test setup** :
Trickster is deployed in our K8s cluster and we tried running the test suite with different number of pod replicas and here are our findings. We are making 10 queries/sec. Test is run continuously for ~10 mins. Same query (test_metrics) which is already present in TSDB is made every time.
Backend used is Prometheus and config is such that recent 5 mins data is not cached, anything older than till 3 weeks is cached.

<img width="529" alt="Screenshot 2022-01-10 at 11 50 15 AM" src="https://user-images.githubusercontent.com/15623801/148725205-e11b86ca-cb0d-4f40-ac3b-ee30d9db5bfa.png">

Fixed time range means test_metrics is queried for same fixed time range every time REST call is made ie. start & end query params is fixed for all calls.
Variable time range uses a dynamic time range ie. end as the currentTime and start as currentTime - 900 secs (15 mins ago) for each API call this should result in partial hit (phit) for Trickster.

Fixed time range means test_metrics is queried for same fixed time range every time REST call is made ie. start & end query params is fixed for all calls.

Variable time range uses a dynamic time range ie. end as the currentTime and start as currentTime - 900 secs (15 mins ago) for each API call this should result in partial hit (phit) for Trickster.
<img width="1980" alt="Trickster-hit-rates-diff-replica-counts" src="https://user-images.githubusercontent.com/15623801/148725503-15d0d59c-aaca-47b3-ba34-02d868d8f5ec.png">

Panel Queries : 
Hit Rate : `sum(rate(trickster_proxy_requests_total{path=~"/api/.*", cache_status=~"hit|phit|nchit|rhit", job="m3-query-pod-name"}[2m])) /  sum(rate(trickster_proxy_requests_total{path=~"/api/.*", job="m3-query-pod-name"}[2m]))`

[...]

PR #687, opened 2023-10-10
Sort the paths in routes to ensure matcher picks nearest regex to request uri

PKhivasra:
When the default backend routes are registered, currently it is as per unordered map. This means that the routes can have "" or "\api\v1" before "api\v1\query_range". When the router matches the regex expression (Refs: https://gecgithub01.walmart.com/Telemetry/trickster-v2/blob/main/pkg/router/route.go#L40 https://gecgithub01.walmart.com/Telemetry/trickster-v2/blob/main/pkg/router/regexp.go#L323) with the routes map, if partial routes appears first that will be considered as a match, which causes request to be "Proxy-Only".
Thus, with sorting we will ensure that routes will be a sorted array in the descending order of path lengths, meaning: "\api\v1\query_range" would appear before "api\v1" and "" so that we have appropriate match.

Issue:https://github.com/trickstercache/trickster/issues/594

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ Trickster is hosted by the [Cloud Native Computing Foundation](https://cncf.io) 
 
 Note: Trickster v1.1 is the production release, sourced from the [v1.1.x](https://github.com/trickstercache/trickster/tree/v1.1.x) branch. The `main` branch sources Trickster 2.0, which is currently in beta.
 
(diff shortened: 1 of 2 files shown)


jranson (approved):

jnichols-git (approved):
Tested and working, thank you!

PKhivasra:
> Tested and working, thank you!

Thanks!

PR #688, opened 2023-10-19
chore: slice loop replace

testwill:


jakenichols2719:
@testwill Chore PRs are welcome, and this is a good subject for one, but this isn't the only instance of this loop pattern. While we're here and looking at it, could you please also fix this pattern where it occurs elsewhere?

jranson (approved):

jnichols-git (approved):

PR #697, opened 2024-01-10
Clickhouse multi-line query fix

### LINKED ISSUE #696: clickhouse multiline query parsing problem
https://github.com/trickstercache/trickster/blob/f31766c669b4dded2cafaccb6acc69a5afa67fb5/pkg/backends/clickhouse/handler_query.go#L55

In the current implementation of the ClickHouse backend, there appears to be an issue when sending multiline SELECT queries. The code doesn't pass these queries to the delta cache; instead, it simply proxies the request.
```sql
SELECT
  toStartOfHour(...
```

jnichols-git:
#696 was describing an issue with multi-line queries in ClickHouse, where if a query started with `SELECT` immediately followed by a newline, it would not get cached. The parser previously only looked for `SELECT ` or ` SELECT ` (including spaces), so this adds `SELECT\n` to the list. Also removed a `make` call that was unused.

### DIFF
--- a/pkg/backends/clickhouse/handler_query.go
+++ b/pkg/backends/clickhouse/handler_query.go
@@ -41,7 +41,6 @@ func (c *Client) QueryHandler(w http.ResponseWriter, r *http.Request) {
 		}
 		var body []byte
 		if sqlQuery != "" {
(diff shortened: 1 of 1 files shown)


jnichols-git:
Whoops. Forgot to set DCO. Please hold.

PR #698, opened 2024-01-10
Clickhouse fix

jnichols-git:
Issue 696 was describing an issue with multi-line queries in ClickHouse, where if a query started with SELECT immediately followed by a newline, it would not get cached. The parser previously only looked for SELECT or SELECT (including spaces), so this adds SELECT\n to the list. Also removed a make call that was unused.


jnichols-git:
Runners are really struggling today... Ready for review.

jranson (approved):

Issue #691, opened 2023-12-04
Image building schedule?


Starttoaster:
We've been running Trickster for years now, and we're starting to get to a point of being more mature about monitoring the lifecycles of applications running in containerized environments (namely Kubernetes.) We've noticed that even the `:latest` tag of `tricksterproxy/trickster` hasn't been pushed to with updates in around 3 years, and is currently built on top of a 3 year old release of Alpine linux. This triggers vulnerability scanners looking for system packages inside containers to notice that there are incredibly outdated packages with critical-level CVEs that have been patched in more recent versions of those packages.

The request for this Issue is that images (at least like, `:latest` and maybe a short list of supported versions) be rebuilt on a schedule, so that updates to the alpine image this gets based on get pulled into the tricksterproxy/trickster image.

If that is something you'd be interested in and don't currently have the maintenance bandwidth, I could assist with making a scheduled workflow for this even.

Starttoaster:
@trickstercache/maintainers been a minute, figured I'd add a ping here in case the Maintainers aren't watching :)

Starttoaster:
Communication is a bit tight around here... It would be nice to get a bit of insight into the internal thought processes of the maintainers here about the idea of rebuilding these images every once in a while, rather than leave me hanging on an eyes emoji after begging for a bit of contact after a month... I'm pretty patient, I just generally expect like a "we'll discuss it in the next maintainers meeting, which happens next year around this time." Or "yeah, that sounds good, would you like to make the workflow for us, as it would be likely to be done quicker?" Rather than utter silence. The current level of communication could lead people who are external to the maintainers team to imagine that this utility has maybe become abandoned.

jnichols-git:
@Starttoaster You sound like you've heard the personal/common answers before, so I'll spare you those--I'm just not in a position to sign off on changes (especially build/actions related) myself, and I'm waiting to hear back from another maintainer regarding making an effective change to the build process. Our schedules don't regularly align and are usually full, so sometimes this takes a second. I'm hearing that this is a pain point for your team, so I'll push to get it resolved sooner rather than later. I'll respond here by next Friday with an update.

Starttoaster:
[...]

PR #712, opened 2024-04-09
Add docker image publishing workflows

Starttoaster:
Sorry for the radio silence on this, I wanted to get to this sooner but I've been extremely busy as of late unfortunately. Eventually carved out time to work on this :smile: Regards https://github.com/trickstercache/trickster/issues/691 FYI @jnichols-git 

These workflows should handle building and pushing images to docker hub and github's container registry, using qemu and docker buildx to build for the arm64 and amd64 platforms. 

We have two files here:

`publish-docker.yaml`:
* This will run when a release is created (which excludes prereleases, like release candidates)
* On a release workflow, it creates some semver docker tags (like `:x.x.x`, `:x.x`, and `:x`), it also updates the `:latest` tag from the latest release code, and creates a `:sha-<commit sha long>` tag
* This will run on merges/pushes to main
* On a main workflow, it refreshes a `:main` tag, and creates a `:sha-<commit sha long>` tag
* It also runs the main workflow on a schedule

`publish-docker-prerelease.yaml`:
* This will run when a prerelease is created (which is a checkbox in the github UI when creating a release, and is typically used for betas or release candidates)
* On a prerelease workflow, it creates a semver docker tags (like `:x.x.x-rc1` for a tagged prerelease like `:x.x.x-rc1`), and creates a `:sha-<commit sha long>` tag

A slight change here that I wanted to specifically call out was that this adds a new kind of tag that I don't think you currently do. And _this_ is the tag that will get rebuilt over time from a fresh alpine base. Which is the `:main` tag. I can also just switch this to be the `:latest` tag, depending on what you want. But `:main` gets refreshed every week. `:latest` will just track the latest released version and will be refreshed every time a release happens. And the semver tags will just be tagged once when a release is cut.

Some important general notes here:
* This begins uploading trickster images to github's container registry, which isn't currently done here. I can strip that out here if you want. I just figure it's valuable to have a second registry and basically free since the workflow should manage tagging/uploading there.
* This relies on a Docker Hub username and access token with read+write privileges in github secrets, which a maintainer here will need to provide.

### DIFF
--- a/.github/workflows/publish-docker.yaml
+++ b/.github/workflows/publish-docker.yaml
@@ -48,7 +48,7 @@ jobs:
         with:
           images: |
             ghcr.io/${{ github.repository }}
-            tricksterproxy/trickster
+            trickstercache/trickster
           flavor: |
             latest=auto
             prefix=
--- a/.github/workflows/publish-docker.yaml
+++ b/.github/workflows/publish-docker.yaml
@@ -1,0 +1,69 @@
+name: Publish docker image on main and releases
+
+on:
+  push:
+    branches:
+      - main
+  release:
+    types: [released]
+  workflow_dispatch:
+  schedule:
+    # Monday at 9am UTC
+    - cron: '0 9 * * 1'
+
+permissions:
+  id-token: write
+  contents: read
+  packages: write
+
+jobs:
+  package:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ github.token }}
+
+      - name: Login to DockerHub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PAT }}
+
+      - name: Docker meta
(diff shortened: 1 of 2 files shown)


Starttoaster:
Ah, my commits are _signed_ but not _signed off on_. So I'm failing that DCO check. Will fix that but will have to force push it up.

Starttoaster:
FYI, I set up mostly the same workflows in a repo of my own just to try them out, and you can see what they did. The only thing I think I stripped out was the docker hub login, for testing purposes I'm just using github's container registry. But here: https://github.com/Starttoaster/ci-playground
I created a release tag, a pre-release tag, and just ran CI against a push to main. Those created the following tagged version https://github.com/Starttoaster/ci-playground/pkgs/container/ci-playground

.github/workflows/publish-docker.yaml
```diff
@@ -0,0 +1,69 @@
...
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_PAT }}
+
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: |
+            ghcr.io/${{ github.repository }}
+            tricksterproxy/trickster
```
Starttoaster, line 51: I notice in Docker Hub, there's a trickstercache user, a tricksterproxy user, and a tricksterio user, that all claim to own the trickster image. Are these all you? Did I even choose the right one in my pull request? :smile:

Starttoaster: I just switched this to trickstercache because I saw that's the user called out in the readme here.

Starttoaster: But is it actually all 3 of them? And if so, is there a reason besides something like a typo-squatting defense?

Starttoaster:
Closing. I think there's a lack of interest here. I'm just going to own my fork of this repo with security patches :)

PR #731, opened 2024-08-13
SQL Parser Performance

jranson:
This patch improves the SQL Lexer/Parser performance by refactoring from the previous scheme of concurrent lexing/parsing via channels to a lex-then-parse scheme that is more performant due to a 100% removal of channels in the request path.


LimitlessEarth (approved):
Love it. This is my favorite kind of PR.

PR #737, opened 2025-03-06
remove unneeded nil checks

jranson:
this removes duplicative nil checks where the length of the slice or map is also being checked

because unit tests on main currently fail from recent package updates, this PR is expected to fail unit tests too. will address failing tests in a separate PR.

### DIFF
--- a/pkg/proxy/headers/headers.go
+++ b/pkg/proxy/headers/headers.go
@@ -162,7 +162,7 @@ func (l Lookup) Clone() Lookup {
 // If a key exists in both maps, the source value wins.
 // If the destination map is nil, the source map will not be merged
 func Merge(dst, src http.Header) {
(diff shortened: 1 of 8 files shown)


pkg/proxy/headers/headers.go
```diff
@@ -162,7 +162,7 @@ func (l Lookup) Clone() Lookup {
 // If a key exists in both maps, the source value wins.
 // If the destination map is nil, the source map will not be merged
 func Merge(dst, src http.Header) {
-	if src == nil || len(src) == 0 || dst == nil {
+	if len(src) == 0 || dst == nil {
```
crandles, line 165: Would this be more consistent:

```suggestion
	if len(src) == 0 || len(dst) {
```
Or have I missed something in the diff

crandles (approved):

PR #739, opened 2025-03-09
fix: fix various race conditions

crandles:
This fixes a subset of observed data races.

The following packages still have races, but appear more complex to fix.
- pkg/proxy/engines
- pkg/proxy/listener
- pkg/timeseries/dataset


crandles:
I'm going to document the remaining races, they don't appear simple to fix.

PR #741, opened 2025-03-13
ci: use reviewdog with golangci-lint to limit feedback to changed files

crandles:
Upgrade the CI lint to use reviewdog, which should limit feedback to just what is changed in the PR.


crandles:
This doesn't seem to add much vs what is there already; though reviewdog was more useful

PR #753, opened 2025-03-30
feat: [ci] add docker image release on "on tag" workflow & minor fixes

crandles:
Did a bit of testing on my private fork, this _should_ have a better chance of succeeding.

* Publishes a docker image on tag
* Fixes bug found in initial run of `publish-image` workflow (https://github.com/trickstercache/trickster/actions/runs/14149543077)
  * Refactors docker publishing to use the docker build-push action + use GHA cache
*  Refactors sha256sum generation to fix observed errors (`/bin/sh: 2: [[: not found`)
* Removes force rebuilding of standard library packages in release workflow (`-a`)
* Only builds for one architecture during PR workflows; we aren't doing anything platform specific
* Replaces to-be-deprecated `::set-output` usages
* Allow PR workflow to succeed in forked repositories (skip code coverage publishing)
* Cross compile trickster from local build platform -- speeds up docker build times when going cross-arch
  * Reduces docker build times from 10 minutes to ~3 

Resolves: https://github.com/trickstercache/trickster-docker-images/issues/3

### DIFF
--- a/.github/workflows/ci-build-tests.yml
+++ b/.github/workflows/ci-build-tests.yml
@@ -34,7 +34,8 @@ jobs:
     steps:
       - name: Get Build Date
         id: builddate
-        run: echo "::set-output name=builddate::$(date -u +%Y-%m-%dT%H:%M:%SZ)"
+        run: |
+          echo "builddate=$(date -u +%Y-%m-%dT%H:%M:%SZ)" | tee $GITHUB_OUTPUT
       - uses: actions/checkout@v4
       - name: Set up QEMU
         uses: docker/setup-qemu-action@v3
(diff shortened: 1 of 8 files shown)


crandles:
Looks like `coveralls` is undergoing maintenance.

> In progress - Scheduled maintenance is currently in progress. We will provide updates as necessary.

https://status.coveralls.io/

I am going to allow skipping coverage publishing error. I will create a follow up ticket for logging this as a PR comment. There's no code change here anyways.

crandles:
I can't tell what has introduced the panic; does not appear related to any code change.

Tried the last few commits to main, and they fail now, although they didn't at PR time.

Best guess is we have something timing related going on (that's deciding to act differently today).

jranson (approved):

PR #759, opened 2025-03-31
fix: [release] combine image publishing platforms into same job run

crandles:
## Bug

The `publish image on main` workflow (https://github.com/trickstercache/trickster/actions/runs/14179942124) failed to publish a valid multi-platform build to Docker hub -- when pulling the image, only one platform is supported.

In order to support two separate `docker push --platform ...` calls of "the same image", we would need to perform some extra "docker manifest" updates in order to create a tag that points to both separately published platform tag. To avoid this complexity, we should drop back to a single execution of the docker command, for both platforms.

## Resolution

Combine image publishing of different platforms into the same build process.

I split these apart while trying to optimize the workflow duration, but failed to check the resulting image. Ultimately splitting the jobs apart didn't speed up the workflow, other optimizations did (using cross compilation), so this is fine to revert .

Planning to test this in my fork first, then I'll mark ready for review.


jranson (approved):

crandles:
`TestPCFWaits` appears sensitive:

```
=== RUN   TestPCFWaits
    progressive_collapse_forwarder_test.go:178: WaitAllComplete returned too quickly, expected wait got finished
--- FAIL: TestPCFWaits (1.26s)
```

Issue #765, opened 2025-04-03
Data Race observed in pkg/proxy/engines


crandles:
Observed via CI, have yet to reproduce locally, assume it is intermittent.

- https://github.com/trickstercache/trickster/actions/runs/14234721400/job/39891872593

```
2025-04-03T04:10:24.0648843Z === RUN   TestDeltaProxyCacheRequestRangeMissChunks
2025-04-03T04:10:24.0649117Z ==================
2025-04-03T04:10:24.0649310Z WARNING: DATA RACE
2025-04-03T04:10:24.0649527Z Read at 0x00c0005e0010 by goroutine 1361:
2025-04-03T04:10:24.0649941Z   github.com/trickstercache/trickster/v2/pkg/proxy/engines.WriteCache()
2025-04-03T04:10:24.0650906Z       /home/runner/work/trickster/trickster/pkg/proxy/engines/cache.go:435 +0x684
2025-04-03T04:10:24.0652167Z   github.com/trickstercache/trickster/v2/pkg/proxy/engines.DeltaProxyCacheRequest.func2()
2025-04-03T04:10:24.0653447Z       /home/runner/work/trickster/trickster/pkg/proxy/engines/deltaproxycache.go:422 +0x5aa
2025-04-03T04:10:24.0654324Z 
2025-04-03T04:10:24.0654576Z Previous write at 0x00c0005e0010 by goroutine 1313:
2025-04-03T04:10:24.0655588Z   github.com/trickstercache/trickster/v2/pkg/timeseries.(*TimeRangeQuery).NormalizeExtent()
2025-04-03T04:10:24.0656356Z       /home/runner/work/trickster/trickster/pkg/timeseries/timerangequery.go:95 +0x1ad
2025-04-03T04:10:24.0656980Z   github.com/trickstercache/trickster/v2/pkg/proxy/engines.DeltaProxyCacheRequest()
2025-04-03T04:10:24.0658052Z       /home/runner/work/trickster/trickster/pkg/proxy/engines/deltaproxycache.go:95 +0x937
2025-04-03T04:10:24.0659364Z   github.com/trickstercache/trickster/v2/pkg/proxy/engines.(*TestClient).QueryRangeHandler()
2025-04-03T04:10:24.0660796Z       /home/runner/work/trickster/trickster/pkg/proxy/engines/client_test.go:378 +0x1f1
[...]

PR #766, opened 2025-04-03
fix: [pkg/proxy/engines] fix data race in time range query normalization

crandles:
Attempted fix for #765.

Having trouble reproducing the race, so still WIP.


crandles:
This package has changed a bit, going to close those out.

I have seen the odd panic or two, but doubt this pr is still relevant.

PR #767, opened 2025-04-04
refactor appends

jranson:
This change refactors a number of instances where slices are constructed and managed using `append()` and replaces those with pre-allocated slice sizes, iterating, a final crop. This tranche targets some of the biggest opportunities for performance gains based on how frequently they are called and the average size of the slices affected by refactoring.

This also addresses a logic defect in the `pkg/timeseries/dataset/series_list.go | SeriesList.merge()` that would allow a duplicated series to make it into the merged list in place of a a unique series (e.g., merging `[ 'r', 'o', 'y', 'o']` with `['r', 'b', 'p', 'g', 'y']` would result in `['o', 'o', 'b', 'p', 'g', 'y']` instead of `['r', 'o', 'b', 'p', 'g', 'y']`). Should not be a thing in the real world as no upstream DB would send the same series twice, but you never know.

Lastly, in `pkg/timeseries/dataset/series_header.go`, a cached version of the header hash is now saved in the Header upon the first call to `CalculateHash()`. When `CalculateHash()` is called a subsequent time, the cached hash is returned instead of recalculating every time.


pkg/proxy/handlers/health/health.go
```diff
@@ -107,24 +107,31 @@ func udpateStatusText(hc healthcheck.HealthChecker, hd *healthDetail) {
...
+	}
 
-		sort.Strings(u)
-		sort.Strings(q)
+	a = a[:al]
+	u = u[:ul]
+	q = q[:ql]
 
-	}
+	sort.Strings(u)
```
crandles, line 133: it looks like this is sorted on L138. Does it need to be sorted here?

pkg/proxy/handlers/health/health.go
```diff
@@ -107,24 +107,31 @@ func udpateStatusText(hc healthcheck.HealthChecker, hd *healthDetail) {
...
+	sort.Strings(q)
```
crandles, line 134: it looks like this should be sorted on L171 (its sorting a, but seems like it should sort q)

pkg/util/bytes/bytes.go
```diff
@@ -0,0 +1,19 @@
...
+	lenA := len(a)
+	lenB := len(b)
+	if lenA > maxInt {
+		return a[:maxInt]
+	}
+	remaining := maxInt - lenA
+	if lenB > remaining {
+		b = b[:remaining]
+	}
+	out := make([]byte, lenA+len(b))
```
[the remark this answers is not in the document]
jranson: Marking as False Positive, since this function safely ensures the byte slice can't be larger than maxInt (based on GOARCH)

```
const maxInt = int(^uint(0) >> 1)

func MergeSlices(a, b []byte) []byte {
	lenA := len(a)
	lenB := len(b)
	if lenA > maxInt {
		return a[:maxInt]
	}
	remaining := maxInt - lenA
	if lenB > remaining {
		b = b[:remaining]
	}
	out := make([]byte, lenA+len(b))
	copy(out, a)
	copy(out[lenA:], b)
	return out
}
```

crandles (approved):

PR #768, opened 2025-04-06
BugFix for DataSet and TimeRangeQuery

jranson:
* Adds `OriginalBody` to `timeseries.TimeRangeQuery`. In the event that an inbound time series request is a `POST` (which is allowed by most supported backends), TimeseriesBackend.Parse() must read in all of the request body's `io.Reader` in order to parse a Time Range Query, meaning it can't be read again. In the event that the inbound request is Proxy-Only for any reason (e.g., it's an UPDATE query and not a SELECT, etc.), this field ensures that the original request body is re-attached to the outbound proxy request.
* Adds a data seeding container for ClickHouse to the developer docker compose.
* Moves the Point Merging logic in `dataset.DefaultMerger` to a standalone `MergePoints` function and some other cleanup in `dataset` (accuracy and performance)
* Moves Series List Merging into a more formal `Merge()` function that also handles any necessary point merging by calling MergePoints as needed.
* Sorts the resulting SeriesList from a merge of two SeriesLists by Tags (ensures series order is exactly as it would be if Prometheus served it directly)
* Fixes linter nits about using IfElse too much or something and change to switches. 🤷🏽


pkg/timeseries/dataset/point.go
```diff
@@ -45,6 +47,61 @@ func (p *Point) Clone() Point {
...
 }
 
+// Equal returns true if p and p2 are exactly equal
+func PointsAreEqual(p1, p2 Point) bool {
+	if p1.Epoch != p2.Epoch || p1.Size != p2.Size ||
+		len(p1.Values) != len(p2.Values) {
+		return false
+	}
+	for i, v := range p1.Values {
+		switch t := v.(type) {
```
crandles, line 57: have you considered this generic func from the stdlib: https://pkg.go.dev/cmp#Compare

pkg/timeseries/dataset/point.go
```diff
@@ -132,3 +189,69 @@ func (p Points) onOrJustBefore(ts epoch.Epoch, s, e int) int {
...
+	// sort, keeping order between equal elements
+	slices.SortStableFunc(p, func(a, b Point) int {
+		if a.Epoch < b.Epoch {
+			return -1
+		} else if a.Epoch > b.Epoch {
+			return 1
+		}
+		return 0
+	})
+	var k int
```
crandles, line 207: have you considered: https://pkg.go.dev/slices#CompactFunc ?

pkg/timeseries/dataset/series_list.go
```diff
@@ -18,82 +18,88 @@ package dataset
...
+			// this series does not exist in sl1; add it into out
+			out[k] = s
+			m[h] = s
 			k++
-			updateLookup(out[:k])
-			continue
+		} else {
+			// series is in both sl1 and sl2; merge their points
+			wg.Add(1)
+			func(s1, s2 *Series) {
```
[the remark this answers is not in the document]
crandles: On second look, I think `seen` is providing this

pkg/timeseries/dataset/point.go
jranson: CompactFunc keeps only the first of consecutive values in the sorted slice, whereas this inline solution keeps only the last element. The idea here is that while the Epochs of two Points being compared could be the same, their underlying values may be different in newer data (which falls on the high/right side of consecutives). And we want to keep the newer data because the cached values could have been preliminary. We could re-arrange some things so the order of the points coming in are newest first, but that could introduce other unforeseen problems, so I'm not sure it's worth it to swap everything out to take advantage of CompactFunc.

crandles:
> Fixes linter nits about using IfElse too much or something and change to switches

Feel free to turn off anything too opinionated.

PR #770, opened 2025-04-06
feat: [config] allow referencing header values from environment variables

crandles:
Following #769, adds environment substitution support to the backend and health-check header values.

Adds new type, `type EnvStringMap map[string]string` to handle substitution into a map of values; this type is chosen so that it is more easily retrofit into the current `map[string]string` usage.


jranson:
in this PR, would you mind adding a bullet describing this and the redis PW to docs/new-changed-2.0.md? And take a peek at some of the other configuration docs and update.

jranson (approved):

PR #772, opened 2025-04-07
refactor: use more stdlib & optimize dataset/points.Size

crandles:
* Removes unused code
* Replaces custom string/slice functions with stdlib
* Removes extra `slices.reverse` call in `pkg/router/lm`
* Optimize point size calculation

---
Equal -- ~no change
Unique --  20% improvement

```
BenchmarkEqual-12     	245202459	         4.393 ns/op
BenchmarkUnique-12    	14429709	        88.49 ns/op
```

vs (main):

```
BenchmarkEqual-12     	263193535	         4.449 ns/op
BenchmarkUnique-12    	10249306	       112.2 ns/op
```
---
Points.Size -- looks like a big improvement -- I'm not sure if the goroutine was helping here

```
BenchmarkPointsSize-12    	     100	       654.5 ns/op
```

vs (main):

```
BenchmarkPointsSize-12    	     100	    291119 ns/op
```

### DIFF
--- a/pkg/util/strings/strings.go
+++ b/pkg/util/strings/strings.go
@@ -38,8 +38,7 @@ func Substring(s string, i int, length int) string {
 func Unique(in []string) []string {
 	out := slices.Clone(in)
 	slices.Sort(out)
-	out = slices.Compact(out)
-	return out
+	return slices.Compact(out)
 }
 
 // ErrKeyNotInMap represents an error for key not found in map
--- a/pkg/util/strings/strings.go
+++ b/pkg/util/strings/strings.go
@@ -20,6 +20,7 @@ package strings
 import (
 	"errors"
 	"fmt"
+	"slices"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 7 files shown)


pkg/util/strings/strings.go
```diff
@@ -33,62 +34,17 @@ func Substring(s string, i int, length int) string {
...
-	if s1 == nil || s2 == nil || len(s1) != len(s2) {
-		return false
-	}
-	for i, v := range s1 {
-		if v != s2[i] {
-			return false
-		}
-	}
-	return true
+	return slices.Equal(s1, s2)
```
jranson, line 39: we should just remove this func and its unit tests altogether, and change the one reference to it (in proxy/tls/options/options.go) over to slices.Equal

jranson (changes requested):
one minor change, but otherwise LGTM. Thanks!

pkg/util/strings/strings.go
```diff
@@ -33,62 +34,12 @@ func Substring(s string, i int, length int) string {
...
-			continue
-		}
-		out[k] = v
-		k++
-		m[v] = nil
-	}
-	return out[:k]
+	out := slices.Clone(in)
+	slices.Sort(out)
+	out = slices.Compact(out)
```
jranson, line 41: just `return slices.Compact(out)` instead of reassigning and then returning out.

jranson (approved):

PR #774, opened 2025-04-08
refactor: minor improvements

crandles:
This PR addresses a number of warnings found by upgrading golangci-lint to v2 and turning on more linters.

I'm still working through a few errors, but will save those and the golangci-lint updates for another PR.

---

Summary of changes:

- various logic simplifications
- usage of strings.ReplaceAll
- reduced assignments where possible
- use `fmt.Fprintf` vs `w.Write([]byte(fmt.Sprintf(...`
- use more (stdlib and custom) constants

### DIFF
--- a/pkg/daemon/setup/setup.go
+++ b/pkg/daemon/setup/setup.go
@@ -95,7 +95,7 @@ func ApplyConfig(si *instance.ServerInstance, newConf *config.Config,
 
 	applyLoggingConfig(newConf, si.Config)
 
-	//Register Tracing Configurations
+	// Register Tracing Configurations
 	tracers, err := tr.RegisterAll(newConf, false)
 	if err != nil {
 		handleStartupIssue("tracing registration failed",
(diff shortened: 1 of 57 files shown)


pkg/daemon/setup/setup.go
```diff
@@ -221,7 +220,8 @@ func applyCachingConfig(si *instance.ServerInstance,
 			// the cache should be preserved between reconfigurations, but only if the Index
 			// is the only change. In this case, we'll apply the new index configuration,
 			// then add the old cache with the new index config to the new cache map
-			if ocfg.ProviderID == v.ProviderID &&
+			// TODO: the comment and logic below don't match (this seems limited to the memory provider); revisit this later
+			if ocfg.ProviderID == v.ProviderID && //nolint:gocritic
```
crandles, line 224: any insight into this? linter calls this out as suspicious, and I think I agree:
```
badCond: `ocfg.ProviderID == v.ProviderID &&
	ocfg.ProviderID == providers.Memory` condition is suspicious (gocritic)
```

why would we check if old providerID == new provider id AND old provider ID == what i think is a _type_ of provider?

In addition, above comment doesn't seem to align with this being limited to memory

crandles: I see a few other caches include an `index.Index`; it would seem they all need this UpdateOptions call?

jranson: This works as intended, even though there may be a better way to do it, and the comment probably needs to be clarified.

If the execution makes it to this point, it means a cache config was found in the new object with the same name as a cache config in the old object, but with different settings. So this is checking whether it is OK to modify and re-use the existing cache, or close it out and spin up a new cache instance in its place.

The part flagged by the linter checks specifically if the old and new configs the same Provider, and if so, is that provider MemoryCache.

Yes -> The existing MemoryCache is reused while passing in the new options to be used by its Indexer. This prevents clearing the MemoryCache unnecessarily if only the size, reap interval or other property changes.

Otherwise -> (e.g., the providers are different between old and new, or the providers are the same but not a MemoryCache) The old cache is closed out and a new cache instance is registered.

You  might be able to fix the linter criticism by changing the second equality to check `v` instead of `ocfg` (e.g., `v.ProviderID == providers.Memory`)

crandles: Thanks, I may just add a note here about why its only the memory provider at the moment

jranson (approved):

PR #780, opened 2025-04-10
chore: add more linters and resolve warnings

crandles:
- adds additional linters and resolves warnings
- adds staticcheck to the CI process and fixes warnings

---

### notable changes:

- handle many unhandled errors (exceptions: .Write and .Close errors)
- remove unused test code

-- -

we can always turn these linters off later on if we feel they are too restrictive, but I find them helpful to guide refactoring.

Please look over the error handling I added. I'm not sure if we were skipping certain errors on purpose (outside of the exceptions I added)


pkg/backends/backend.go
```diff
@@ -80,9 +80,9 @@ type Registrar func(map[string]http.Handler)
 
 // New returns a new Backend
 func New(name string, o *bo.Options, registrar Registrar,
-	router http.Handler, cache cache.Cache) (Backend, error) {
+	router http.Handler, c cache.Cache) (Backend, error) {
```
crandles, line 83: there's a rule that warns against shadowing package imports. I'm not entirely sold on its importance. When it becomes a problem, its _usually_ obvious (usually based on code editor autocomplete, in my experience).

It would reduce this PR by a little, and lessen up a possibly needless restriction.

Thoughts?

crandles: yeah this seems like a too-late opinion to apply, going to revert

crandles:
This seems like it could conflict with #768 a bit (which seems more important); I'm ok to wait on this PR if you'd like to get that merged first

pkg/cache/bbolt/bbolt.go
```diff
@@ -218,7 +218,7 @@ func (c *Cache) SetTTL(cacheKey string, ttl time.Duration) {
 
 // Remove removes an object in cache, if present
 func (c *Cache) Remove(cacheKey string) {
-	c.remove(cacheKey, false)
+	c.remove(cacheKey, false) //revive:disable:unhandled-error
```
jranson, line 221: for these, we can probably capture and use `logger.Error()` on the error to resolve linter warnings

jranson (approved):

pkg/cache/bbolt/bbolt.go
crandles: will update

crandles:
fyi improvement on points equal:
```
BenchmarkPointEqual-12    	1000000000	         1.223 ns/op	       0 B/op	       0 allocs/op
vs main
BenchmarkPointEqual-12    	486653690	         2.531 ns/op	       0 B/op	       0 allocs/op
```

pkg/timeseries/dataset/point_test.go
```diff
@@ -106,6 +106,14 @@ func TestPointEqual(t *testing.T) {
 	}
 }
 
+func BenchmarkPointEqual(b *testing.B) {
```
crandles, line 109: todo: 
```suggestion
func BenchmarkPointsAreEqual(b *testing.B) {
```

Issue #785, opened 2025-04-13
review gosec security findings


crandles:
`gosec` is a static analysis tool for security problems. ([github](https://github.com/securego/gosec))

The following issues were reported from scanning [main](https://github.com/trickstercache/trickster/commit/9f33abb87f2c5aa391ea6defe5685cc5dd508bfb):

```console
gosec -exclude-generated -fmt=golint -quiet  -exclude G104 ./... | grep -v "gen.go:"
/Users/c/code/oss/trickster/pkg/timeseries/sqlparser/sqlparser.go:173:28: [CWE-190] integer overflow conversion uint64 -> int64 (Rule:G115, Severity:HIGH, Confidence:MEDIUM)
/Users/c/code/oss/trickster/pkg/timeseries/sqlparser/sqlparser.go:171:28: [CWE-190] integer overflow conversion uint64 -> int64 (Rule:G115, Severity:HIGH, Confidence:MEDIUM)
/Users/c/code/oss/trickster/pkg/timeseries/sqlparser/sqlparser.go:169:33: [CWE-190] integer overflow conversion uint64 -> int64 (Rule:G115, Severity:HIGH, Confidence:MEDIUM)
/Users/c/code/oss/trickster/pkg/timeseries/sqlparser/sqlparser.go:167:33: [CWE-190] integer overflow conversion uint64 -> int64 (Rule:G115, Severity:HIGH, Confidence:MEDIUM)
/Users/c/code/oss/trickster/pkg/testutil/timeseries/model/model.go:114:27: [CWE-190] integer overflow conversion uint64 -> int64 (Rule:G115, Severity:HIGH, Confidence:MEDIUM)
/Users/c/code/oss/trickster/pkg/backends/prometheus/model/timeseries.go:133:27: [CWE-190] integer overflow conversion uint64 -> int64 (Rule:G115, Severity:HIGH, Confidence:MEDIUM)
/Users/c/code/oss/trickster/pkg/backends/influxdb/model/marshal.go:97:40: [CWE-190] integer overflow conversion uint64 -> int64 (Rule:G115, Severity:HIGH, Confidence:MEDIUM)
/Users/c/code/oss/trickster/pkg/backends/influxdb/model/marshal.go:92:25: [CWE-190] integer overflow conversion uint64 -> int64 (Rule:G115, Severity:HIGH, Confidence:MEDIUM)
[...]

PR #787, opened 2025-04-13
fix: address gosec G115 overflow potential in epoch.Epoch

crandles:
Address overflow concerns in `epoch.Epoch` usage by utilizing an int64 as the backing type.

~~I'm not sure if we were using uint64 for any particular reason, it wasn't obvious if so.~~ (Confirmed: no outside requirement)

We seem to utilize Epoch around `time.Time{}.UnixNano` which is int64.

Related: #785 (this reduces the number of reported issues from 37 -> 19)

### DIFF
--- a/pkg/timeseries/epoch/epoch.go
+++ b/pkg/timeseries/epoch/epoch.go
@@ -20,7 +20,7 @@ package epoch
 
 // Epoch represents an Epoch timestamp in Nanoseconds and has possible values
 // between 1970/1/1 and 2262/4/12
-type Epoch uint64
+type Epoch int64
 
 // Epochs is a slice of type Epoch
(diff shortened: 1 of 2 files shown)


crandles:
Think i'm wrong here, this probably needs to be a uint64, will look at the conversions closer.

crandles:
I'm going to add a simple `time.Time -> Epoch` function that handles the "overflow" concern; although I don't think there will be one in practice, I think there would only be a concern with "negative" time values.

crandles:
After some discussion, it seems like int64 should be safe here, and it'll remediate the security warnings by avoiding the type conversion entirely -- should save us a few nanoseconds as well for each conversion.

PR #789, opened 2025-04-14
improve extent list performance

jranson:
This flattens `ExtentList.CalculateDeltas` to use an `O(n + m)` merge-sweep in place of the `O(n²)` nested loops, while eliminating appends. It also removes a few appends from ExtentList's `Remove` and `Compress` functions. Optimizes a number of ExtentList functions for splicing, merging, cropping, cloning, and removing ExtentList data.

### DIFF
--- a/pkg/timeseries/extent_list.go
+++ b/pkg/timeseries/extent_list.go
@@ -301,21 +301,11 @@ func (el ExtentList) Clone() ExtentList {
 // CloneRange returns a perfect copy of the ExtentList, cloning only the
 // Extents in the provided index range (upper-bound exclusive)
 func (el ExtentList) CloneRange(start, end int) ExtentList {
-	if end < start || start < 0 || end < 0 {
-		return nil
+	if end < start || start < 0 || end > len(el) {
+		return ExtentList{}
 	}
-	size := end - start
-	if size > len(el) {
-		return nil
-	}
-	c := make(ExtentList, size)
-	j := start
-	for i := 0; i < size; i++ {
-		c[i].Start = el[j].Start
-		c[i].End = el[j].End
-		c[i].LastUsed = el[j].LastUsed
-		j++
-	}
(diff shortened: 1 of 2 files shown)


pkg/timeseries/extent_list.go
```diff
@@ -445,47 +450,53 @@ func (el ExtentList) TimestampCount(d time.Duration) int64 {
...
-				// our earliest datapoint in cache is after the first point the user wants
-				break
+		return ExtentList{need}
+	}
+	sort.Sort(el)
+	out := make(ExtentList, len(el)+1)
+	var missStart time.Time
+	var j, k int
+	for ts := need.Start; !ts.After(need.End); ts = ts.Add(step) {
+		// this advances to j to the point in el where ts would be if it were
```
crandles, line 387: (minor) nit, wording seems off. perhaps:
```suggestion
		// this advances j to the point in el where ts would be if it were
```

pkg/timeseries/extent_list.go
```diff
@@ -514,11 +444,7 @@ func (el ExtentListLRU) Swap(i, j int) {
 // Clone returns a true copy of the ExtentListLRU
 func (el ExtentListLRU) Clone() ExtentListLRU {
 	c := make(ExtentListLRU, len(el))
-	for i := range el {
-		c[i].Start = el[i].Start
-		c[i].End = el[i].End
-		c[i].LastUsed = el[i].LastUsed
-	}
+	copy(c, el)
```
crandles, line 447: q: is copy preferrable over slices.Clone?

crandles (approved):

pkg/timeseries/extent_list.go
jranson: yeah this is faster if we're trying to squeeze out every nanosecond (and we are!)

PR #794, opened 2025-04-19
feat: use time.duration throughout config

### LINKED ISSUE #793: use time.Duration instead if int milliseconds in config
The current trickster configuration prefers time duration to be specified in milliseconds, example:

```yaml
    healthcheck:
      interval_ms: 1000
```

we then have two custom parse functions (duplicated logic) that converts the time to a `time.Duration`.

Instead, we should convert the configs to use the `time.Duration` type from the start, and skip the DIY-parsing.

crandles:
Converts tricksters config to utilize `time.Duration` inputs instead of specifying milliseconds in `int` fields.

Resolves #793

~~To Do:~~
- [x] rebase after #792 merges
- [x] review README / docs / examples further for missing updates

### DIFF
--- a/pkg/backends/healthcheck/healthcheck_test.go
+++ b/pkg/backends/healthcheck/healthcheck_test.go
@@ -62,7 +62,7 @@ func TestRegister(t *testing.T) {
 	logger.SetLogger(testLogger)
 	hc := New().(*healthChecker)
 	o := ho.New()
-	o.IntervalMS = 500
+	o.Interval = 500
 	_, err := hc.Register("test", "test", o, http.DefaultClient)
 	if err != nil {
 		t.Error(err)
@@ -90,7 +90,7 @@ func TestUnregister(t *testing.T) {
 	hc := New().(*healthChecker)
 	logger.SetLogger(testLogger)
 	o := ho.New()
-	o.IntervalMS = 500
+	o.Interval = 500
 	_, err := hc.Register("test", "test", o, http.DefaultClient)
 	if err != nil {
 		t.Error(err)
@@ -106,7 +106,7 @@ func TestStatus(t *testing.T) {
 	logger.SetLogger(testLogger)
 	hc := New().(*healthChecker)
 	o := ho.New()
-	o.IntervalMS = 500
+	o.Interval = 500
 	_, err := hc.Register("test", "test", o, http.DefaultClient)
 	if err != nil {
 		t.Error(err)
@@ -132,7 +132,7 @@ func TestStatuses(t *testing.T) {
 	logger.SetLogger(testLogger)
 	hc := New().(*healthChecker)
 	o := ho.New()
-	o.IntervalMS = 500
+	o.Interval = 500
 	_, err := hc.Register("test", "test", o, http.DefaultClient)
 	if err != nil {
 		t.Error(err)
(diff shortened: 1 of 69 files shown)


pkg/backends/healthcheck/healthcheck_test.go
```diff
@@ -62,7 +62,7 @@ func TestRegister(t *testing.T) {
 	logger.SetLogger(testLogger)
 	hc := New().(*healthChecker)
 	o := ho.New()
-	o.IntervalMS = 500
+	o.Interval = 500
```
jranson, line 65: while this (and the 3 similar operations below)  probably works, the 1:1 would be `500 * time.Millisecond`

pkg/backends/options/options_test.go
```diff
@@ -269,25 +269,11 @@ func TestValidate(t *testing.T) {
 		sw       []intSwapper
 		expected any
 	}{
-		{ // case 0 - MaxShardSizeMS > 0 and MaxShardSizePoints > 0 are mutually exclusive
```
jranson, line 272: Double checking on the removal of this test case. the intent is to error if both `MaxShardSizeMS` and `MaxShardSizePoints` are both non-zero (only one can be, which determines the pruning strategy to use).  In removing MS, `MaxShardSizeMS` might be better named `MaxShardSizeTime` to be better distinguished from `Points`

crandles: the specific issue here is the types are no longer both ints, so the test structure doesn't quite fit

I'll add a more specific test for this

crandles: added a new test

pkg/backends/healthcheck/options/options.go
```diff
@@ -182,14 +182,14 @@ func (o *Options) SetExpectedBody(body string) {
 
 // CalibrateTimeout returns a time.Duration representing a calibrated
 // timeout value based on the milliseconds of duration provided
-func CalibrateTimeout(ms int) time.Duration {
+func CalibrateTimeout(ms time.Duration) time.Duration {
```
jranson, line 185: can we use `d` instead of `ms` here for the variable name?

pkg/backends/prometheus/routes.go
```diff
@@ -68,7 +69,7 @@ func (c *Client) DefaultPathConfigs(o *bo.Options) map[string]*po.Options {
 	var rhts map[string]string
 	if o != nil {
 		rhts = map[string]string{
-			headers.NameCacheControl: fmt.Sprintf("%s=%d", headers.ValueSharedMaxAge, o.TimeseriesTTLMS/1000)}
+			headers.NameCacheControl: fmt.Sprintf("%s=%d", headers.ValueSharedMaxAge, o.TimeseriesTTL/(1000*time.Millisecond))}
```
jranson, line 72: can just do `o.TimeseriesTTL / time.Second`

pkg/cache/registration/registration_test.go
```diff
@@ -95,8 +95,8 @@ func newCacheConfig(t *testing.T, cacheProvider string) *co.Options {
 		BBolt:      &bbo.Options{Filename: "/tmp/test.db", Bucket: "trickster_test"},
 		Badger:     &bao.Options{Directory: bd, ValueDirectory: bd},
 		Index: &io.Options{
-			ReapIntervalMS:        3000,
-			FlushIntervalMS:       5000,
+			ReapInterval:          3000,
+			FlushInterval:         5000,
```
jranson, line 99: missing `* time.Millisecond`

jranson (changes requested):

jranson (approved):

PR #802, opened 2025-04-28
clickhouse updates

jranson:
* improves the dev env's clickhouse data seeder by saving the seed data files from s3 unchanged (with the old dates in tact instead of transforming to current dates). Instead, the date transformation is applied on the database load. This way the seed files can be re-used/transformed indefinitely instead of just once. Every run of the seed container will truncate and re-seed the `trips` table. 
* swaps all non-generated `interface{}` for `any` (project-wide)
* Adds a `CacheKeyElements` field to `TimeRangeQuery`, used to ensure that `DeriveCacheKey` does not have issues with POST/PUT requests.
  * fixes project-wide defect w/ POSTs 
  * used now by ClickHouse and InfluxDB (not required for Prom)
* Fixes misc defects in `pkg/backend/influxdb`, `pkg/backend/clickhouse`,  and `pkg/proxy/params` that were causing issues with proper cache key construction
* Updates ClickHouse JSON Marshaler to use struct-based marshaling.


pkg/proxy/engines/key.go
```diff
@@ -57,87 +57,128 @@ func (pr *proxyRequest) DeriveCacheKey(extra string) string {
...
+		return pc.KeyHasher(r.URL.Path, qp, r.Header, b, trq, extra)
 	}
 
-	vals := make([]string, 0, (len(pc.CacheKeyParams) + len(pc.CacheKeyHeaders) + len(pc.CacheKeyFormFields)*2))
+	var k int
+	vals := make([]string, 2+len(qp)+len(r.Header)+len(pc.CacheKeyFormFields)+ckeCnt)
+	// overrides contains query data modified by the backend provider when
+	// parsing the time range (e.g., a tokenized version of the query statement)
+	var overrides map[string]string
+	used := make(map[string]struct{})
```
crandles, line 82: this could be replaced with our sets package

pkg/proxy/engines/key.go
```diff
@@ -57,87 +57,128 @@ func (pr *proxyRequest) DeriveCacheKey(extra string) string {
...
-	vals = append(vals, fmt.Sprintf("%s.%s.", "method", r.Method))
+	vals[k] = fmt.Sprintf("%s.%s.", "method", r.Method)
+	k++
 
 	if len(pc.CacheKeyParams) == 1 && pc.CacheKeyParams[0] == "*" {
 		for p := range qp {
-			vals = append(vals, fmt.Sprintf("%s.%s.", p, qp.Get(p)))
+			if v, ok := overrides[p]; ok {
+				vals[k] = fmt.Sprintf("%s.%s.", p, v)
+				used[p] = struct{}{}
```
crandles, line 101: could replace this `struct{}{}` with an add via the set package

pkg/proxy/engines/key.go
```diff
@@ -57,87 +57,128 @@ func (pr *proxyRequest) DeriveCacheKey(extra string) string {
...
+			} else {
+				vals[k] = fmt.Sprintf("%s.%s.", p, qp.Get(p))
+			}
+			k++
 		}
 	} else {
 		for _, p := range pc.CacheKeyParams {
+			if v, ok := overrides[p]; ok {
+				vals[k] = fmt.Sprintf("%s.%s.", p, v)
+				used[p] = struct{}{}
```
crandles, line 111: could replace this struct{}{} with an add via the set package

pkg/proxy/request/body.go
```diff
@@ -27,28 +27,32 @@ import (
...
 	if body != nil {
-		return body
+		return body, nil
 	}
 	if r.Body == nil {
-		return nil
+		return nil, nil
+	}
+	var err error
+	body, err = io.ReadAll(r.Body)
```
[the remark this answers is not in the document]
jranson: I would want to make something like that user-configurable (with a reasonable default). So if we can defer to a separate PR for you or me to do, that would be 👌.

crandles (approved):

pkg/proxy/request/body.go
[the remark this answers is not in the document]
crandles: https://github.com/trickstercache/trickster/issues/806

PR #803, opened 2025-04-30
perf: replace cache index mutex with atomics and sync.map

crandles:
Replaces #801

~4x speedup (in simple benchmark) from optimizing the cache index -- replacing mutexes with atomics and sync.Map:

```console
BenchmarkUpdateObject-12    	 4141452	       278.3 ns/op	     204 B/op	       6 allocs/op
```

vs main
```console
BenchmarkUpdateObject-12    	 1000000	      1327 ns/op	     190 B/op	       3 allocs/op
```

This makes it safe to update disjoint index objects concurrently -- given the use of a sync.Map; this should improve overall concurrent usage of the proxy (when using this internal index) -- whereas previously this was guarded by a single mutex. (Although we did appear to limit the actual locking)

This also includes the removal of a mutex from the alb package and a fix to the Makefile.

---

to do:
- [x] carry over benchmarks from #801
- [x] verify test changes


pkg/cache/index/index_test.go
```diff
@@ -110,68 +108,68 @@ func TestReap(t *testing.T) {
...
 		t.Errorf("expected key %s to be missing", "test.1")
 	}
 
-	if _, ok := idx.Objects["test.2"]; ok {
+	if _, ok := idx.Objects.Load("test.2"); ok {
 		t.Errorf("expected key %s to be missing", "test.2")
 	}
 
-	if _, ok := idx.Objects["test.3"]; ok {
+	if _, ok := idx.Objects.Load("test.3"); !ok {
```
crandles, line 142: behavior appears to be a littler different, investigating why still

pkg/util/atomicx/time_test.go
```diff
@@ -0,0 +1,110 @@
...
+		err = at2.UnmarshalBinary(buf)
+		require.NoError(t, err)
+		require.True(t, at2.Equal(at.Load()), "expected %v, got %v", ts, at2.Load())
+		require.True(t, at2.Equal(now), "expected %v, got %v", now, at2.Load())
+	})
+
+	now := time.Now()
+	t.Log(ZeroTime, ZeroTime.Before(now), time.Time{}.Before(now), ZeroTime.Equal(time.Time{}))
+
+	// t.Run("msp", func(t *testing.T) {
```
crandles, line 73: to do: remove this

pkg/cache/index/index_test.go
```diff
@@ -110,71 +108,74 @@ func TestReap(t *testing.T) {
...
+	if _, ok := idx.Objects.Load("test.6"); ok {
 		t.Errorf("expected key %s to be missing", "test.6")
 	}
 
-	if _, ok := idx.Objects["test.7"]; ok {
+	if _, ok := idx.Objects.Load("test.7"); ok {
 		t.Errorf("expected key %s to be missing", "test.7")
 	}
 
+	objects := idx.Objects.ToObjects()
```
crandles, line 176: note -- items are eventually reaped as expected

pkg/cache/index/index_test.go
crandles: adding some debugging locally, this is what I see in this function:
```
=== RUN   TestReap
    index_test.go:120: objects before reap 4
    index_test.go:122: objects after reap 2
    index_test.go:134: objects before reap 5
    index_test.go:136: objects after reap 5
    index_test.go:168: objects before reap 6
    index_test.go:170: objects after reap 1
```
vs similar changes to main
```
=== RUN   TestReap
    index_test.go:121: objects before reap 4
    index_test.go:124: objects after reap 3
    index_test.go:136: objects before reap 6
    index_test.go:138: objects after reap 2
    index_test.go:170: objects before reap 3
    index_test.go:172: objects after reap 1
``

crandles: the initial reap is definitely a bug, some of the others seem ok. continuing to debug. May be something to do with the change in zero time value.

crandles: The issue is definitely with the atomicx.Time zero value; going to rework that

jranson (approved):

PR #819, opened 2025-05-03
cache client refactor

crandles:
Work in progress.

- [x] Convert redis
- [x] Convert memory
- [x] Convert filesystem
- [ ] Convert bbolt
- [ ] Convert bbadger
- [ ] Consider modeling as interface?

### DIFF
--- a/pkg/cache/filesystem/filesystem.go
+++ b/pkg/cache/filesystem/filesystem.go
@@ -23,67 +23,65 @@ import (
 	"os"
 	"strconv"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/trickstercache/trickster/v2/pkg/cache"
 	"github.com/trickstercache/trickster/v2/pkg/cache/index"
+	"github.com/trickstercache/trickster/v2/pkg/cache/internal"
 	"github.com/trickstercache/trickster/v2/pkg/cache/metrics"
 	"github.com/trickstercache/trickster/v2/pkg/cache/options"
 	"github.com/trickstercache/trickster/v2/pkg/cache/status"
-	"github.com/trickstercache/trickster/v2/pkg/locks"
 	"github.com/trickstercache/trickster/v2/pkg/observability/logging"
 	"github.com/trickstercache/trickster/v2/pkg/observability/logging/logger"
 	"github.com/trickstercache/trickster/v2/pkg/util/atomicx"
 )
 
+func NewCache(name string, config *options.Options) *Cache {
+	c := &Cache{
+		lockPrefix: fmt.Sprintf("%s.file.", name),
+	}
+	c.Cache = *internal.NewCache(name, c.lockPrefix, &internal.CacheOptions{
+		Options:  config,
+		Connect:  c.connect,
+		Store:    c.store,
+		Retrieve: c.retrieve,
+		Delete: func(key string) error {
+			return os.Remove(c.getFileName(key))
+		},
+	})
+	return c
+}
+
 // Cache describes a Filesystem Cache
 type Cache struct {
-	Name       string
-	Config     *options.Options
+	internal.Cache
 	Index      *index.Index
-	locker     locks.NamedLocker
 	lockPrefix string
 }
 
-// Locker returns the cache's locker
-func (c *Cache) Locker() locks.NamedLocker {
-	return c.locker
-}
-
-// SetLocker sets the cache's locker
-func (c *Cache) SetLocker(l locks.NamedLocker) {
-	c.locker = l
-}
-
-// Configuration returns the Configuration for the Cache object
-func (c *Cache) Configuration() *options.Options {
-	return c.Config
-}
-
 // Connect instantiates the Cache mutex map and starts the Expired Entry Reaper goroutine
-func (c *Cache) Connect() error {
+func (c *Cache) connect() error {
 	logger.Info("filesystem cache setup", logging.Pairs{"name": c.Name,
 		"cachePath": c.Config.Filesystem.CachePath})
 	if err := makeDirectory(c.Config.Filesystem.CachePath); err != nil {
 		return err
 	}
-	c.lockPrefix = c.Name + ".file."
 
 	// Load Index here and pass bytes as param2
 	indexData, _, _ := c.retrieve(index.IndexKey, false, false)
-	c.Index = index.NewIndex(c.Name, c.Config.Provider, indexData,
+
+	var b []byte
+	if indexData != nil {
+		b = indexData.Value
+	}
+	c.Index = index.NewIndex(c.Name, c.Config.Provider, b,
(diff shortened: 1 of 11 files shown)


pkg/cache/filesystem/filesystem.go
```diff
@@ -105,7 +103,7 @@ func (c *Cache) store(cacheKey string, data []byte, ttl time.Duration, updateInd
 
 	dataFile := c.getFileName(cacheKey)
 
-	nl, _ := c.locker.Acquire(c.lockPrefix + cacheKey)
+	nl, _ := c.Locker().Acquire(c.lockPrefix + cacheKey)
```
crandles, line 106: instead of utilizing this locker in each cache client, I'm going to move this usage up into a wrapper around a much simpler cache client (read(key),write(key, data),delete(keys ...string). + the extra signature for the memory cache

likewise for the index.Index usage

crandles: thinking roughly this for the overall structure:
```mermaid
---
config:
      theme: redux
---
flowchart TD
        A(["Cache 'Manager'"])
        A --> B{"Needs Index?"}
        B --> |Yes| C["index"]
        B --> |No| D["namedLocker"]
        D --> E["cache x client"]
        C --> D
```

crandles: additionally, planning to pull the instrumentation of the cache client to a single layer, if possible

crandles: Potential area of optimization:
- the memory cache is backed by a sync.map
- the index is backed by a sync.map

could these be the same map?

Unsure if separating is beneficial, will look at access patterns.

index is used for reap process, but it it also used per each store and (tbd check where else it is used)

This could be a potentially small change, given already refactoring client. Memory client or index init just needs to change slightly to share a map reference.

crandles:
to do:
- [ ] re-integrate the namedLocker
- [ ] re-integrate metrics over the cache usage
- [ ] tests

.github/workflows/ci-build-tests.yml
```diff
@@ -70,10 +66,6 @@ jobs:
           python-version: '3.x'
       - name: Install python dependencies
         run: python -m pip install --upgrade pip setuptools wheel
-      - name: installing codespell
```
crandles, line 73: todo: Add this back; golangci-lint misspell isn't as affective as I thought it might be

examples/packages/cache/bbolt/main.go
```diff
@@ -16,15 +16,16 @@ func main() {
 	const myKey, myValue = "myKeyName", "myValue"
 
 	// create a new memory cache
-	c, err := bbolt.New("/Users/jranson/test.bbolt", "trickster")
+	c := bbolt.New("bbolt", "/Users/jranson/test.bbolt", "trickster", nil)
```
crandles, line 19: backlog thought: should these examples go to the package instead of an examples package?

e.g. https://go.dev/blog/examples

pkg/cache/index/client.go
```diff
@@ -0,0 +1,465 @@
...
+)
+
+var (
+	ErrIndexInvalidCacheKey = errors.New("cannot store index")
+	ErrInvalidCacheBackend  = errors.New("invalid cache backend for reference access")
+)
+
+// IndexedClientOptions modify an IndexedClient's behavior.
+// Meant for configuring a client for a specific usage -- not meant
+// for general configuration.
```
crandles, line 51:
```suggestion
```

pkg/cache/index/client.go
```diff
@@ -0,0 +1,465 @@
...
+		}
+		if o.FlushInterval > 0 {
+			go idx.flusher(ctx)
+		} else {
+			logger.Warn("cache index flusher was not started",
+				logging.Pairs{"cacheName": idx.name, "flushInterval": o.FlushInterval})
+		}
+	}
+
+	// FIXME: this triggers a failure in pkg/proxy/engines -- need to investigate
```
crandles, line 96: TODO: investigate

pkg/cache/index/client_test.go
```diff
@@ -0,0 +1,268 @@
...
+	_, ok = ic.Objects.Load("test.2")
+	require.False(t, ok, "expected key test.2 to be missing")
+
+	_, ok = ic.Objects.Load("test.3")
+	require.False(t, ok, "expected key test.3 to be missing")
+
+	_, ok = ic.Objects.Load("test.4")
+	require.False(t, ok, "expected key test.4 to be missing")
+
+	t.Skip("wip -- seeing failures here, atime might be broken?")
```
crandles, line 223: todo: investigate

pkg/cache/memory/memory.go
```diff
@@ -23,197 +23,112 @@ import (
...
 		metrics.ObserveCacheOperation(c.Name, c.Config.Provider, "set", "none", float64(l))
-		o1 = &index.Object{Key: cacheKey, Value: byteData, Expiration: *atomicx.NewTime(exp)}
-		o2 = &index.Object{Key: cacheKey, Value: byteData, Expiration: *atomicx.NewTime(exp)}
+		o1 = byteData
+		o2 = byteData
 	} else if refData != nil {
-		metrics.ObserveCacheOperation(c.Name, c.Config.Provider, "setDirect", "none", 0)
-		o1 = &index.Object{Key: cacheKey, ReferenceValue: refData, Expiration: *atomicx.NewTime(exp)}
-		o2 = &index.Object{Key: cacheKey, ReferenceValue: refData, Expiration: *atomicx.NewTime(exp)}
+		metrics.ObserveCacheOperation(c.Name, c.Config.Provider, "setDirect", "none", float64(refData.Size()))
```
crandles, line 93: todo: is this moved up to manager level?

pkg/proxy/engines/cache.go
```diff
@@ -330,7 +331,7 @@ func writeConcurrent(_ context.Context, c cache.Cache, key string, d *HTTPDocume
 
 	// for memory cache, don't serialize the document, since we can retrieve it by reference.
 	if c.Configuration().Provider == "memory" {
-		mc := c.(cache.MemoryCache)
+		mc := c.(*manager.Manager)
```
crandles, line 334: consider using interface here?

examples/packages/cache/bbolt/main.go
jranson: The idea here was that is easy to click through all of the examples in one place in case one might prove interesting.  I think they are fine to relocate, but we would need a new Markdown doc that curates all of the important examples into a single list / place. It's fine to file an issue for the curated doc and we can address later. But the path should probably just be `"./test.bbolt"` without the absolute home directory. Clearly a copy/paste oversight. Thanks!

jranson (approved):

PR #822, opened 2025-05-13
clickhouse & influxdb updates

jranson:
This is a very large change - sorry!

* New Makefile actions `check-fmtprints` and `check-todos` to make sure no new `fmt.Print*` or `TODO:` items are added to the codebase.
* Refactors `bytes.NewReader([]byte("mystring"))` to use `strings.NewReader("mystring")`
* Adds convenience func to Detect `Accept` and `Content-Type` headers for known mime types
* Addresses a longstanding `TODO` where `request.SetBody()` was required to return the updated `*http.Request` due to a context change. An element inside of the context is now changed to avoid the reassignment.
* Fixes an nit where variables and comments used `Dataset` when the struct type is `DataSet`
* Adds a `numbers` Package that includs `SafeAdd` functions
* Adds a `CounterSet` capability to atomically maintain counter values in map
* Consolidates Fields in `timeseries.TimeRangeQuery`, `timeseries.RequestOptions`, `timeseries.FieldDefinition' so that all backend TSDB providers use similar fields in similar ways
* Adds a CSV -> DataSet parser that is adopted by both ClickHouse and InfluxDb/Flux (which serve data from the origin via CSV)
* Flux wasn't actually working, and InfluxQL was only mostly working. This bifurcates InfluxDB Flux and InfluxDB InfluxQL implementations into separate packages instead of trying to both in a single package, and they both work reliably.
* Fix linter nits around unused parameters and non-conforming names.


Makefile
```diff
@@ -212,7 +215,51 @@ check-license-headers:
...
+	echo "" ; echo "\033[1;32m✓\033[0m No unexpected fmt.Print* calls." ; echo ""
+
+.PHONY: check-todos
+check-todos: SHELL:=/bin/sh
+check-todos: # fails if there are any fmt.Print* calls outside of the 3 approved files
+	@cd pkg && \
+	todos=$$(git grep -in todo | grep -v 'context\.TODO'); \
+	count=0; \
+	if [ -n "$$todos" ]; then \
+		count="$$(echo "$$todos" | wc -l | tr -d '[:space:]')" ; \
```
crandles, line 247: fyi, golangci-lint can do this: https://golangci-lint.run/usage/linters/#godox

pkg/backends/influxdb/influxql/influxql.go
```diff
@@ -0,0 +1,217 @@
...
+	if f.IsPost() {
+		b, err := request.GetBody(r)
+		if err != nil {
+			return nil, nil, false, err
+		}
+		trq.OriginalBody = b
+	} else {
+		qv := url.Values(http.Header(uv).Clone())
+		qv.Set(ParamQuery, trq.Statement)
+		// Swap in the Tokenzed Query in the Url Params
```
crandles, line 175:
```suggestion
		// Swap in the Tokenized Query in the Url Params
```

Makefile
jranson: Currently we have linting of `*_test.go` disabled (enabling it would create hundreds of new issues), and I want to pick up the 36% of TODOs in the test files. I would propose switching this to the linter once we have burned down the current list to 0 and have the linter enabled for tests too.

crandles (approved):

PR #832, opened 2025-05-22
add authenticator feature w/ basic auth

jranson:
This PR adds an `Authenticator` capability to Trickster that allows any Backend to be protected with an Authentication wall. It initially supports Basic Auth. User Pool data can be loaded from an `htpasswd`-formatted file, a csv, or embedded directly into the Trickster config yaml. Embedded user manifests support env injection.

### DIFF
--- a/examples/conf/example.auth.yaml
+++ b/examples/conf/example.auth.yaml
@@ -23,6 +23,7 @@ authenticators:
     users_file_format: csv
     users: # optional embedded users list (name: credential)
       user1: red123
+      user2: ${USER2_PASS} # password for user2 is sourced from the ENV
     users_format: plaintext
     config: # optional provider-specific configs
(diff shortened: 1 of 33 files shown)


examples/conf/example.auth.yaml
```diff
@@ -0,0 +1,42 @@
...
+
+authenticators:
+  # loads users from a CSV and an embedded plaintext manifest, causes the browser to show the login form
+  example_auth_1:
+    provider: basic # http basic auth (required)
+    proxy_preserve: true # the authenticator will preserve auth headers instead of stripping them
+    users_file: /path/to/user-manifest.csv # optional users source file
+    users_file_format: csv
+    users: # optional embedded users list (name: credential)
+      user1: red123
```
crandles, line 25: should we add an env var example?

crandles (approved):

PR #835, opened 2025-05-23
fix: [backend/rules] rmatch is operation not input type

### LINKED ISSUE #677: rules code question
image
 
in rule doc , input_type permitted values ( string num bool ).  don't understand   this line .

crandles:
This resolves #677, which points out what looks to be a bug in the rules option parsing code.

`rmatch` is not an `InputType`, it is an `Operation`.

This is causing a panic and needs investigating.


crandles:
this panic appears to be in main: https://github.com/trickstercache/trickster/actions/runs/15216945479/job/42804534628

PR #837, opened 2025-05-24
wip: convert the time series extent list to a generic implementation

crandles:
- [x] POC extent list conversion
  - should support: time.Time, int64 or arbitrary types  
- [ ] "in-place" rewrite of the existing extent_list code
  - [ ] avoid field access in favor of functions
  - [ ] add necessary helper functions to extent type
- [ ] convert refactored extent list to generic
  -  TBD on whether I'll save the byterange conversion to another PR

---

```go
        // working
	var e ranges.Extent[time.Time]
	e = timeseries.Extent{Start: time.Unix(100, 0), End: time.Unix(1000, 0)}
	require.True(t, e.StartsAt(time.Unix(100, 0)))
	list := ranges.ExtentList[time.Time]{e}
...
        // goal -- support the byterange.Range type
	var e2 ranges.Extent[int64]
	e = byterange.Range{Start:5, End: 5}
	require.True(t, e.StartsAt(5))
```

### DIFF
--- a/pkg/proxy/ranges/ranges.go
+++ b/pkg/proxy/ranges/ranges.go
@@ -1,0 +1,187 @@
+/*
+ * Copyright 2018 The Trickster Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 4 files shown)


crandles:
fyi a bit stalled here.

I pivoted to more of a generic approach (vs interfaces). I was able to convert the extent / extent list code. However, when working it back into usage, it became overly verbose. Likely need to start over.

PR #849, opened 2025-05-30
segments package

jranson:
adds a `segments` package that provides a single generic `Diff` function for all slices that have structs w/ start/ends. `Diff` replaces the difference logic in all the various `CalculateDelta[s]`.

- [x] Segments Package w/ Generic `Diff()`
- [x] time.Time and Int64 Diffable implementations 
- [x] timeseries.ExtentList working / tests passing
- [x] standardize on `CalculateDeltas` (plural) and use` haves.CalculateDeltas(needs)` and not `needs.CalculateDeltas(haves)`
- [x] byterange.Ranges working / tests passing

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -57,8 +57,10 @@ jobs:
         env:
           GH_TOKEN: ${{ github.token }}
         run: |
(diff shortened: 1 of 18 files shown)


jranson:
Looks like i goofed my branch. Will resubmit another PR.

Issue #806, opened 2025-05-01
io.ReadAll: review usage and consider replacing with configurable io.LimitReader


crandles:
Review `io.ReadAll` usage throughout codebase, identifying usage that could be replaced with an `io.LimitReader`.

`io.LimitReader` size should be user configurable or e.g. depend on discovered Content Length or alternative headers.

This safeguard may help prevent against some denial of service attacks.

jranson:
I’m wondering if it would be possible — and a good approach — to use a first-hop middleware handler that inspects the request and, whenever it’s a PUT, POST, or PATCH, reads and starts streaming the request body into a new buffer via a limitreader before passing the request to the next handler with the new buffer set as r.Body. This way, we can check the request body length once and in one centralized place across the entire app. We would need to either update the Content-Length header accordingly or return an error code whenever the limitreader truncates. 

As a proxy/cache, we’ll always need to inspect the request body, so it makes sense to me to just do it once upfront. Currently Backends call pkg/proxy/request/body.go:GetBody() when they need to inspect it. In GetBody(), if the body isn’t already cached in the context, it is read from r.Body and added to the context for subsequent GetBody calls. However, some backends — or even engine functions like DeltaProxyCache — _may_ still use their own io.ReadAll instead of GetBody if they were missed during refactoring.

As part of instrumenting a limitreader middleware, we should also identify and replace (if they exist) any remaining direct io.ReadAll calls on the request body outside of the new middleware with calls to request.GetBody(). GetBody should be updated to only check the context for the body (and never try to read r.Body, since the proposed limitreader middleware would have already read and saved the request body byte slice to the context when applicable) and return it — including when the slice in the context is nil.

Tagging @trumant for input.

As a side note, our middleware is currently scattered between pkg/proxy/handlers/trickster (all in separate packages alongside non-middleware handler packages) and pkg/util/middleware (which contains several unrelated handlers lumped together in one large middleware package). We should reorganize their locations—for example, placing all middleware under pkg/proxy/middleware with per-handler subpackages — so that when we add the proposed limitreader middleware, it goes to its proper, final destination instead of contributing to the existing disorder.

trumant:
@jranson I like where you're head is at here, but need to dive in to the existing codebase further before sharing any truly well informed reactions.

PR #854, opened 2025-06-21
add Request Body Size Limiter

### LINKED ISSUE #808: net/http.Request.*Body: consider usage vs custom pkg/proxy/request.*Body
Consider the current internal implementation of `pkg/proxy/request.(Get|Set)Body` vs the standard library functions ([link](https://pkg.go.dev/net/http#Request)).

jranson:
This patch fixes #808 and #806 by centralizing reading of the request body to a middleware handler that also implements the size limiter using an `io.LimitReader` as part of the centralized read. The default max request body size is 10MB but can be udpated in the config yaml. This patch also addresses POST-related defects in the ALB due to incompatibilities between the various non-centralized, slightly different DRY body reading implementations that are now refactored.

### DIFF
--- a/docs/body.md
+++ b/docs/body.md
@@ -1,0 +1,20 @@
+# Request Body Handling Customizations
+
+## Request Body Size Limiter
+
+By default, the max allowed Request Body size is 10 MB. If the client request
+body in a POST, PUT or PATCH are over 10 MB, the request will receive a
+response of `413 Request Payload is too large`
+
+You can change (or bypass) this limit in Frontend Config Section:
+
+```yaml
+frontend:
+    max_request_body_size_bytes: 5000 # request bodies must be <= 5kb or returns 413
+```
+
+```yaml
+frontend:
+    max_request_body_size_bytes: 5000 # request bodies > 5kb truncated to 5kb
+    truncate_request_body_too_large: true
+```
--- a/docs/body.md
+++ b/docs/body.md
@@ -10,11 +10,11 @@ You can change (or bypass) this limit in Frontend Config Section:
 
 ```yaml
 frontend:
-    max_request_body_size_bytes: 5000 # request bodies must be <= 5kb or returns 413
+  max_request_body_size_bytes: 5120 # request bodies must be <= 5kb or returns 413
 ```
 
 ```yaml
 frontend:
-    max_request_body_size_bytes: 5000 # request bodies > 5kb truncated to 5kb
-    truncate_request_body_too_large: true
+  max_request_body_size_bytes: 5120 # request bodies > 5kb truncated to 5kb
+  truncate_request_body_too_large: true # truncate request bodies that are too large
 ```
--- a/pkg/backends/options/defaults.go
+++ b/pkg/backends/options/defaults.go
@@ -58,8 +58,8 @@ const (
 	DefaultMaxIdleConns = 20
 	// DefaultForwardedHeaders defines which class of 'Forwarded' headers are attached to upstream requests
 	DefaultForwardedHeaders = "standard"
-	// DefaullALBMechansimName defines the default ALB Mechanism Name
-	DefaullALBMechansimName = "rr" // round robin
+	// DefaultALBMechansimName defines the default ALB Mechanism Name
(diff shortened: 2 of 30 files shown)


docs/body.md
```diff
@@ -0,0 +1,20 @@
...
+body in a POST, PUT or PATCH are over 10 MB, the request will receive a
+response of `413 Request Payload is too large`
+
+You can change (or bypass) this limit in Frontend Config Section:
+
+```yaml
+frontend:
+    max_request_body_size_bytes: 5000 # request bodies must be <= 5kb or returns 413
+```
+
```
crandles, line 15: _Maybe_ an extra sentence here?

```suggestion

To truncate request bodies that are too large, set `frontend.truncate_request_body_too_large: true`:
```

pkg/backends/options/defaults.go
```diff
@@ -58,8 +58,8 @@ const (
 	DefaultMaxIdleConns = 20
 	// DefaultForwardedHeaders defines which class of 'Forwarded' headers are attached to upstream requests
 	DefaultForwardedHeaders = "standard"
-	// DefaullALBMechansimName defines the default ALB Mechanism Name
-	DefaullALBMechansimName = "rr" // round robin
+	// DefaultALBMechansimName defines the default ALB Mechanism Name
+	DefaultALBMechansimName = "rr" // round robin
```
crandles, line 62:
```suggestion
	DefaultALBMechanismName = "rr" // round robin
```

pkg/frontend/options/options.go
```diff
@@ -34,6 +36,14 @@ type Options struct {
 	TLSListenPort int `yaml:"tls_listen_port,omitempty"`
 	// ConnectionsLimit indicates how many concurrent front end connections trickster will handle at any time
 	ConnectionsLimit int `yaml:"connections_limit,omitempty"`
+	// MaxRequestBodySize indicates the maximum allowed size of the request body.
+	// If the body is too large. Trickster will truncate the payload or return a
+	// 413 Payload Too Large response depending upon truncate_request_body_too_big.
+	// Use 0 for no body allowed, and < 0 for no maximum.
+	MaxRequestBodySizeBytes *int64 `yaml:"max_request_body_size_bytes"`
```
crandles, line 43: q: why is this a pointer?

we're applying a default (L64), so it doesn't seem like this is to allow us to tell if the field is unset, unless i'm missing something subtle.

pkg/util/middleware/bodyfilter/bodyfilter.go
```diff
@@ -0,0 +1,46 @@
...
+package bodyfilter
+
+import (
+	"net/http"
+
+	"github.com/trickstercache/trickster/v2/pkg/proxy/handlers/trickster/failures"
+	"github.com/trickstercache/trickster/v2/pkg/proxy/request"
+)
+
+// Handler returns a handler for the BodyFilter, which supports
```
crandles, line 26: comment looks unfinished

pkg/proxy/request/request.go
```diff
@@ -17,3 +17,36 @@
...
+	}
+	rsc := GetResources(r)
+	if rsc != nil {
+		rsc = rsc.Clone()
+	}
+	ctx := context.Background()
+	if rsc != nil {
+		ctx = tctx.WithResources(ctx, rsc.Clone())
+	}
+	out := r.Clone(context.Background()).WithContext(ctx)
```
crandles, line 42: why not set the context during the initial clone?
```suggestion
	out := r.Clone(ctx)
```

pkg/frontend/options/options.go
jranson: This matters.  The default is 10MB when the pointer is nil, but the user can supply `0` to indicate no body allowed. So if this is not a pointer, it is not possible to know if the user wants to use the default (10MB) or 0.

pkg/backends/options/defaults.go
jranson: Will need to change the references to this too, so i'll make a commit that includes `sim` -> `ism` here in the definition and also update the couple of places where it is referenced.

PR #857, opened 2025-07-07
test: add simple integration test

crandles:
* Extends `pkg/daemon.Start` and `pkg/daemon/signaling` with `context.Context` for cancellation
* Adds an `integration` top-level package and simple test calling equivalent to trickster binary
* Adds a new PR workflow job that runs the integration test

---

To Do:
- [ ] add prometheus backend tests

### DIFF
--- a/.github/workflows/ci-build-tests.yml
+++ b/.github/workflows/ci-build-tests.yml
@@ -99,9 +99,19 @@ jobs:
           path-to-profile: .coverprofile
           flag-name: Go-${{ matrix.go }}
           parallel: true
+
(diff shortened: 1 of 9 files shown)


crandles:
I could see ordering being a nice to have there. depending on when backends return their results we may end up with a different result. Seems like the label merge strategy is probably worth following up on.

crandles:
Good to know

Though to clarify, I was talking about the order of the ALB responses -- if the ALB members return different responses for this endpoint, the order in which we perform the merge could impact the end result.

Issue #870, opened 2025-09-07
Refactor WaitGroup usage for Go 1.25


Elia-Renzoni:
@jranson 

Now that the project has been updated to Go 1.25, I will refactor the current usage of WaitGroup to adopt the more idiomatic approach recommended in this version.
The goal of this change is to improve readability, maintainability, and consistency across the codebase, ensuring that contributors and users benefit from clearer concurrency patterns.

Expected outcome:

* WaitGroup usage follows Go 1.25 idiomatic patterns.

* Code is easier to read and reason about.

* The project stays consistent with current Go standards.

PR #872, opened 2025-09-25
chore(waitgroup): update usage to Go 1.25

Elia-Renzoni:
This PR updates the usage of WaitGroups to align with Go 1.25.

Changes
- Update WaitGroup usage to the Go 1.25 API.
- Remove closures that previously captured loop/index variables.
- Introduce helper variables to avoid index-out-of-bound errors that appeared after removing those closures.

Notes
- The helper variables were added because the update removed closures that previously enclosed loop/index variables; without these helpers the code produced index-out-of-bound errors. 
- If the current helper variable names are unclear or not descriptive enough, I'm happy to rename them based on your suggestions.
- Some parts of the updated code may still be optimized further; I welcome suggestions and feedback on possible improvements.

Related issue: #870 

Thanks for reviewing — any feedback is appreciated.

### DIFF
--- a/pkg/backends/influxdb/influxql/unmarshal.go
+++ b/pkg/backends/influxdb/influxql/unmarshal.go
@@ -101,13 +101,9 @@ func UnmarshalTimeseriesReader(reader io.Reader, trq *timeseries.TimeRangeQuery)
 			errs := make([]error, len(wfd.Results[i].SeriesList[j].Values))
 			for vi, v := range wfd.Results[i].SeriesList[j].Values {
 				wg.Go(func() {
-					var (
-						vals  = v
-						idx = vi
-					)
-					pt, cols, err := pointFromValues(vals, sh.TimestampField.OutputPosition)
+					pt, cols, err := pointFromValues(v, sh.TimestampField.OutputPosition)
 					if err != nil {
-						errs[idx] = err
+						errs[vi] = err
 						return
 					}
 					if pt.Epoch == 0 {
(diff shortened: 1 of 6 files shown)


Elia-Renzoni:
@jranson  I was wondering if it might be possible to merge it, so I can work on some new changes without running into conflicts.
Please let me know if there’s anything else I should adjust.
Thanks a lot 🙏

pkg/backends/influxdb/influxql/unmarshal.go
```diff
@@ -99,29 +99,29 @@ func UnmarshalTimeseriesReader(reader io.Reader, trq *timeseries.TimeRangeQuery)
 			var sz int64
 			var wg sync.WaitGroup
 			errs := make([]error, len(wfd.Results[i].SeriesList[j].Values))
-			wg.Add(len(wfd.Results[i].SeriesList[j].Values))
 			for vi, v := range wfd.Results[i].SeriesList[j].Values {
-				go func(vals []any, idx int) {
+				wg.Go(func() {
+					var (
+						vals  = v
```
jranson: @Elia-Renzoni I don't think the loop redeclarations like this are necessary. As of go 1.22, `for` loop variables [are properly assigned per-iteration](https://go.dev/blog/loopvar-preview), so they should be concurrency-safe when passed into goroutines.

pkg/backends/prometheus/model/timeseries.go
```diff
@@ -239,16 +239,18 @@ func populateSeries(ds *dataset.DataSet, result []*WFResult,
 		if !isVector && l > 0 {
 			pts = make(dataset.Points, l)
 			var wg sync.WaitGroup
-			wg.Add(len(pr.Values))
 			for i, v := range pr.Values {
-				go func(index int, vals []any) {
+				wg.Go(func() {
+					var (
+						index  = i
```
jranson: `index` and `vals` should not be needed.

jranson (changes requested):
@Elia-Renzoni thanks for this, and sorry I needed more time to review it. As of go 1.22, `for` loop variables [are properly assigned per-iteration](https://go.dev/blog/loopvar-preview), so they should be concurrency-safe when passed into goroutines. See notes below for when you can do that without needing any loop helper vars. The index-out-of-bounds errors are specific to variables that are not controlled by / declared in the for clause, so those non-clause helpers were OK to keep but must be declared before/outside of the `wg.Go()` to avoid data race conditions.

jranson (approved):

PR #897, opened 2025-12-11
perf: work in progress

crandles:
WIP:

## Changes

1. Remove 2 extra goroutines from being created per usage of the `pkg/timeseries/dataset` cropper(s).

Instead of looking for the start and end time in two separate goroutines, look for the start and end at the same time.

2. Fix the ALB "first responder" mechanism: previously, the non-first responses were not being cancelled as described.

---

## TODO

- [x] make waitgroup usage consistent
- [ ] look for goroutine optimizations
  - [x] pkg/timeseries/dataset -- look for start & end sequentially; end check is improved because we check after known start index.
  - [x] pkg/backends/alb/mech/fr -- ensure cancellation of secondary responses
- [x] look at [errgroup](https://pkg.go.dev/golang.org/x/sync/errgroup) & [SetLimit](https://pkg.go.dev/golang.org/x/sync/errgroup#Group.SetLimit) for concurrency limiting
  - [x] initial adds to pkg/proxy/engines cache (read/write)
  - [ ] expose via config options

---

Notes:

Deleted benchmark:
```
go test -bench=BenchmarkFindRange ./pkg/timeseries/dataset

goos: darwin
goarch: arm64
pkg: github.com/trickstercache/trickster/v2/pkg/timeseries/dataset
cpu: Apple M2 Pro
BenchmarkFindRange/Combined-12         	47423721	        24.65 ns/op
BenchmarkFindRange/Individual-12       	28404972	        41.20 ns/op
PASS
ok  	github.com/trickstercache/trickster/v2/pkg/timeseries/dataset	2.668s
```

### DIFF
--- a/pkg/backends/influxdb/influxql/unmarshal.go
+++ b/pkg/backends/influxdb/influxql/unmarshal.go
@@ -19,10 +19,8 @@ package influxql
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
 	"io"
 	"sort"
-	"sync"
 	"time"
 
 	"github.com/trickstercache/trickster/v2/pkg/timeseries"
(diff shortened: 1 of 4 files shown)


pkg/backends/prometheus/model/timeseries.go
```diff
@@ -239,22 +237,17 @@ func populateSeries(ds *dataset.DataSet, result []*WFResult,
 		var ps int64 = 16
 		if !isVector && l > 0 {
 			pts = make(dataset.Points, l)
-			var wg sync.WaitGroup
```
crandles: todo: consider breaking out into separate pr

pkg/backends/influxdb/influxql/unmarshal.go
```diff
@@ -97,32 +94,27 @@ func UnmarshalTimeseriesReader(reader io.Reader, trq *timeseries.TimeRangeQuery)
 			sh.CalculateSize()
 			pts := make(dataset.Points, len(wfd.Results[i].SeriesList[j].Values))
 			var sz int64
-			var wg sync.WaitGroup
-			errs := make([]error, len(wfd.Results[i].SeriesList[j].Values))
 			for vi, v := range wfd.Results[i].SeriesList[j].Values {
-				wg.Go(func() {
```
crandles: todo: consider breaking out into separate pr

pkg/timeseries/dataset/dataset.go
```diff
@@ -146,18 +146,8 @@ func (ds *DataSet) CroppedClone(e timeseries.Extent) timeseries.Timeseries {
 				sc := &Series{
 					Header: s.Header.Clone(),
 				}
-				start, end, l := 0, -1, len(s.Points)
-				var iwg sync.WaitGroup
```
crandles: todo: consider breaking out into separate pr

PR #903, opened 2025-12-16
refactor alb mechanisms

jranson:
This PR makes improvements to the ALB for performance and un-dry:

* Adds `Pool()` and `HealthyTargets()` functions to ALB Mechanism interface
* Exports consts for mechanism names and utilizes throughout
* Removes `ResponseGate` and `ResponderClaim` types in favor of a generic `CaptureResponseWriter`
* Refactors/simplifies all mechanisms to use `CaptureResponseWriter`

Also:
* Added bonus, revives the `caller` log entry field from trickster 1.x and earlier versions of trickster 2.x that had disappeared, like:
```time=2025-12-16T23:46:50.587335Z app=trickster level=warn event="configuration reload starting now" caller=pkg/daemon/daemon.go:132 source=sighup```
* Updates the default HTTP/TCP backend timeouts as follows: 
  * `keep_alive_timeout`: 300s -> 120s
  * `timeout`: 180s -> 60s
* Sets backend http clients' `IdleConnTimeout` value based on `o.KeepAliveTimeout` to close stale goroutines, and adds a max 10s TCP/HTTP/TLS connect/handshake timeout.

### CODE REVIEW

File: pkg/util/middleware/metrics.go [hunk 1]
```diff
@@ -56,15 +56,15 @@ type responseObserver struct {
...
-	case statusCode >= 300 && statusCode < 399:
+	case statusCode >= 300 && statusCode < 400:
 		w.status = "3xx"
-	case statusCode >= 400 && statusCode < 499:
+	case statusCode >= 400 && statusCode < 500:
 		w.status = "4xx"
-	case statusCode >= 500 && statusCode < 599:
+	case statusCode >= 500 && statusCode < 600:
 		w.status = "5xx"
 	}
```
crandles: 🤦 I was just looking at this and missed this flaw.

Should we go one step further and remove the switch statement?

```suggestion
	w.status = fmt.Sprintf("%dxx", statusCode / 100)
```

pkg/backends/alb/mech/nlm/newest_last_modified.go
```diff
@@ -81,111 +84,84 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		statusCode := crw.StatusCode()
+		w.WriteHeader(statusCode)
+		w.Write(crw.Body())
+
+		// Cancel all other contexts
+		for j, cancel := range cancels {
+			if j != newestIdx {
+				cancel()
+			}
+		}
```
crandles: i don't think this cancel is necessary (here and in the next branch around L160)

on L132 we are waiting for all responses to complete. Is there anything to cancel at this point, aren't the contexts done being used?

pkg/backends/alb/mech/fr/first_response.go
```diff
@@ -101,67 +108,79 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+	serveAndCancelOthers := func(i int, crw *capture.CaptureResponseWriter) {
+		go func() {
+			// cancels all other contexts
+			for j, cancel := range cancels {
+				if j != i {
+					cancel()
+				}
+			}
+		}()
```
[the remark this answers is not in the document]
jranson: What if in the fallback section, we change `atomic.LoadInt64(&claimed) == -1` to `atomic.CompareAndSwapInt64(&claimed, -1, -2)`.  Then only the first response or the fallback (but not both) can call `serveAndCancelOthers`, with the fallback basically being index -2 while the main section continues to specifically swap only if -1. That should do for us what a separate mutex would.

(pushed this ^^ up to the PR)

crandles (approved):

crandles (approved):

crandles (approved):

Issue #493, opened 2020-10-01
Subdirectory support for configs


jranson:
We currently support a single TOML-based config file (by default living at `/etc/trickster/trickster.conf`). We would like to provide a more robust configuration solution by permitting an entire directory of config files to be concatenated when constructing the running config. A good approach might be to look for a `conf.d` subdirectory in the same directory as the trickster.conf file and sourcing anything within, while optionally providing the user a configuration parameter in the main config to specify the include directory.

mohdasif2294:
Hey, @jranson Can you please assign this. I have skimmed through the codebase. It will require code and deployment modifications.

jranson:
Hi mohdasif2294, just wanted to check on this and see how it is going? Thanks!

jranson:
@mohdasif2294 i went ahead and opened this back up to any interested person. if you have made progress, please let us know here and we will re-assign to you again.

Tanmoy741127:
Is anyone working on this issue? If not, can you assign the issue to me?

houyuwushang:
I missed #908 when I posted this earlier. I will not open a competing implementation; I will keep the local findings for reviewing the existing PR.

houyuwushang:
@jranson @Elia-Renzoni I revisited this because the current #908 head no longer compiles on `main`, and the directory add/remove and reload-state concerns from the review are still present. I have a current-main branch that keeps #908's `-config <directory>` behavior and also covers the original file + `conf.d`/configured include-directory path from this issue.

I don't want to open a competing PR or add to the review queue without agreement, especially under the new five-PR guideline. Would you prefer that I open this as a replacement once my current queue is within the limit, or make the commits available for #908 to take? The branch is here: https://github.com/houyuwushang/trickster/tree/feat-config-directory-493

Elia-Renzoni:
@houyuwushang 
Feel free to checkout my PR and push directly to it if you'd like to make changes. I think that would be the easiest way to work on this together.

houyuwushang:
That works for me, but I don't currently have push access to `Elia-Renzoni/trickster:config-493`, so GitHub won't let me update #908 directly. Could you add `houyuwushang` as a collaborator on your fork? Once that's done, I can push the changes to the existing PR branch.

Elia-Renzoni:
@houyuwushang 
Sorry for the delay, I forgot to send it earlier! I have now added you as a collaborator, you should get an invite from GitHub shortly.

jranson:
Fixed in #908

PR #908, opened 2025-12-29
feat: add support for loading multiple config files and staleness check

Elia-Renzoni:
## Description

This PR improves Trickster’s configuration system by extending support from a single configuration file to multiple YAML files. It automatically loads and merges all YAML files when a directory path is provided (e.g., /etc/trickster), while keeping -config <directory> support and adding primary file + conf.d handling.

Related Issue: #493 

## Type of Change

- - [ ] Bug fix
- - [x] New feature
- - [ ] Optimization
- - [ ] Test coverage
- - [ ] Documentation
- - [ ] Infrastructure

## AI Disclosure


houyuwushang (commented):
I rebased these two commits onto current `main` and traced the directory path through `daemon.Hup`. The basic case with two existing files still works, but I found several cases that can miss configuration changes or break existing callers. I left the reproductions and affected behavior inline. The main structural issue is that staleness is represented as parallel slices for only the files present at startup, rather than as a snapshot of the directory contents.

pkg/config/config.go
```diff
@@ -309,15 +391,31 @@ func (c *Config) IsStale() bool {
...
 	if c.MgmtConfig == nil {
 		c.MgmtConfig = mgmt.New()
 	}
 	c.Main.configRateLimitTime = time.Now().Add(c.MgmtConfig.ReloadRateLimit)
-	t := c.CheckFileLastModified()
+
+	if len(c.Main.configFilesPath) > 0 {
+		for index, file := range c.Main.configFilesPath {
+			t := c.CheckFileLastModified(file)
+			if t.IsZero() {
```
houyuwushang: This only walks `configFilesPath` captured during the initial load. A newly added YAML file is never examined, while a removed file returns a zero timestamp and is skipped here. `daemon.Hup` calls this method before loading again, so both changes make an explicit reload report "not stale" and leave the running config unchanged. I reproduced both cases after rebasing onto current `main`. Could this compare a fresh directory snapshot (names and modification times) with the loaded snapshot instead?

pkg/config/config.go
```diff
@@ -287,8 +354,10 @@ func (c *Config) IsStale() bool {
 	c.Main.stalenessCheckLock.Lock()
 	defer c.Main.stalenessCheckLock.Unlock()
 
-	if c.Main == nil || c.Main.configFilePath == "" ||
-		time.Now().Before(c.Main.configRateLimitTime) {
+	if c.Main == nil ||
+		(len(c.Main.configFilesPath) == 0 &&
+			(c.Main.configFilePath == "" ||
+				time.Now().Before(c.Main.configRateLimitTime))) {
```
houyuwushang: For directory-backed configs, `len(c.Main.configFilesPath) == 0` is false, so the rate-limit check is never evaluated. After an existing file changes, `IsStale()` scans the files and returns true on every call during the rate-limit window. The rate-limit condition should apply to both single-file and directory modes.

pkg/config/config.go
```diff
@@ -172,15 +228,24 @@ func (c *Config) loadYAMLConfig(yml string) error {
 }
 
 // CheckFileLastModified returns the last modified date of the running config file, if present
-func (c *Config) CheckFileLastModified() time.Time {
-	if c.Main == nil || c.Main.configFilePath == "" {
+func (c *Config) CheckFileLastModified(confFile string) time.Time {
```
houyuwushang: This changes an exported v2 method from `CheckFileLastModified()` to `CheckFileLastModified(string)`, which breaks external callers at compile time. Could the path-taking helper stay private while retaining the existing no-argument method for compatibility?

pkg/config/config.go
```diff
@@ -353,7 +451,10 @@ func (c *Config) String() string {
 // ConfigFilePath returns the file path from which this configuration is based
 func (c *Config) ConfigFilePath() string {
 	if c.Main != nil {
-		return c.Main.configFilePath
+		if len(c.Main.configFilesPath) == 0 {
+			return c.Main.configFilePath
+		}
+		return c.Flags.ConfigPath
```
houyuwushang: `Clone()` copies `configFilesPath` but does not copy `Flags`, so `cfg.Clone().ConfigFilePath()` panics for every directory-backed config at this dereference. I reproduced that on current `main`. Storing the directory path in `MainConfig` and deep-copying the directory snapshot would avoid making this accessor depend on runtime flags.

pkg/config/config.go
```diff
@@ -143,10 +160,49 @@ func (c *Config) loadFile(flags *Flags) error {
...
 	}
 	c.Main.configFilePath = flags.ConfigPath
-	c.Main.configLastModified = c.CheckFileLastModified()
+	c.Main.configLastModified = c.CheckFileLastModified("")
 	return nil
 }
 
+// loadAndMergeFiles loads application configuration from multiple YAML files
+func (c *Config) loadAndMergeFiles(flags *Flags) error {
+	files, err := fs.Glob(os.DirFS(flags.ConfigPath), "*.yaml")
```
houyuwushang: This only matches `.yaml`. A directory containing `trickster.yml` loads no files and eventually returns `no valid backends configured`, although the PR describes loading all YAML files. Is `.yml` intentionally unsupported? If not, this should enumerate both extensions and return a specific error when no matching files are present.

pkg/config/config.go
```diff
@@ -143,10 +160,49 @@ func (c *Config) loadFile(flags *Flags) error {
...
+	}
+
+	for _, file := range files {
+		path := flags.ConfigPath + "/" + file
+		data, err := os.ReadFile(path)
+		if err != nil {
+			return err
+		}
+
+		if err = c.loadYAMLConfig(string(data)); err != nil {
```
houyuwushang: Repeated unmarshalling keeps distinct map keys, but a later file replaces an entire same-named backend. In a local case where the second fragment only overrode `origin_url`, the loaded backend lost its `provider`. Since the PR says files are merged in order, this needs an explicit contract: deep-merge fields, reject duplicate resource names, or document that a repeated resource must be complete.

pkg/config/config.go
Elia-Renzoni: Thanks for pointing this out. This is indeed a bug and not intentional. I will push a commit as soon as possible to support both extensions and return a specific error when no matching files are found.

pkg/config/config.go
Elia-Renzoni: Thanks for reporting this. You are absolutely right I verified the code and confirmed the behavior you described.

The `loadAndMergeFiles` function in `pkg/config/config.go` loads the YAML files sequentially using `yaml.Unmarshal(&c)`, which completely overwrites the configuration structure on each iteration instead of performing a true merge. This means that if the same backend is defined across multiple files with partial fields, the fields from the earlier files are lost.

I agree that we need an explicit contract and among the options you suggested I propose implementing an explicit rejection.

houyuwushang:
I've pushed the collaborative changes to `config-493`. The current branch keeps `-config <directory>` and also supports a primary file with a sibling `conf.d` or configured include directory, with the review fixes folded in.

I don't have permission to edit the PR description. Could you update it to reflect the current scope and use the current template, including the "DOES include AI-generated changes" disclosure for the commits I added?

pkg/config/config.go
```diff
@@ -143,18 +149,30 @@ func NewConfig() *Config {
...
 	}
-	err = c.loadYAMLConfig(string(b))
-	if err != nil {
+	configData := sources[0].data
+	if plan.mode == configSourceModeDirectory || len(sources) > 1 {
+		configData, err = mergeConfigSources(plan, sources)
+		if err != nil {
+			return err
+		}
+	}
```
jranson, line 158: Since `parseConfigDocument` (which contains many strict checks) is only called by `mergeConfigSources`, and `mergeConfigSources` is only called when there are multiple sources, it means that single-file configs are not parsed against the strict checks in the same way that multiple-file configs are. It would be good to have an `else` block here that calls `parseConfigDocument` on `source[0]` and checks the error.

pkg/config/config.go
```diff
@@ -100,16 +100,22 @@ type MainConfig struct {
...
+	configSourceFingerprint string
+	configLastModified      time.Time
+	configRateLimitTime     time.Time
+	stalenessCheckLock      sync.Mutex
 }
 
 func (mc *MainConfig) SetStalenessInfo(fp string, lm, rlt time.Time) {
 	mc.stalenessCheckLock.Lock()
 	mc.configFilePath = fp
+	mc.configSourcePlan = configSourcePlan{}
```
jranson: This completely wipes source plan, which clears the fingerprint and silently downgrades a multi-source config to single-path / os.Stat semantics. For directory mode that means statting the directory, which won't notice edits to files inside it. 

It looks like `SetStalenessInfo` isn't being called from anywhere currently, so it might be fine to just remove this function altogether rather than fix.

pkg/config/sources.go
```diff
@@ -0,0 +1,406 @@
...
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"slices"
+	"strings"
+	"time"
+
+	yamlv3 "go.yaml.in/yaml/v3"
+	yamlv2 "gopkg.in/yaml.v2"
```
jranson: rather than straddle 2 version imports, we should update the pre-existing v2 imports app-wide to v3 so that v2 is no longer present in the codebase.

jranson (changes requested):
@Elia-Renzoni and @houyuwushang just a few changes noted in-line below, as well as this item:

* it looks like when multiple files are loaded, the pre-existing info log event in setup.go only logs the base filename. It would be good to print all files loaded (in the order of concatenation/application).

pkg/config/config.go
houyuwushang: Single-file loads now call the same strict document parser before decoding, so duplicate keys, multiple documents, non-mapping roots, and alias cycles are rejected consistently.

pkg/config/config.go
houyuwushang: Removed SetStalenessInfo and its test-only callers. The reload tests now mutate the backing files and exercise the source fingerprint path directly.

pkg/config/sources.go
houyuwushang: Migrated the remaining direct YAML v2 imports in the root and integration modules to go.yaml.in/yaml/v3. Listener overrides now decode from yaml.Node, and the shared encoder keeps the existing two-space, compact-sequence output.

houyuwushang:
The startup log now uses the exact source list captured during loading, joined in application order under the existing config field. Single-file logging remains unchanged.

jranson:
The failing integration test is below. It may be related to the yaml v2 -> v3 refactor.

```
=== RUN   TestALB
time=2026-08-21T04:46:03.264814172Z app=trickster level=error event="http listener stopping" caller=pkg/proxy/listener/listener.go:290 detail="http: Server closed" listenerName=listener.metrics.http
time=2026-08-21T04:46:03.264869355Z app=trickster level=error event="http listener stopping" caller=pkg/proxy/listener/listener.go:290 detail="http: Server closed" listenerName=listener.default.http
time=2026-08-21T04:46:03.26495487Z app=trickster level=error event="http listener stopping" caller=pkg/proxy/listener/listener.go:290 detail="http: Server closed" listenerName=listener.mgmt.http
    main_test.go:64: 
        	Error Trace:	/home/runner/work/trickster/trickster/integration/main_test.go:64
        	            				/opt/hostedtoolcache/go/1.26.6/x64/src/runtime/asm_amd64.s:1771
        	Error:      	Received unexpected error:
        	            	yaml: unmarshal errors:
        	            	  line 17: cannot unmarshal !!int `3` into time.Duration
        	            	  line 18: cannot unmarshal !!int `3` into time.Duration
        	            	  line 19: cannot unmarshal !!int `3` into time.Duration
        	Test:       	TestALB
time=2026-08-21T04:46:03.266011291Z app=trickster level=error event="Could not load configuration:" caller=pkg/daemon/setup/setup.go:102 error="yaml: unmarshal errors:
  line 17: cannot unmarshal !!int `3` into time.Duration
  line 18: cannot unmarshal !!int `3` into time.Duration
  line 19: cannot unmarshal !!int `3` into time.Duration"
time=2026-08-21T04:46:03.286530575Z app=trickster level=error event="healthcheck failed" caller=pkg/backends/healthcheck/target.go:369 error="Get \"http://127.0.0.1:45709/api/v1/status/buildinfo?query=up\": dial tcp 127.0.0.1:45709: connect: connection refused" targetName=prom-b
time=2026-08-21T04:46:03.33821158Z app=trickster level=error event="healthcheck failed" caller=pkg/backends/healthcheck/target.go:369 error="Get \"http://127.0.0.1:43711/api/v1/status/buildinfo?query=up\": dial tcp 127.0.0.1:43711: connect: connection refused" targetName=prom-a
    alb_test.go:45: 
        	Error Trace:	/home/runner/work/trickster/trickster/integration/main_test.go:105
        	            				/opt/hostedtoolcache/go/1.26.6/x64/src/runtime/asm_amd64.s:1771
        	Error:      	Received unexpected error:
        	            	Get "http://127.0.0.1:8491/metrics": dial tcp 127.0.0.1:8491: connect: connection refused
    alb_test.go:45: 
        	Error Trace:	/home/runner/work/trickster/trickster/integration/main_test.go:103
        	            				/home/runner/work/trickster/trickster/integration/alb_test.go:45
        	Error:      	Condition never satisfied
        	Test:       	TestALB
        	Messages:   	endpoint did not become ready: http://127.0.0.1:8491/metrics
--- FAIL: TestALB (10.00s)
FAIL
```

houyuwushang:
Yes, it was. Those integration fixtures still used bare 3 values for negative-cache TTLs. YAML v2 decoded them as 3ns, while v3 rejects them. I changed both fixtures to explicit 3s in 14581559, matching the documented format; the integration rerun completed.

jranson (approved):

PR #913, opened 2026-02-16
chore: Bump Go version to 1.26

crandles:
- Bumping the Go version used by CI/Dockerfile/go.mod.
- Move `golangci-lint` to be installed via `go get -tool`, given it must be rebuilt for each Go version
  - Developers must run `make get-tools` to install updated linter.

---

These notes looks relevant to our use cases:

> Runtime[¶](https://go.dev/doc/go1.26#runtime)
New garbage collector[¶](https://go.dev/doc/go1.26#new-garbage-collector)
The Green Tea garbage collector, previously available as an experiment in Go 1.25, is now enabled by default after incorporating feedback.
>
> This garbage collector’s design improves the performance of marking and scanning small objects through better locality and CPU scalability. Benchmark results vary, but we expect somewhere between a 10–40% reduction in garbage collection overhead in real-world programs that heavily use the garbage collector. Further improvements, on the order of 10% in garbage collection overhead, are expected when running on newer amd64-based CPU platforms (Intel Ice Lake or AMD Zen 4 and newer), as the garbage collector now leverages vector instructions for scanning small objects when possible.
>
> The new garbage collector may be disabled by setting GOEXPERIMENT=nogreenteagc at build time. This opt-out setting is expected to be removed in Go 1.27. If you disable the new garbage collector for any reason related to its performance or behavior, please [file an issue](https://go.dev/issue/new).
>
> Compiler[¶](https://go.dev/doc/go1.26#compiler)
The compiler can now allocate the backing store for slices on the stack in more situations, which improves performance. If this change is causing trouble, the [bisect tool](https://pkg.go.dev/golang.org/x/tools/cmd/bisect) can be used to find the allocation causing trouble using the -compile=variablemake flag. All such new stack allocations can also be turned off using -gcflags=all=-d=variablemakehash=n. If you encounter issues with this optimization, please [file an issue](https://go.dev/issue/new).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -278,7 +278,7 @@ get-tools: get-msgpack
 
 .PHONY: get-msgpack
 get-msgpack:
-	$(GO) get -tool github.com/tinylib/msgp@v1.6.3
+	$(GO) get -tool github.com/tinylib/msgp@$(shell go list -m github.com/tinylib/msgp | cut -d' ' -f2)
 
 .PHONY: developer-start
 developer-start:


jranson (changes requested):
@crandles i think this is missing the update to 1.26 in `go.mod`

PR #925, opened 2026-02-23
refactor: [caches/memory] drop "referenced object" support 

crandles:
## Changes:
- refactor: [caches/memory] drop "referenced object" support
  -  no noticeable performance differences; this optimization did not seem to be worth the complexity it caused within the overall codebase
  - calling code should be simpler now; should make it simpler to optimize the overall codebase (no more provider specific code paths)
- chore: [tests] use t.Context() over context.Background()
- - chore: [docs/dashboards] add redis datasource to grafana
- chore: [docs/dashboards] add more memory & request duration panels
  - added: "GOMAXPROCS" stat, 2 extra memory stats, cpu usage, go gc duration, processed bytes, frontend request duration histogram, + overall request duration histogram
  - (processed bytes not working via darwin at this time) 

---
new panels:
Screenshot 2026-02-23 at 6 52 00 PM
more new panels:

Screenshot 2026-02-23 at 6 52 11 PM

### DIFF
--- a/docs/chunked_caching.md
+++ b/docs/chunked_caching.md
@@ -16,7 +16,6 @@ Chunked caching can be enabled and disabled using `use_cache_chunking`:
 
 ```yaml
 fs1:
-    cache_type: filesystem
     provider: filesystem
     use_cache_chunking: true
     timeseries_chunk_factor: 420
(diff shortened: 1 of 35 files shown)


crandles:
good call, seeing some difference:
image

though I should increase my durations & ensure this load test is acting correctly

jranson:
given the cpu findings here + drift from `main`, i'm going to close this for now. we can re-open or resubmit against current branch if needed.

PR #979, opened 2026-04-15
ci+fix: integration coverage, quality infrastructure, bug fixes across prometheus/clickhouse/merge layers

crandles:
### Infrastructure
- `make integration-cover` + separate coveralls flag for integration coverage (~43%).
- `make fuzz` target runs all `Fuzz*` functions (default 30s each, `FUZZ_TIME=5m make fuzz`).
- `make lint` now includes `govulncheck` via `go tool govulncheck`.
- CI: govulncheck step (non-blocking) + 50% coverage floor enforcement.
- `integration/go.mod` — separate module so SDK test deps don't pollute root.
- `waitForPrometheusData` waits for step-aligned range data to eliminate DPC-truncation flakiness.
- Fix `getCaller` matching Go 1.26 toolchain `/pkg/mod/` paths.

### JSON escaping
- Fix `Tags.JSON()` — used string concat, now `encoding/json`.
- Fix `Envelope.StartMarshal` — Status/Error/ErrorType/Warnings all unescaped.
- Fix timeseries marshaler writing tag keys/values raw — now delegates to `Tags.JSON()`.
- Fix `MergeAndWriteLabelDataRespondFunc` using `strings.Join` without escaping.
- Fix `MergeAndWriteAlertsRespondFunc` separator between objects.
- Fix ClickHouse `WFDataItem.MarshalJSON` using `"%s":"%s"`.

### Metadata merge
- Fix `WFSeriesData` dropping arbitrary labels — now `map[string]string`.
- Fix `/series` + `/labels` rounding `end` down, dropping up to 59s of newest samples.
- Fix `/labels` + `/label/<name>/values` not surfacing `prometheus.labels`-injected keys/values.

### Data integrity
- Fix `pointFromValues` truncating sub-second epoch precision. Same fix in testutil.
- Fix `populateSeries` zero-epoch holes from malformed points — panicked on marshal.
- Fix `MakeMergeFunc` / `TimeseriesMergeFunc` silently returning nil on unexpected type.
- Fix `UnmarshalTimeseriesReader` nil-pointer panic on `{}` input.
- Fix `CalculateHash` and `SortByTags` separator collisions with UTF-8 label names (Prom 3.0).
- Fix ~14 ignored errors in cache write, request clone, serialization, and ALB mechanism paths.
- Fix ClickHouse `marshalTimeseriesJSON` swallowing `toWireFormat` errors.

### ClickHouse
- Fix `CacheKeyParams` missing `query` — different SQL statements shared the same OPC cache key.
- Add `/ping` route for SDK HTTP protocol compatibility.
- Propagate `trq` with `CacheKeyElements` on parse failure — non-time-range queries (LIMIT, `SELECT 1`, SDK handshake) now get per-query OPC cache keys instead of all sharing one entry.
- ClickHouse Go SDK (`clickhouse-go/v2`) Native binary protocol now works through trickster via OPC transparent proxy.

### Test quality
- 6 fuzz targets across Prometheus model, Tags, ClickHouse — found 2 additional bugs.
- Round-trip value assertions in 6 companion `roundtrip_test.go` files (15 tests).
- Body-parse assertions added to labels and alerts merge tests.
- SDK-based black-box integration tests: Prometheus (`client_golang/api/v1`), ClickHouse (Native SDK + HTTP/JSON), InfluxDB (`influxdb-client-go/v2`).

## Type of Change

- [x] Bug fix
- [x] Test coverage
- [x] Infrastructure

## AI Disclosure


pkg/backends/clickhouse/model/marshal_json.go
```diff
@@ -37,7 +37,11 @@ func (d WFDataItem) MarshalJSON() ([]byte, error) {
 		if sep {
 			buf.Write([]byte{','})
 		}
-		fmt.Fprintf(buf, `"%s":"%s"`, e.Key, e.Value)
+		kb, _ := json.Marshal(e.Key)
```
crandles, line 38: open question: do we extend this to support utf-8, or go with the upstream implementation? (likewise for similar following instances). At this micro level (a single field), I suspect we're still performant.

crandles:
last addition -- using client sdks in the integration tests, once i get the tests sorted out i'll call this done

seems clickhouse prefers binary, so that won't be so simple

pkg/backends/clickhouse/model/marshal_json.go
jranson: I think as-is is fine, and we can revisit if unicode is needed/required in a separate patch.

jranson (approved):

PR #981, opened 2026-04-18
feat: [clickhouse] native binary protocol + parser + grafana v4 plugin

crandles:
## Description

- ClickHouse native binary protocol both directions: frontend `protocol_listeners` (raw TCP, extensible to MySQL) and backend `protocol: native` upstream via clickhouse-go
- Native DPC modeler: detects Native via `X-ClickHouse-Format`, deserializes for caching and delta-merging, re-serializes for response — no client config needed
- `Fetcher` abstraction in `PrepareFetchReader` opens non-HTTP upstreams more generally
- Wire protocol coverage: handshake + addendum, query, data blocks, ping/pong, LZ4. INSERT-as-SQL works; inline INSERT data blocks not yet supported
- Parser: 17 `toStartOf*` functions, `date_trunc`, `timeSlot`, `toStartOfInterval` with all units; `toDateTime64(value, precision)` and `now64()` in WHERE; Date-type interpolation for `toStartOfMonth/Quarter/Year` (Grafana v4+ shapes)
- DPC→OPC fallback warns at warn level, togglable via `dpc_fallback_warning`; structural non-TS queries (health checks, version) stay at debug
- Dev env: both Vertamedia and official Grafana plugins installed, dashboard for each
- CI: skip CodeQL on forks

## Type of Change

- - [x] New feature
- - [x] Test coverage
- - [x] Documentation
- - [x] Infrastructure

## AI Disclosure

- - [x] This contribution DOES include AI-generated changes, and I have reviewed the relevant contributing guidelines.

PR #985, opened 2026-04-18
feat: [influxdb] v3 support — http + flight sql + flux fixes

### LINKED ISSUE #984: add InfluxDB 3.x support
InfluxDB 3.x (GA April 2025, now at 3.9.1) is a ground-up rewrite -- Flux is gone, SQL is the primary query language, and there are new `/api/v3/` endpoints for queries and writes. Trickster's existing InfluxQL support works against v3 via compat endpoints, but there's no caching for SQL queries or the native v3 API.

Main gaps: SQL query caching (`/api/v3/query_sql`), v3 InfluxQL endpoint (`/api/v3/query_influxql`), and new response formats (`jsonl`, `csv`). Can model the SQL time-range parsing after the clickhouse backend.

## Flight SQL (gRPC)

v3 also exposes Apache Arrow Flight SQL on gRPC. Grafana's InfluxDB datasource in SQL mode uses Flight SQL, not HTTP -- so without this, SQL panels in Grafana can't go through Trickster. Other high-perf v3 clients (Python/Rust/Java SDKs) also default to Flight SQL.

To support this we need:

- gRPC listener (extend the non-HTTP protocol listener pattern from #981, or a new gRPC-aware listener variant -- `grpc.Server.Serve(lis)` owns the whole listener, unlike raw-TCP per-conn handling)
- Flight SQL server implementation -- handle `CommandStatementQuery` + `DoGet`
- Apache Arrow Go (`github.com/apache/arrow-go/v18`) for Arrow RecordBatch serialization
- Reuse existing SQL parser + delta-proxy cache by querying upstream over HTTP SQL (`/api/v3/query_sql?format=json`), converting to Arrow on the way out. Upstream-via-Flight-SQL is a later optimization.
- Cache interop: `dataset.DataSet` round-trips to Arrow record batches

Tracking as part of this issue since the goal is the same (cache v3 SQL). Scope may split into separate PRs.

crandles:
## Description

Closes #984.

- v3 HTTP SQL + InfluxQL via `/api/v3/*`, json/jsonl/csv response formats, `/api/v3/write_lp` passthrough
- Flight SQL (gRPC) proxy: per-backend `flight_port`, IPC-byte cache, header forwarding, daemon shutdown + SIGHUP reload
- Flight SQL metadata RPCs (Tables, Catalogs, Schemas, TableTypes, SqlInfo) and prepared-statement lifecycle (Create/Execute/Close/DoPut), cache key includes bound-param hash
- SQL parser handles both `date_bin()` and `date_trunc()` bucketing
- Flux: `now()` in `range()`, balanced-paren tokenization for nested calls, multi-table CSV reads — was silently falling through to HTTPProxy on common Grafana shapes
- Dev env: `influxdb3` compose service + side-by-side dashboard exercising metadata RPCs; `verify-influxdb3.sh` for direct HTTP SQL

InfluxDB 3 Core 3.10 returns a parameter schema at prepare time but doesn't resolve bound values during planning ("No value found for placeholder"). Parameterized prepared statements fail at the upstream, not the proxy; integration test for that path omitted pending upstream fix.

Follow-ups:

- [x] refactor Flight listener to use #981's `ProtocolListener` once that merges

## Type of Change

- - [x] Bug fix
- - [x] New feature
- - [x] Test coverage
- - [x] Documentation

## AI Disclosure


jranson:
Finishing this out by swapping out the deprecated/removed internal SQL parser for cockroach's parser (it's compatible with Influx 3.x SQL and can be re-used for Druid SQL), refactoring Arrow Flight SQL to use the listener pattern from `main`, and finishing out DPC support for Flight SQL.  For DPC on binary/non-http, added a `nativedelta` package that is now used by both mysql and Flight SQL.

jranson (approved):

PR #986, opened 2026-04-20
fix: [proxy/engines] reset cacheStatus on DPC empty-body recovery; pin chunk test clock

crandles:
## Description

`TestDeltaProxyCacheRequestRangeMissChunks` flakes during ~2-hour UTC windows per day. Cause: when `QueryCache` returns a meta-level `Hit` but chunk reads all miss, the "empty document body" recovery path re-fetches but forgets to reset `cacheStatus`, so `status=hit` gets reported for requests that actually round-tripped to origin.

- Reset `cacheStatus = LookupStatusKeyMiss` after recovery fetch
- Pin existing test clock to a same-bucket date
- Add `_CrossBucket` regression test with a clock that triggers the bug

-- Verified empirically that the subsequent `cache.Remove(key)` call is benign: chunks are keyed by deterministic subkey (not via the meta), and the next write immediately rewrites the meta marker. A follow-up query for a different range in a still-cached bucket remains a hit.

## Type of Change

- - [x] Bug fix
- - [x] Test coverage

## AI Disclosure

### DIFF
--- a/pkg/proxy/engines/deltaproxycache.go
+++ b/pkg/proxy/engines/deltaproxycache.go
@@ -316,6 +316,8 @@ func DeltaProxyCacheRequest(w http.ResponseWriter, r *http.Request, modeler *tim
 					if err != nil {
 						return buildErrorResult(doc.StatusCode, doc.SafeHeaderClone(), doc.Body), nil
 					}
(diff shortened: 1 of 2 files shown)


jranson (approved):

jranson:
(still approved!)

Issue #291, opened 2019-10-01
better handling on failed upstream partial requests


jranson:
when the delta proxy cache determines which ranges to request from the origin, if those ranges are separated by any amount of time, the origin will make multiple parallel requests to get the separate ranges. If one of those requests fails, the result to the client can be unpredictable based on which request finished first and what its response was, and the order that the were reconstituted. we should identify the proper way to handle these cases and implement the best practice. At a minimum, I would like to report back to the end client via a response header if a part of the content failed.

u5surf:
For example, if `getMatrixFromPrometheus` returns err, what you wanna said, it does `r.Writer.Header().Set(“foo”,  originParams.Encode())`, don’t you?
https://github.com/Comcast/trickster/blob/67cf188025f01e4e516ac032ec43d62760d52ad7/handlers.go#L750-L757

jranson:
Fixed by @Elia-Renzoni in #989  ❤️

PR #989, opened 2026-04-30
fix(deltaproxycache): better handling of failed requests

Elia-Renzoni:
## Description

This update improves how DeltaProxyCache handles and reports failures when serving partial hit requests that involve multiple upstream extents.  With this change failed extents are explicitly tracked during the fetchExtents() process using a dedicated parallel structure,rather than aggregating errors. This allows the system to preserve detailed information about which specific extents failed.

fetchExtents() now returns (ExtentList, bool) to indicate failed extents and whether a full failure occurred. A full failure (or severe fault) occurs when all the requests fail, in this case the severeFault flag is set to true, otherwise it is false.

Related Issue: #291 
## Type of Change

- - [ ] Bug fix
- - [ ] New feature
- - [x] Optimization
- - [ ] Test coverage
- - [ ] Documentation
- - [ ] Infrastructure

## AI Disclosure

- - [x] This contribution DOES include AI-generated changes, and I have reviewed the relevant contributing guidelines. (the unit test is AI generated)


pkg/proxy/engines/deltaproxycache.go
```diff
@@ -784,5 +803,27 @@ func fetchExtents(el timeseries.ExtentList, rsc *request.Resources, h http.Heade
...
+func trimEmptyExtents(failedExtents timeseries.ExtentList) timeseries.ExtentList {
+	var trimmedList timeseries.ExtentList
+	var emptyExtent = timeseries.Extent{}
+
+	for _, extent := range failedExtents {
+		if extent == emptyExtent {
+			continue
+		}
+
+		trimmedList = append(trimmedList, extent)
```
crandles: we tend to prefer over-allocating/trimming slices vs using append in this project, can we do so here?

example: https://github.com/trickstercache/trickster/blob/275e8f1354ae654d88f1eec8fe261e2efd810326/pkg/util/strings/strings.go#L119-L121

pkg/proxy/engines/deltaproxycache.go
```diff
@@ -748,7 +767,7 @@ func fetchExtents(el timeseries.ExtentList, rsc *request.Resources, h http.Heade
 				appendLock.Unlock()
 				mts[i] = nts
 			} else if resp.StatusCode != http.StatusOK {
-				errs[i] = tpe.ErrUnexpectedUpstreamResponse
+				errTs[i] = el[i]
```
crandles, line 804: when are we using these error values? (it doesn't look like we are, is that intended? would expect we log these somewhere). previously we were returning `errors.Join(errs...)`

pkg/proxy/headers/result_header.go
crandles: think this is a bug? failed is writing to fetched? but should write to FailedFetched?

pkg/proxy/engines/deltaproxycache_test.go
```diff
@@ -565,6 +565,111 @@ func TestDeltaProxyCacheRequestPartialHit(t *testing.T) {
 	}
 }
 
+// TestDeltaProxyCacheRequestPartialHitWithFailedExtents verifies that when
+// a partial hit occurs and the upstream request for the missing fragment fails,
+// the failed extents are properly tracked and reported in the response header.
+func TestDeltaProxyCacheRequestPartialHitWithFailedExtents(t *testing.T) {
```
crandles, line 613: i think we need more tests cases

  - multi-extent partial failure (some succeed, some fail) 
  - header round-trip parse test for `failed=`
  - MergeResultHeaderVals with a `failed=` header on either side
  - sharded fetchTimeseries with a single shard failing

crandles: here are some failing / panic'ing tests https://github.com/trickstercache/trickster/commit/141a37495eb400feaef277516b79a5beba239ae6

pkg/proxy/engines/deltaproxycache.go
Elia-Renzoni: I removed the use of errors in favor of returning the failed extents in the headers. The errors are still being logged, in particular UnexpectedUpstreamResponse and the unmarshaling errors. The only case that is not currently logged is the error coming from the Fetch function call.

Elia-Renzoni: The error from the Fetch function is also logged within the function itself, e.g.:
```go
if err != nil {
		logger.Error("error reading body from http response",
			logging.Pairs{"url": pr.URL.String(), "detail": err.Error()})
		return body, resp, 0, err
}
```

jranson (approved):
@Elia-Renzoni thanks for this contribution, and sorry it took us a while to get back to it. A lot of other focused PRs came in affecting these same files, so we had some merge conflicts to work out. This looks good now and I'll go ahead and merge it. Thanks for your patience.

PR #998, opened 2026-05-13
fix: [alb/tsm] prom avg outer aggregation with POST

jranson:
## Description

This corrects a defect in ALB Time Series Merge for Prometheus on `avg` outer aggregators. To avoid an improperly weighted average, we clone the original request twice for the proxy call, replacing one's `avg` with `sum` and the other's with `count`, and then perform the average internally per-timestamp against the results.

The defect surfaces when the request is POST. The cloned requests were pointing to a shallow copy / pointer of a context item, causing both requests to issue `count` requests to the backends, causing the average value to always be `1` for every timestamp.

This updates `prometheus.RewriteForWeightedAvg` to deep copy the resources context, and `params.SetRequestValues` to update the deep-copied resource's cached body to stay in sync with the outbound body.

Tests added for both.

## Type of Change

- - [x] Bug fix
- - [x] Test coverage

## AI Disclosure

### DIFF
--- a/pkg/backends/prometheus/tsm_provider.go
+++ b/pkg/backends/prometheus/tsm_provider.go
@@ -17,10 +17,14 @@
 package prometheus
 
 import (
+	"maps"
 	"net/http"
 
 	"github.com/trickstercache/trickster/v2/pkg/backends/prometheus/promql"
+	"github.com/trickstercache/trickster/v2/pkg/observability/logging"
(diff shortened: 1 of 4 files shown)


jranson:
still having an issue with `1` so drafting this until i get it fully resolved

crandles (approved):

PR #1001, opened 2026-05-13
fix+refactor: [alb] hardening sweep + shared fanout primitive

crandles:
## Description

ALB hardening, continued. Before this PR, the following could silently go wrong:

- TSM client disconnects didn't cancel upstreams
- NLM rejected the IMF-fixdate "GMT" `Last-Modified` format real servers emit
- UR options raced against SIGHUP reload
- TSM error aggregation surfaced min status when all members failed (one 400 could hide 502s)
- FR and NLM raced on the POST request body during fanout clone
- One oversized upstream could OOM the proxy; capped via `max_capture_bytes` (default 256 MiB)
- `UnmarshalTimeseriesReader` panicked on a nil reader

Adds `trickster_alb_fanout_failures_total{mechanism,reason}`.

Also extracts `pkg/backends/alb/mech/fanout`.

```
code:   +593 -278
test:   +1693 -4
```

Not seeing anything obvious left to test at this point, going to call it done for now.

## Type of Change

- - [x] Bug fix
- - [x] New feature
- - [x] Test coverage

## AI Disclosure

### DIFF
--- a/pkg/backends/alb/mech/fanout/fanout.go
+++ b/pkg/backends/alb/mech/fanout/fanout.go
@@ -30,6 +30,7 @@ import (
 	"net/http"
 
 	"github.com/trickstercache/trickster/v2/pkg/backends/alb/mech"
+	"github.com/trickstercache/trickster/v2/pkg/backends/alb/mech/types"
 	"github.com/trickstercache/trickster/v2/pkg/backends/alb/pool"
 	"github.com/trickstercache/trickster/v2/pkg/observability/logging"
 	"github.com/trickstercache/trickster/v2/pkg/observability/logging/logger"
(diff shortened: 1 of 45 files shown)


pkg/backends/alb/mech/fanout/fanout.go
```diff
@@ -0,0 +1,211 @@
...
+	// is reflected only in Failed; the panic value is logged + metered
+	// inside the fanout goroutine.
+	Err error
+}
+
+// Config configures one fanout call.
+type Config struct {
+	// Mechanism is the short name ("fr", "nlm", "tsm", "tsm/avg-sum", ...)
+	// used in panic recovery logs and the fanout_failures_total metric.
+	Mechanism string
```
jranson: should we use mech/types.Name here?

jranson (approved):

pkg/backends/alb/mech/fanout/fanout.go
crandles: good call, done

PR #1006, opened 2026-05-17
fix+refactor+test: [alb/cache/proxy] continued hardening sweep

crandles:
## Description

Another pass at ALB stability, started as tests/fixes and pulled in some refactoring along the way.

Adds opt-in `dedup_tolerance_ms` (TSM sub-step skew tolerance) and `max_fanout_capture_bytes` (aggregate in-flight cap, documented in `docs/alb.md`).

Adds [goleak](https://github.com/uber-go/goleak) and chaos testing helpers.

## Breaking / observable changes

- `pool.Pool`: `Healthy`/`HealthyTargets` removed; `LiveTargets` -> `Targets`
- `Mechanism` split into `Mechanism` + `PoolMechanism`; `types.ID` removed
- `X-Trickster-Result` `fetched=[...]` separator flipped `,` -> `;`
- cyclic ALB / `user_router` configs rejected at load
- truncated upstreams (bytes < declared Content-Length) disqualified in ALB fanout
- TSM `Set-Cookie` is now winner-only (cross-tenant leak fix)
- TSM returns 502 when no fanout member produces a merge contribution (was empty 200 from a synthetic harness)
- `user_router` skips the Basic-auth rewrite when both the inbound credential and `ToCredential` are empty (was: emitted Basic auth with empty password, collapsing SSO users into one cache key)
- `fanout.WaitForFirst` returns on winner-claim without draining losers (was: drained every spawned goroutine before return). Trickster's HTTP client propagates ctx-cancel through transport, so in prod the orphan scatter goroutines drain at transport-timeout speed
- `pkg/testutil/albpool.ServeAndWait` re-raises recovered panics via `t.Fatalf` (was `t.Errorf`); callers can no longer proceed to assert against a zero-valued ResponseRecorder

## Observability

New panic-recovered counters across healthcheck / pool / engines / cache index. New `trickster_alb_fanout_attempts_total` and `trickster_healthcheck_probe_latency_seconds`. `trickster_alb_fanout_failures_total` gains a `reason="short_read"` label.

## Type of Change

- - [x] Bug fix
- - [x] New feature
- - [x] Optimization
- - [x] Test coverage

## AI Disclosure

---

## Notes

Apologies for the gargantuan PR, this is a few days worth of writing failing tests / fixes. We could attempt to break this up into smaller PRs, but I believe all of the failures are valid / require fixes. 

Most of this should be tests (I tried to cut those down, but it was difficult to recreate the many failure scenarios that found bugs), with some light refactoring.

```
code:   +2791 -533
test:   +8315 -2065
docs:   +21 -0
config: +900 -0
other:  +26 -2
```

See https://github.com/trickstercache/trickster/pull/1005 for the 30 commit version of this, which i've squashed here.


crandles:
I think this is about as far as I can take it with the current unit and integration tests, will be looking into more of an e2e harness for further ALB errors

jranson (approved):

PR #1019, opened 2026-05-29
fix: [listener] correct active-connections gauge under connections_limit

thinker0:
## Description
Keeps the `trickster_proxy_active_connections` gauge accurate when `frontend.connections_limit` is set.

**Root cause**
`Listener.Accept()` only wrapped `*net.TCPConn` in `observedConnection`. When `connections_limit > 0` the listener is wrapped by `netutil.LimitListener`, whose accepted conns are `*netutil.limitListenerConn` — not `*net.TCPConn`. Those conns were returned unwrapped, so their `Close()` never decremented `ProxyActiveConnections`: the gauge climbed monotonically and stopped reflecting the real connection count.

A related issue: `net/http` can call `Close()` on the same conn more than once, which previously double-decremented the gauge and drifted it negative under load.

**Fix**
- `Accept()` now wraps every accepted conn except `*tls.Conn` (left bare so `http.Server` can negotiate HTTP/2 via ALPN). `observedConnection` embeds `net.Conn` instead of `*net.TCPConn`.
- `observedConnection.Close()` decrements the gauge only on the first successful close; a repeat close returns an error, which guards the decrement — replacing the previous `sync.Once` per @crandles's review.

**Tests**
- `TestObservedConnectionIdempotentClose` asserts a repeated close returns an error and the gauge nets a single decrement.
- `TestAcceptWrapsLimitListenerConn` asserts `Accept` wraps a `LimitListener` conn.

## Type of Change
- [x] Bug fix
- [x] Test coverage

## AI Disclosure

### DIFF
--- a/pkg/proxy/listener/listener.go
+++ b/pkg/proxy/listener/listener.go
@@ -66,13 +66,17 @@ type Listener struct {
 }
 
 type observedConnection struct {
-	*net.TCPConn
+	net.Conn
 }
 
 func (o *observedConnection) Close() error {
-	err := o.TCPConn.Close()
-	metrics.ProxyActiveConnections.Dec()
-	metrics.ProxyConnectionClosed.Inc()
+	err := o.Conn.Close()
+	// Only the first successful Close adjusts the gauge; a subsequent Close
+	// returns an error (net/http may close the same conn more than once).
+	if err == nil {
+		metrics.ProxyActiveConnections.Dec()
+		metrics.ProxyConnectionClosed.Inc()
+	}
 	return err
 }
 
@@ -88,12 +92,15 @@ func (l *Listener) Accept() (net.Conn, error) {
 
 	metrics.ProxyActiveConnections.Inc()
 	metrics.ProxyConnectionAccepted.Inc()
-	// this is necessary for HTTP/2 to work
-	if t, ok := c.(*net.TCPConn); ok {
-		return &observedConnection{t}, nil
+	// *tls.Conn is left unwrapped so http.Server can type-assert it for
+	// HTTP/2 (ALPN); every other conn -- including *net.TCPConn and the
+	// netutil.LimitListener wrapper -- is wrapped so Close decrements
+	// ProxyActiveConnections.
+	if _, ok := c.(*tls.Conn); ok {
+		return c, nil
 	}
 
-	return c, nil
+	return &observedConnection{Conn: c}, nil
 }
 
 // CertSwapper returns the CertSwapper reference from the Listener
--- a/pkg/proxy/listener/listener.go
+++ b/pkg/proxy/listener/listener.go
@@ -70,14 +70,14 @@ type observedConnection struct {
 }
 
 func (o *observedConnection) Close() error {
-	err := o.Conn.Close()
+	if err := o.Conn.Close(); err != nil {
+		return err
+	}
 	// Only the first successful Close adjusts the gauge; a subsequent Close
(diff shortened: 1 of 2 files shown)


pkg/proxy/listener/listener.go
```diff
@@ -67,12 +67,15 @@ type Listener struct {
...
 }
 
 func (o *observedConnection) Close() error {
 	err := o.TCPConn.Close()
-	metrics.ProxyActiveConnections.Dec()
-	metrics.ProxyConnectionClosed.Inc()
+	o.once.Do(func() {
+		metrics.ProxyActiveConnections.Dec()
+		metrics.ProxyConnectionClosed.Inc()
+	})
```
crandles: a second close should return an err
```suggestion
	if err == nil {
		metrics.ProxyActiveConnections.Dec()
		metrics.ProxyConnectionClosed.Inc()
	}
```

I think it would be simpler to skip sync.Once here if we add an err check

pkg/proxy/listener/listener_test.go
```diff
@@ -354,3 +356,40 @@ func TestCloseObservedConnection(t *testing.T) {
...
+	metrics.ProxyActiveConnections.Set(10.0)
+
+	// First Close should succeed and decrement metric
+	if err := oconn.Close(); err != nil {
+		t.Error(err)
+	}
+
+	// Subsequent Closes should also succeed but NOT decrement metric further
+	for i := 0; i < 5; i++ {
+		_ = oconn.Close()
```
crandles: you're not checking for success here, unlike what the comment says

pkg/proxy/listener/listener.go
thinker0: Done — dropped `sync.Once` and used the `if err == nil` guard you suggested. `observedConnection` now embeds `net.Conn`, and only the first successful `Close` decrements the gauge; a repeat `Close` returns an error, so it's skipped. The same embed-`net.Conn` change also lets `Accept` wrap the `netutil.LimitListener` conn — that was the actual source of the gauge running away under `connections_limit`, now folded into this PR (see the updated description). (6370fa69)

pkg/proxy/listener/listener_test.go
thinker0: Fixed — the loop now asserts each subsequent `Close` returns an error (that error is exactly what guards the single decrement), so the test matches its comment. Also added `TestAcceptWrapsLimitListenerConn` to cover the LimitListener path. (6370fa69)

pkg/proxy/listener/listener.go
```diff
@@ -66,13 +66,17 @@ type Listener struct {
...
 type observedConnection struct {
-	*net.TCPConn
+	net.Conn
 }
 
 func (o *observedConnection) Close() error {
-	err := o.TCPConn.Close()
-	metrics.ProxyActiveConnections.Dec()
-	metrics.ProxyConnectionClosed.Inc()
+	err := o.Conn.Close()
```
jranson: here just do `if err := o.Conn.Close(); err != nil { return err } ` and keep the metrics outside of the guard.

jranson (changes requested):
@thinker0 one minor nit here for idiomacy. otherwise looks good.

pkg/proxy/listener/listener.go
thinker0: Addressed the nit. Thanks for the suggestion!

jranson (approved):

PR #1020, opened 2026-05-29
fix: [engines] set Request.GetBody to support HTTP/2 GOAWAY retries

thinker0:
## Description
Fixes `502 Bad Gateway` on POST/PUT requests when the upstream returns an HTTP/2 GOAWAY frame (e.g. keep-alive rotation or rolling updates).

**Root cause**
Go's `http2.Transport` (and `http.Transport` on reused connections) only auto-retries a request after a GOAWAY/connection error if `Request.GetBody` is set, so it can re-read the body. Trickster never set `GetBody`, so the transport aborted the retry with `cannot retry err after Request.Body was written; define Request.GetBody to avoid this error` and the proxy returned 502.

**Fix**
`PrepareFetchReader` buffers the request body via `request.GetBody(r, MaxObjectSizeBytes)` — the existing helper @crandles pointed to — which:
- bounds the read with `io.LimitReader` (returns `413 Request Entity Too Large` past `MaxObjectSizeBytes`),
- caches the body on the request resources to avoid re-reading,
- resets `r.Body` to a re-readable reader.

It then mirrors the bytes into `r.GetBody` so the transport can replay the body on a retry.

**Tests**
- `TestPrepareFetchReader_GetBody` — body replays across multiple `GetBody()` calls.
- `TestHTTP2GOAWAYRetry` — end-to-end retry after a simulated GOAWAY.
- `TestPrepareFetchReader_GetBodyTooLarge` — oversized body is rejected with 413 and the upstream is not called.

## Type of Change
- [x] Bug fix
- [x] Test coverage

## AI Disclosure

### DIFF
--- a/pkg/proxy/engines/httpproxy.go
+++ b/pkg/proxy/engines/httpproxy.go
@@ -36,6 +36,7 @@ import (
 	"github.com/trickstercache/trickster/v2/pkg/observability/tracing"
 	tspan "github.com/trickstercache/trickster/v2/pkg/observability/tracing/span"
 	"github.com/trickstercache/trickster/v2/pkg/proxy/forwarding"
+	"github.com/trickstercache/trickster/v2/pkg/proxy/handlers/trickster/failures"
 	"github.com/trickstercache/trickster/v2/pkg/proxy/headers"
 	"github.com/trickstercache/trickster/v2/pkg/proxy/methods"
 	"github.com/trickstercache/trickster/v2/pkg/proxy/params"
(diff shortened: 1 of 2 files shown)


pkg/proxy/engines/httpproxy.go
```diff
@@ -225,6 +225,23 @@ func PrepareFetchReader(r *http.Request) (io.ReadCloser, *http.Response, int64)
 		r.Header.Set(headers.NameAcceptEncoding, ep.SupportedHeaderVal)
 	}
 
+	if r.Body != nil && r.GetBody == nil {
+		bodyBytes, err := io.ReadAll(r.Body)
```
crandles: we should use a limit reader here, I think. We should be doing so elsewhere, will look for a ref

crandles: Examples:
- https://github.com/trickstercache/trickster/blob/070dff061bbefeaa26210abbb5740ac372bf8973/pkg/proxy/engines/deltaproxycache.go#L650
- https://github.com/trickstercache/trickster/blob/070dff061bbefeaa26210abbb5740ac372bf8973/pkg/proxy/engines/proxy_request.go#L179
- https://github.com/trickstercache/trickster/blob/070dff061bbefeaa26210abbb5740ac372bf8973/pkg/proxy/request/body.go#L63

thinker0: Done — switched to `request.GetBody(r, o.MaxObjectSizeBytes)`, the helper you linked (`pkg/proxy/request/body.go`). It bounds the read with `io.LimitReader`, returns `ErrPayloadTooLarge` past the limit (mapped to 413 here), and caches the body on the request resources so it isn't re-read; the hand-rolled `io.ReadAll` is gone. Added `TestPrepareFetchReader_GetBodyTooLarge` to cover the oversized→413 path. (baa19a97)

jranson (approved):

PR #1035, opened 2026-06-25
fix: finalize topk and bottomk tsm results

### LINKED ISSUE #1034: topk is not returning k series but all
Hello team, 

Expecting topk() returns number of k series but it returns all. 
At the same time, seeing this warning:
`"text": "trickster: outer aggregator \"topk\" cannot be correctly merged across fanout backends; results may be inaccurate"`

On a side note, while nesting sort_desc(topk()) together, warning doesn't show, still showing all series though. 

Thanks.

houyuwushang:
## Description
Fixes #1034.

This makes Prometheus TSM fanout handle `topk` and `bottomk` by unwrapping the rank aggregation before fanout. Shards receive the inner query, Trickster merges the inner results using that inner query's merge strategy, and then the Prometheus finalizer applies the final top/bottom-k rank-and-trim step per timestamp and aggregation group.

This avoids merging backend-local `topk`/`bottomk` candidate sets as if every shard had equal weight. For example, `topk(5, sum by (service) (...))` now fans out `sum by (service) (...)`, merges with the sum strategy, and then ranks the globally merged values. `topk(5, avg by (service) (...))` uses the existing weighted-average path by fanning out sum/count rewrites for the inner avg, finalizing the weighted average, then applying the rank trim.

It also recognizes `sort(...)` and `sort_desc(...)` wrappers around rank aggregations. Unfinalizable forms such as non-literal k still fall back to the existing unsupported-aggregator warning behavior.

Docs are updated to describe inner-query fanout and final rank trimming for `topk` and `bottomk`.

Tests run:

- `go test -count=1 ./pkg/backends/alb/... ./pkg/backends/prometheus/... ./pkg/timeseries/dataset`
- `go tool golangci-lint run --timeout 5m ./pkg/backends/prometheus/... ./pkg/backends/alb/... ./pkg/timeseries/dataset`
- `git diff --check`

Also ran `go test -count=1 ./...`; it still has unrelated local Windows/environment failures in cache access-denial tests using `/root/...` paths and a config golden line-ending mismatch. The Prometheus and ALB packages touched by this PR pass in that full run.

## Type of Change

- - [x] Bug fix
- - [ ] New feature
- - [ ] Optimization
- - [x] Test coverage
- - [x] Documentation
- - [ ] Infrastructure

## AI Disclosure

### DIFF
--- a/docs/alb.md
+++ b/docs/alb.md
@@ -100,12 +100,15 @@ For Federation use cases where backends hold different, non-overlapping data, Tr
 | `max` | Maximum value per unique label set + timestamp |
 | `group` | Deduplicate per unique label set + timestamp |
 | `avg` | Dual queries (avg→sum and avg→count); weighted arithmetic mean per unique label set + timestamp |
-| `stddev`, `stdvar`, `quantile`, `topk`, `bottomk`, `limitk`, `limit_ratio` | Deduplicate + inject a warning into the Prometheus response |
+| `topk`, `bottomk` | Query the inner expression across backends, merge it, then apply final top/bottom-k selection per timestamp and aggregation group |
+| `stddev`, `stdvar`, `quantile`, `limitk`, `limit_ratio` | Deduplicate + inject a warning into the Prometheus response |
 | _(none)_ | Deduplicate (default) |
 
 For `avg` queries, Trickster issues two concurrent sub-queries per backend shard — one rewriting the outer `avg` to `sum` and another to `count` — then computes a true weighted arithmetic mean (`sum_total / count_total`) per series per timestamp. This avoids the skew introduced by a naïve avg-of-averages when backends have different data cardinalities.
 
-For non-supportable aggregators (`stddev`, `stdvar`, `quantile`, `topk`, `bottomk`, `limitk`, `limit_ratio`), Trickster falls back to deduplication and injects a `warnings` entry in the Prometheus response body to alert the caller that results may be inaccurate.
+For `topk` and `bottomk`, Trickster sends the inner expression to each backend, merges those inner results using the inner expression's merge strategy, then applies the final rank-and-trim step per timestamp and aggregation group. This prevents each backend's local `topk`/`bottomk` result from being weighted equally during the merge. If the inner expression is `avg`, Trickster still uses the weighted `sum`/`count` rewrite before applying the final rank. This also applies when the rank aggregation is wrapped in `sort()` or `sort_desc()`.
+
+For non-supportable aggregators (`stddev`, `stdvar`, `quantile`, `limitk`, `limit_ratio`), Trickster falls back to deduplication and injects a `warnings` entry in the Prometheus response body to alert the caller that results may be inaccurate.
 
 When a non-dedup strategy is in effect and backends have [injected labels](./prometheus.md#injecting-labels) configured, those labels are automatically stripped before merging. This ensures series from different backends hash identically for aggregation, and the injected labels do not appear in the response.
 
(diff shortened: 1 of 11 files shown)


jranson:
@houyuwushang thanks for this contribution - it is a good start. However, it only works for simple `topk`/`bottomk` queries (e.g.,   `topk(5, up)`).  It would not properly weight shards with aggregations in the inner query, like `topk(5, sum by (service) (...))` or `topk(5, avg by (service) (...))` because each shard still calculates its own `topk` remotely before merging. Therefore each shard would receive equal weighting during the merge process and cause shards to be over-ranked or under-ranked based on their result lengths. The correct application here would be to unwrap the inner query (e.g., `sum by (service) (...)`) from the `topk`, query the shards with the inner query; then in the time series merge, implement `topk` and `bottomk` rank-and-trim against the merged inner query result, to ensure proper weighting across the shards. We use a similar approach with `avg`, so it might be a good reference for parsing and fanning out the parsed inner query, and then applying the outer aggregator function on the merged inner query results in the TSM. In fact, in the case of a `topk`/`bottomk` with an `avg` in the inner query (e.g., `topk(5, avg by (service) (...))`), we'd need to perform the existing `avg` magic for the inner query:  query the shards for separate sum/count instead of avg, then calculate the properly-weighted average in TSM; and then finally implement the rank-and-trim for topk/bottomk against the internally-weighted averages.

houyuwushang:
Thanks for the detailed guidance. I updated the PR to follow that approach.

The current version now unwraps the rank aggregation before fanout, so shards receive the inner expression instead of the outer `topk`/`bottomk`. `ClassifyMerge` is based on that inner expression, so cases like `topk(5, sum by (service) (...))` use the sum merge strategy before the final rank trim.

I also wired the rank wrapper into the weighted-avg path: `topk(5, avg by (service) (...))` now uses the existing sum/count rewrite for the inner `avg`, finalizes the weighted average in TSM, and only then applies the final top/bottom-k trimming.

Added coverage for:

- parsing and retaining the inner rank query
- classifying rank-wrapped `sum`, `avg`, and `max` inner queries
- rewriting standard rank fanout to the inner query
- rewriting rank-wrapped `avg` to inner sum/count queries
- invoking finalization after the weighted-average path
- docs describing inner-query fanout plus final rank trimming

Local validation:

- `go test -count=1 ./pkg/backends/alb/... ./pkg/backends/prometheus/... ./pkg/timeseries/dataset`
- `go tool golangci-lint run --timeout 5m ./pkg/backends/prometheus/... ./pkg/backends/alb/... ./pkg/timeseries/dataset`
- `git diff --check`

jranson:
@houyuwushang this generally looks good now, thank you. One gap I see is the lack of a test ensuring that the internal `topk`/`bottomk` implementation properly drops values over all shards outside of `k`. For example, `topk(1, ....)` and `bottomk(1, ...)` query tests that merge 2 or more shards and keep only 1 value (since `k` is 1). If you can add those test verifications, this will be ready to merge. A good place for the missing test would be in `integration/alb_tsm_correctness_test.go` since it already bootstraps multiple TSM endpoints for merge testing.

houyuwushang:
Thanks, I added the requested integration coverage in `integration/alb_tsm_correctness_test.go`.

The new case boots two TSM upstreams and verifies both `topk(1, ...)` and `bottomk(1, ...)` return exactly one globally-ranked series after merge. It also asserts the upstream mocks only receive the unwrapped inner query, so the rank wrapper is applied inside TSM after fanout/merge rather than by each shard.

Validation:

- `go test -count=1 ./pkg/backends/alb/... ./pkg/backends/prometheus/... ./pkg/timeseries/dataset`
- `go tool golangci-lint run --timeout 5m ./pkg/backends/prometheus/... ./pkg/backends/alb/... ./pkg/timeseries/dataset`
- `git diff --check`
- `CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go test -c` from `integration/`

jranson (approved):

PR #1036, opened 2026-06-25
refactor: keep user-router runtime state out of options

### LINKED ISSUE #1010: ur: move runtime handler state out of options.Options
`ur.Handler.SetDefaultHandler` writes runtime state into the options struct in [pkg/backends/alb/mech/ur/user_router.go](https://github.com/trickstercache/trickster/blob/main/pkg/backends/alb/mech/ur/user_router.go):

```go
func (h *Handler) SetDefaultHandler(h2 http.Handler) {
    h.mu.Lock()
    if h.options != nil {
        h.options.DefaultHandler = h2
    }
    h.mu.Unlock()
}
```

[pkg/backends/alb/client.go](https://github.com/trickstercache/trickster/blob/main/pkg/backends/alb/client.go) (`validateAndStartUserRouter`) does similar: mutates `o.NoRouteStatusCode` and assigns runtime handlers into `m.ToHandler` / `m.ToStatus`. Setup logic inside what reads as config validation.

Proposal: move runtime handler state (`DefaultHandler`, per-user resolved handlers, `ToStatus`) out of options into `Handler` fields.

houyuwushang:
## Description
Fixes #1010.

Moves resolved user-router runtime state out of `ur/options.Options`:

- stores the default handler, effective no-route status, and per-user route handlers on `ur.Handler`
- removes `DefaultHandler`, `ToHandler`, and `ToStatus` from the user-router options structs
- builds runtime routes locally during `validateAndStartUserRouter` and applies them only after validation succeeds
- keeps the configured `NoRouteStatusCode` unchanged while still using the effective startup response code

Tests:

- `go test -count=1 ./pkg/backends/alb/mech/ur ./pkg/backends/alb`
- `go test -count=1 ./pkg/backends/alb/...`
- `go tool golangci-lint run --timeout 5m ./pkg/backends/alb/...`
- `go test -run '^$' ./...`

## Type of Change

- - [ ] Bug fix
- - [ ] New feature
- - [x] Optimization
- - [x] Test coverage
- - [ ] Documentation
- - [ ] Infrastructure

## AI Disclosure

### DIFF
--- a/pkg/backends/alb/client.go
+++ b/pkg/backends/alb/client.go
@@ -40,6 +40,7 @@ import (
 	"github.com/trickstercache/trickster/v2/pkg/observability/metrics"
 	authopt "github.com/trickstercache/trickster/v2/pkg/proxy/authenticator/options"
 	authreg "github.com/trickstercache/trickster/v2/pkg/proxy/authenticator/registry"
+	at "github.com/trickstercache/trickster/v2/pkg/proxy/authenticator/types"
 	"github.com/trickstercache/trickster/v2/pkg/proxy/handlers"
 	"github.com/trickstercache/trickster/v2/pkg/proxy/handlers/trickster/failures"
 	"github.com/trickstercache/trickster/v2/pkg/proxy/methods"
@@ -230,6 +231,8 @@ func observeOnlyOpts() *authopt.Options {
 func (c *Client) validateAndStartUserRouter(clients backends.Backends, hcs healthcheck.StatusLookup) error {
 	conf := c.Configuration()
 	var canReplaceCreds bool
+	var authenticator at.Authenticator
+	var defaultHandler http.Handler
 	o := conf.ALBOptions.UserRouter
 	h, ok := c.handler.(*ur.Handler)
 	if !ok {
@@ -239,7 +242,7 @@ func (c *Client) validateAndStartUserRouter(clients backends.Backends, hcs healt
 		// credential replacement is only allowed if users will be positively
 		// authenticated and not just observed.
 		canReplaceCreds = !(conf.AuthOptions.Authenticator.IsObserveOnly())
-		h.SetAuthenticator(conf.AuthOptions.Authenticator, canReplaceCreds)
+		authenticator = conf.AuthOptions.Authenticator
 	} else {
 		a, err := authreg.NewObserverFromProviderName(o.TargetProvider,
 			map[string]any{"options": observeOnlyOpts()})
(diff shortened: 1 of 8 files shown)


jranson:
@houyuwushang thanks for this contribution, and for your patience while I tested it out.

I've identified a regression in this change that breaks the 'user remapping' feature by causing a remapped user to be sent through the incorrect user route (or the default fallthrough route), instead of the expected/configured route for the inbound username when the user is configured with `to_backend`.

The main issue is that at pkg/backends/alb/mech/ur/user_router.go:124-125, `username` can be mutated (e.g., `alice` -> `admin`). Then at L150, `routes[username]` is used to route the request to a backend target. Even if the username undergoes remapping, the _original_ `username` is what should be used for routing (so in my example, the `alice` route should still be used, even though the effective username to the backend is remapped to `admin`). However, in your change, my hypothetical inbound `alice` user would be improperly routed to the `admin` route (or the `default` route if `admin` is not configured) instead of their configured `to_backend ` route due to the mutation of the `username` var in the remapping branch.

One fix might be to set an `outboundUsername` value instead of mutating `username` at all, and use `outboundUsername` in the `SetCredentials` call: 
```
outboundUsername := username
if u.ToUser != "" {
    outboundUsername = u.ToUser
}
…
    if err := auth.SetCredentials(r, outboundUsername, cred); err != nil {
        h.handleDefault(w, r)
        return
    }
…
if route, ok := routes[username]; ok && route.Handler != nil {
    …
}
```
This way, the outbound username is properly remapped while the inbound username is still used for the routes map lookup.

jranson (changes requested):
See comments above.

houyuwushang:
Thanks for catching and explaining the regression. I pushed 584ff7fc32d1a0fd8f861bf6863c5f40cb8a29b3 to address it.

The fix keeps the inbound `username` unchanged for `routes[username]` lookup and uses a separate `outboundUsername` only for `SetCredentials`. I also added a regression test where inbound `alice` is remapped to outbound `admin` while both `alice` and `admin` routes exist; the request must still dispatch through the `alice` route while writing `admin` credentials downstream.

Local validation:

- `go test ./pkg/backends/alb/mech/ur -run 'TestServeHTTP/credential_remapping|TestURRetainsInboundCredWhenToCredentialEmpty' -count=1`
- `go test -count=1 ./pkg/backends/alb/mech/ur ./pkg/backends/alb`
- `go test -count=1 ./pkg/backends/alb/...`
- `go test -race ./pkg/backends/alb/mech/ur -run 'TestServeHTTP/credential_remapping_keeps_inbound_username_for_routing' -count=1`
- `go tool golangci-lint run --timeout 5m ./pkg/backends/alb/mech/ur ./pkg/backends/alb`
- `git diff --check`

jranson (approved):

Issue #1011, opened 2026-05-18
tsm: narrow constructor input to a tsm.Config value type


crandles:
tsm's constructor reaches into `o.TSMOptions`, `o.MaxCaptureBytes`, `o.MaxFanoutCaptureBytes`, and `o.OutputFormat` from the full options struct in [pkg/backends/alb/mech/tsm/time_series_merge.go](https://github.com/trickstercache/trickster/blob/main/pkg/backends/alb/mech/tsm/time_series_merge.go). `OutputFormat` is provider metadata, not load-balancing behavior, so threading it through tsm construction couples a mechanism to backend-format knowledge.

Proposal: introduce a `tsm.Config` value type populated once at the registry layer, pass it into `New()` in place of the full options.

Elia-Renzoni:
@crandles I'm interested in working on this issue. Could you assign it to me? Thanks!

jranson:
implemented in #1051

PR #1051, opened 2026-07-05
refactor(alb): avoid options in ALB constructors

Elia-Renzoni:
## Description

This PR resolves a design issue in the Application Load Balancer (ALB) constructor, previously  the time_series_merge constructor was relying on fields from the general `Options` struct that were not actually related to the ALB itself, but rather belonged to a more Backend-style configuration. 

To address this and improve separation of concerns, I introduced a new `ALBConfigs` struct to replace `Options` in the constructor. This is because, in practice, load balancing algorithm constructors rarely need the entire `Options` struct, but rather just a few specific fields.

#### Changes
- **Introduced `ALBConfigs`**: Created a new struct tailored specifically for ALB requirements, decoupling it from the broader `Options`.
- **Refactored Constructors**: Updated the `New` factory functions to accept `ALBConfigs` instead of `Options`.
- **Updated Tests**: Refactored the test suite to align with the new constructor signatures and data structures.

Related Issue: #1011

## Type of Change

- - [ ] Bug fix
- - [ ] New feature
- - [x] Optimization
- - [ ] Test coverage
- - [ ] Documentation
- - [ ] Infrastructure

## AI Disclosure

- - [x] This contribution DOES NOT include AI-generated changes
- - [ ] This contribution DOES include AI-generated changes, and I have reviewed the relevant contributing guidelines.


Elia-Renzoni:
I'm not sure if this is the best approach to solve the issue, so let me know what you think about it. 

An alternative approach could be to create a custom `NewTSM()` constructor that accepts a specific `TSMConfig` struct rather than `Options`. However, having a different struct type just for this constructor might be confusing compared to the other algorithms. That said ALBs generally only use a subset of the `Options` fields in their constructors anyway.

houyuwushang:
I took a pass over this from the #1011 angle. The CI is green, and I also checked the related packages locally:

- `go test -count=1 ./pkg/backends/alb/...`
- `go test -count=1 ./pkg/backends/prometheus/...`
- `go tool golangci-lint run --timeout 5m ./pkg/backends/alb/...`
- `git diff --check origin/main...HEAD`

Two things I think are worth considering before this lands:

1. #1011 specifically proposed a narrower `tsm.Config` populated at the registry layer because `OutputFormat` is provider metadata rather than load-balancing behavior. This PR narrows the input compared with the full `Options`, but the new shared `options.ALBConfigs` still exposes `OutputFormat` to every mechanism constructor, and the TSM constructor still reads `conf.OutputFormat`. That may be the right tradeoff for constructor consistency, but it seems like the main design question for this PR: should this remain a shared ALB constructor config, or should TSM get its own small config type matching the issue proposal?

2. `registry.New` now calls `options.NewALBConfigsFromOptions(opts)` before dispatching to the mechanism constructor. Since `NewALBConfigsFromOptions` dereferences `opts`, `registry.New(names.MechanismRR, nil, ...)` would now panic even though the round-robin constructor itself does not need options and accepts nil. I do not see that on the main ALB startup path because `alb.NewClient` only calls registry.New with non-nil `ALBOptions`, so this may only affect direct package use/tests. Still, it may be worth either adding a nil guard in the conversion helper or adding a regression test that documents nil options are intentionally unsupported at the registry layer.

Not blocking from my side, but I think those two points would make the intended API boundary clearer.

pkg/backends/alb/mech/registry/registry.go
```diff
@@ -40,20 +40,34 @@ var registry = []types.RegistryEntry{
...
 			panic("alb/mech/registry: duplicate mechanism name " + name)
 		}
-		out[name] = fn
+		out[name] = fns
 	}
+
 	for _, entry := range entries {
-		add(entry.ShortName, entry.New)
-		add(entry.Name, entry.New)
+		if entry.New == nil && entry.NewTSM != nil {
```
jranson: **Detected via Claude - Registration and dispatch key off different things — latent nil-func panic**

Here Registration decides which func slot to populate by *inspecting which fields are set*:

```go
if entry.New == nil && entry.NewTSM != nil { ... tsmMechanismFunc ... }
fns := &funcs{mechanismFunc: entry.New}
```

but on L79, Dispatch decides which slot to *call* by **mechanism name**:

```go
if name == tsm.Name || name == tsm.ShortName {
    return f.tsmMechanismFunc(tsmConfigs, factories)
}
return f.mechanismFunc(opts, factories)
```

These two rules can disagree, and every disagreement is a nil func call (panic at config load, taking the process down rather than returning an error):

- An entry with **both** `New` and `NewTSM` set registers as a plain mechanism; if its name is `tsm`, dispatch calls the nil `tsmMechanismFunc`.
- Any **second** mechanism that adopts `NewTSM` registers as TSM-style, but dispatch falls through to the nil `mechanismFunc`.
- An entry with **neither** set now stores a non-nil `*funcs` with a nil `mechanismFunc` → panic.

That last one is a real regression in an existing guard. Previously `f` was the func value itself, so `ok && f != nil` at [registry.go:78](pkg/backends/alb/mech/registry/registry.go:78) correctly rejected a nil-`New` entry with `ErrUnsupportedMechanism`. Now `f` is a pointer that is *always* non-nil when `ok` is true — the `f != nil` check is dead code, and the failure mode changed from a returned error to a panic. Dispatch should test the func, not the name: `if f.tsmMechanismFunc != nil`.

pkg/backends/alb/mech/registry/registry.go
```diff
@@ -40,20 +40,34 @@ var registry = []types.RegistryEntry{
 
 var registryByName = compileSupportedByName(registry)
```
jranson, line 41: **Detected via Claude - The registry now hardcodes knowledge of one specific mechanism**

The whole point of `registryByName` is that adding a mechanism means adding one line to the `registry` slice. After this change, adding a mechanism that wants a narrowed config means editing `types.RegistryEntry`, editing `compileSupportedByName`, and adding a branch to `New` — an N-branch `if` chain in the dispatcher, one per mechanism. That's the opposite direction from the extensibility the registry exists to provide, and it makes `alb/mech/registry` depend on `tsm`'s *identity*, not just its entry.

The PR's stated goal — `tsm.New` takes only what it needs — is fully achievable with **no** change to `types.go` or `registry.go`, via an adapter at the registration site:

```go
// pkg/backends/alb/mech/tsm/time_series_merge.go
func RegistryEntry() types.RegistryEntry {
	return types.RegistryEntry{Name: Name, ShortName: ShortName, New: newFromOptions}
}

func newFromOptions(o *options.Options, factories rt.Lookup) (types.Mechanism, error) {
	return New(options.NewTSMConfigs(o), factories)
}
```

`registry.go` and `types.go` revert to their pre-PR state, the `funcs` struct disappears, both nil-func classes above become unrepresentable, and `nlm`/`fgr`/`ur` can adopt the identical pattern later without touching the registry. If you want it uniform and typed, a one-line generic adapter in `mech/types` expresses it directly:

```go
func Adapt[T any](narrow func(*options.Options) T,
	fn func(T, rt.Lookup) (Mechanism, error)) NewMechanismFunc {
	return func(o *options.Options, l rt.Lookup) (Mechanism, error) { return fn(narrow(o), l) }
}
// New: types.Adapt(options.NewTSMConfigs, New)
```

Either way the diff goes net-negative and the registry stays closed for modification. Worth noting the sibling [`proxy/authenticator/registry`](pkg/proxy/authenticator/registry/registry.go:31) uses the same single-func `RegistryEntry` shape — this PR makes the ALB registry the odd one out.

pkg/backends/alb/options/options.go
```diff
@@ -141,6 +141,24 @@ var (
 	ErrOutputFormatOnlyForTSM = errors.New("'output_format' option is only valid for provider 'alb' and mechanism 'tsmerge'")
 )
 
+type TSMConfigs struct {
```
jranson: This is probably in the wrong package and, along with `TimeSeriesMergeOptions` and `NewTSMConfigs`, should be moved to its own/new `options` package at `pkg/backends/alb/mech/tsm/options/options.go`.

The PR is about decoupling TSM from `alb.Options`, but `TSMConfigs` is currently declared in ALB Options, while `tsm` still imports ALB options. So the dependency is narrowed, not removed, and ALB options still owns the TSM Config shape. The codebase demonstrates the correct pattern in the User Router mechanism: `ur` keeps its mechanism options in `alb/mech/ur/options` while `alb.Options` has a `*ur.Options` member, so `ur` does not depend on ALB Options.

We should repeat that pattern by moving `TimeSeriesMergeOptions` and `TSMConfigs`/`NewTSMConfigs` to a `tsm`-owned options package and have `alb/options` reference it, just like it does for `ur`. This inverts the dependency and fully decouples TSM from ALB Options.

pkg/backends/alb/options/options.go
```diff
@@ -141,6 +141,24 @@ var (
...
+	TimeSeriesMergeOptions
+	MaxCaptureBytes       int
+	MaxFanoutCaptureBytes int
+	OutputFormat          string
+	MechanismName         string
```
jranson: `MechanismName` appears to be a dead field that is only set in `NewTSMConfigs` and never actually read. i recommend removing.

pkg/backends/alb/mech/types/types.go
```diff
@@ -31,6 +31,9 @@ type Name = string
 // provided Options
 type NewMechanismFunc func(*options.Options, types.Lookup) (Mechanism, error)
 
+// NewTSMMechanismFUn defines a function that returns a Time Series Merge mechanism
```
jranson: typo here "NewTSMMechanismFUn" should be "NewTSMMechanismFunc"

pkg/backends/alb/options/options.go
jranson: also please add comments on `TSMConfigs` and `NewTSMConfigs` for godoc.

pkg/backends/alb/mech/tsm/time_series_merge_ctor_test.go
```diff
@@ -35,7 +35,7 @@ func TestRegistryEntry(t *testing.T) {
 	if entry.Name != Name || entry.ShortName != ShortName {
 		t.Fatalf("RegistryEntry = %+v, want Name=%q ShortName=%q", entry, Name, ShortName)
 	}
-	if entry.New == nil {
+	if entry.NewTSM == nil {
 		t.Fatal("RegistryEntry.New is nil")
```
jranson, line 39: this error text should be `RegistryEntry.NewTSM is nil` (New -> NewTSM)

pkg/backends/alb/client.go
```diff
@@ -96,6 +96,7 @@ func NewClient(name string, o *bo.Options, router http.Handler,
 		if err != nil {
 			return nil, err
 		}
+
```
jranson: since there is no material change in this file, it would be good to remove this stray added newline to keep client.go out of the diff.

jranson (changes requested):
@Elia-Renzoni thanks for taking this one. a couple items inline below to fix up. In addition to these, I also noticed that the registry package's tests do not exercise the new TSM dispatch branch, and should also test an entry with a nil `New` to verify `ErrUnsupportedMechanism` vs panicking. Would you mind adding those?

pkg/backends/alb/mech/types/types.go
Elia-Renzoni: fixed in 5fa48258a5c1a43db658e855f9df160d9365d76a

pkg/backends/alb/client.go
Elia-Renzoni: fixed in 5fa48258a5c1a43db658e855f9df160d9365d76a

pkg/backends/alb/mech/tsm/time_series_merge_ctor_test.go
Elia-Renzoni: fixed in d3ef79ed42457d940b46a00676eda75fa8b30948

pkg/backends/alb/options/options.go
Elia-Renzoni: fixed in cc31c0daedbbe0dfca6ac7b60f3f332681b9b7fa

pkg/backends/alb/mech/registry/registry.go
Elia-Renzoni: fixed in e1c739e5511e8218b2eb26d642a336840c26bafb and also in 48c9abe69021ea318e1db7ddd7bd0e736d6565d9

pkg/backends/alb/options/options.go
Elia-Renzoni: fixed in: e91b9586a080b639f403d5f63846bf0cd1bbd6c5 and in 2cacb6d357dda144053b02069e4a219f136bbbc7

I decided to revert TSMConfigs because narrowing the constructor signatures didn't seem worth the added complexity it forced onto the registry. Changing tsm.New's signature meant the registry had to know how to build that narrower type, which is what led to the funcs/NewTSM branching you flagged, so I backed that out and moved OutputFormat into TimeSeriesMergeOptions instead, since it's a field only the TSM mechanism actually uses.

Your suggestion to move TimeSeriesMergeOptions/TSMConfigs/NewTSMConfigs into a TSM-owned options package, mirroring how ur does it, is the right long-term direction and correctly identifies that my change only narrowed the dependency instead of removing it.

The complication is that TimeSeriesMergeOptions, NewestLastModifiedOptions, and FirstGoodResponseOptions all embed ConcurrencyOptions, which lives in the options package. If I move TimeSeriesMergeOptions into a new mech/tsm/options package, that package would need to import options for ConcurrencyOptions, but alb.Options would also need to import mech/tsm/options to hold the *tsm.Options reference and that's an import cycle. ur avoids this because its options don't depend on anything back in options but TSM's do, because of ConcurrencyOptions.

And since NewestLastModifiedOptions and FirstGoodResponseOptions have the same ConcurrencyOptions dependency, moving TSM alone would leave the codebase inconsistent, the same cycle problem would need solving for all three mechanisms together, not just TSM.

That felt like a bigger, separate change, so for now I went with the smaller fix: move OutputFormat into TimeSeriesMergeOptions, revert the registry to its original single-New shape, and keep it clean.

Does this make sense to you? 
My plan is to open a follow-up issue that aims to moves each mechanism's options struct into its own package (mech/tsm/options, mech/nlm/options, mech/fgr/options, etc.), following the ur pattern properly.

jranson:
@Elia-Renzoni the failing test is because, since you added new yaml-serialized fields to an Options struct, you need to run tests with UPDATE_GOLDENS=true to update the golden files (used for unit test verification) with the new default values. `UPDATE_GOLDENS=true make test`, then commit the testdata files that are changed.

also, i think the most recent changes are close but still not quite aligned with other packages' layouts. i [forked your branch](https://github.com/jranson/trickster/commit/2fab26603a3d2dec47722e26ed571ffaf8f48e18) and did some very minor re-arranging to resolve all of the issues. The solution was to put tsm.Config (the active/running object) in the base `tsm` package, while putting the TSM Options (what is unmarshaled when the config file loads) in its own child `options` package. That gets around the import cycle issues while keeping the desired inversion. Separately there was a regression in your changes where the alb's `OutputFormat` value is silently ignored, so I fixed that up too and included the updated golden files.  You can update your branch using mine as a guide, or I can merge my fork in if you'd like. Let me know.  Thanks again for helping get this completed!

Elia-Renzoni:
@jranson 
Thanks for looking into this and making those fixes. Given the upcoming release, feel free to go ahead and merge your fork or push your commits directly to the branch. That sounds like the quickest and most immediate way to get everything wrapped up.

jranson (approved):

PR #1052, opened 2026-07-06
feat: add elasticsearch backend provider

### LINKED ISSUE #7: Support New Origin Type: Elasticsearch
Dependent on #9 completion

houyuwushang:
## Description
Adds a first-class Elasticsearch backend provider.

Delta Proxy Cache is used only for aggregation requests whose date-histogram buckets can be merged independently: `size: 0`, one mandatory range on the configured timestamp field, one top-level `date_histogram`, complete aligned buckets, ascending key order, and no pipeline or scripted aggregations. Supported intervals are fixed intervals and UTC calendar minute/hour/day intervals. Partial edge buckets, response-shaping URL options, and unsupported query shapes fall back to Object Proxy Cache using the exact request as the cache key. Incomplete Elasticsearch responses are rejected by the DPC modeler and are not cached.

The provider preserves each bucket's JSON and aggregation metadata, including nested metrics, and reconstructs `hits.total` from the returned buckets. The timestamp field defaults to `@timestamp` and can be changed with `backends.<name>.elasticsearch.timestamp_field`. Elasticsearch targets are also registered with the Basic authenticator observer for `user_router` configurations.

The developer environment adds Elasticsearch, generated log data, Kibana routed through Trickster, and seeded data-view, saved-search, log-volume visualization, and dashboard objects.

Fixes #7.

## Type of Change

- - [x] Bug fix
- - [x] New feature
- - [ ] Optimization
- - [x] Test coverage
- - [x] Documentation
- - [x] Infrastructure

## AI Disclosure

## Validation

- `go test ./pkg/backends/elasticsearch/... ./pkg/backends/options ./pkg/backends/providers/... ./pkg/backends/alb ./pkg/proxy/authenticator/registry ./pkg/proxy/handlers/elasticsearch ./pkg/config/validate -count=1`
- `go test ./pkg/backends/elasticsearch -run '^TestQueryHandlerCachesOnlyCompleteHistogramBuckets$' -count=20`
- `go test -race ./pkg/backends/elasticsearch/... ./pkg/proxy/authenticator/registry -count=1`
- `go vet ./pkg/backends/elasticsearch/... ./pkg/proxy/authenticator/registry`
- `go tool golangci-lint run --timeout 5m ./pkg/backends/elasticsearch/... ./pkg/proxy/authenticator/registry`
- `go test -run '^$' ./...`
- validated both the full example and developer-environment configurations with Trickster
- checked the Compose YAML, POSIX seed-script syntax, and embedded Kibana saved-object JSON
- `git diff --check origin/main...HEAD`

An earlier revision was exercised through a clean Linux Docker Compose startup, including Elasticsearch/Kibana, Trickster-backed count and date-histogram requests, and the seeded Kibana objects. The log-volume visualization added in this audit has been structurally validated but has not been rerun in Docker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,8 @@ Trickster works with virtually any Dashboard application that makes queries to a
 
 <img src="./docs/images/external/influx_logo_60.png" width=16 /> InfluxDB
 
+Elasticsearch
+
 See the [Supported TSDB Providers](./docs/supported-backend-providers.md) document for full details
 
 ### How Trickster Accelerates Time Series
--- a/README.md
+++ b/README.md
@@ -50,6 +50,8 @@ Trickster works with virtually any Dashboard application that makes queries to a
 
 <img src="./docs/images/external/mysql_logo_60.png" width=24 /> [MySQL](docs/mysql.md)
 
+Elasticsearch
+
 See the [Supported TSDB Providers](./docs/supported-backend-providers.md) document for full details
 
 ### How Trickster Accelerates Time Series
--- a/deploy/kube/configmap.yaml
+++ b/deploy/kube/configmap.yaml
@@ -229,7 +229,7 @@ data:
     default:
 
         # provider identifies the backend provider.
-        # Valid options are: prometheus, influxdb, clickhouse, reverseproxycache (or just rpc)
+        # Valid options are: prometheus, influxdb, clickhouse, elasticsearch, reverseproxycache (or just rpc)
         # provider is a required configuration value
         provider: prometheus
 
--- a/deploy/kube/configmap.yaml
+++ b/deploy/kube/configmap.yaml
@@ -326,8 +326,9 @@ data:
         listener_name: default
 
         # provider identifies the backend provider.
-        # Valid options are: prometheus, influxdb, clickhouse, mysql, rule, alb,
-        # reverseproxycache (or just rpc), or reverseproxy (or just rp)
+        # Valid options are: prometheus, influxdb, clickhouse, elasticsearch,
+        # mysql, rule, alb, reverseproxycache (or just rpc), or reverseproxy
+        # (or just rp)
         # provider is a required configuration value
         provider: prometheus
 
--- a/docs/developer/environment/README.md
+++ b/docs/developer/environment/README.md
@@ -34,6 +34,10 @@ for verification purposes.
 Elasticsearch is seeded on startup with the `trickster-dev-logs` index. The seed
 data includes recent and older `@timestamp` values so developers can verify
 Elasticsearch date histogram caching through Trickster.
+After Kibana connects through Trickster, the Kibana seed container creates the
+`Trickster Dev Logs` data view, saved search, and dashboard for that index. The
+dashboard is available at
+<http://127.0.0.1:5601/app/dashboards#/view/trickster-dev-logs-dashboard>.
 
 You can stop the developer environment by running `make developer-stop`. To
 delete the developer environment, run `make developer-delete` which will destroy
--- a/docs/developer/environment/README.md
+++ b/docs/developer/environment/README.md
@@ -36,7 +36,8 @@ data includes recent and older `@timestamp` values so developers can verify
 Elasticsearch date histogram caching through Trickster.
 After Kibana connects through Trickster, the Kibana seed container creates the
 `Trickster Dev Logs` data view, saved search, and dashboard for that index. The
-dashboard is available at
+dashboard includes a minute-aligned log-volume histogram that exercises Delta
+Proxy Cache requests, followed by the generated log documents. It is available at
 <http://127.0.0.1:5601/app/dashboards#/view/trickster-dev-logs-dashboard>.
 
 You can stop the developer environment by running `make developer-stop`. To
--- a/docs/developer/environment/README.md
+++ b/docs/developer/environment/README.md
@@ -21,12 +21,19 @@ You can combine these make actions `make developer-start serve-dev` if you want.
 
 Once you have the Docker Compose running, and Trickster running locally, visit
 the Grafana Dashboard at <http://127.0.0.1:3000/d/uAJ8w1wZz/trickster-status>.
+The Kibana frontend is available at <http://127.0.0.1:5601> and is configured
+to use Trickster's `es1` Elasticsearch backend at
+`http://host.docker.internal:8480/es1`.
 
 The data in this dashboard is polled by Prometheus from your local Trickster
 dev instance. So the longer you keep this dashboard up and refreshing, the more
 you can test out Trickster acceleration features. You can change the Data Source
 selector to go between various Trickster configs, or bypass Trickster altogether
 for verification purposes.
+
+Elasticsearch is seeded on startup with the `trickster-dev-logs` index. The seed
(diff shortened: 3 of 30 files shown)


jranson:
@houyuwushang Thanks for this! One thing we'd like to see is an updated developer environment [docker-compose.yml](https://github.com/trickstercache/trickster/blob/main/docs/developer/environment/docker-compose.yml) that includes:
* a new Elasticsearch backend added to the existing Trickster config yaml
* Elasticsearch containers included in the compose, with the Elasticsearch UX/Frontend configured with Trickster as the Elasticsearch API (instead of pointing the frontend directly at the underlying API)
* Seeded log data to enable out-of-box testing/verification (e.g., logs dynamically generated on environment startup that include recent timestamps to ensure they are properly cached by Trickster - and possibly some older timestamps to exercise time-based cacheable boundaries; refer to the ClickHouse data seeding in the developer env for ideas)

houyuwushang:
Updated this PR to include the developer Elasticsearch environment requested here:

- added an `es1` Elasticsearch backend to the developer Trickster config, using the memory cache and `@timestamp` field
- added `elasticsearch`, `elasticsearch_seed`, and `kibana` services to the developer docker-compose file
- configured Kibana's `ELASTICSEARCH_HOSTS` to point at Trickster's `es1` API (`http://host.docker.internal:8480/es1`) rather than the underlying Elasticsearch service
- added a startup seed script that recreates `trickster-dev-logs` with generated recent and older `@timestamp` log records for date histogram testing
- documented the Kibana URL and seeded index in the developer environment README

Validated with:

- `go run cmd/trickster/main.go -validate-config -config docs/developer/environment/trickster-config/trickster.yaml`
- `go test ./pkg/backends/elasticsearch/... ./pkg/backends/options ./pkg/backends/providers ./pkg/config/validate -count=1`
- `go test -run '^$' ./...`
- `go tool golangci-lint run --timeout 5m ./pkg/backends/elasticsearch/... ./pkg/backends/options ./pkg/backends/providers ./pkg/proxy/handlers/elasticsearch ./pkg/config/validate`
- YAML structural check of `docs/developer/environment/docker-compose.yml` via Python/PyYAML
- `git diff --check`

I could not run `docker compose config` or start the compose stack locally because this environment does not have the Docker CLI available.

jranson:
> I could not run docker compose config or start the compose stack locally because this environment does not have the Docker CLI available.

@houyuwushang this comment appears to be from your AI agent. While it's OK to use agentic AI, you will need to install docker locally, run the docker compose and visually verify the new functionality for the developer environment. As reviewers, we are using and verifying changes with the developer environment; and other contributors will expect it to work out of the box without any additional manual steps. My initial tests of the updated docker compose indicate that the new Elasticsearch containers are running, but the Kibana instance prints a lot of errors and does not fully integrate with the trickster backend. For example, the Kibana UX does not show the `trickster-dev-logs` index in the index list, and also lacks any pre-configured dashboards to demonstrate and exercise the new functionality using the generated logs.

houyuwushang:
@jranson Thanks, that feedback is fair. To clarify the earlier validation note: the machine I used for the first update did not have a Docker-capable developer environment available, so I should have moved the runtime validation to a suitable environment before posting that update. I have now done that and pushed `4dc4df7` based on the runtime testing.

That update adds:

- `host.docker.internal:host-gateway` for Kibana so Linux Docker environments can resolve the host Trickster process
- a `kibana_seed` service that waits for Kibana, then creates the `trickster-dev-logs` data view, saved search, and `Trickster Dev Logs` dashboard
- README notes for the seeded Kibana dashboard URL

Additional validation performed on a Docker-capable Linux environment:

- `docker compose config`
- `sh -n docs/developer/environment/docker-compose-data/kibana-config/seeding/seed.sh`
- clean compose rebuild/start with fresh volumes for the ES/Kibana developer services
- `go run cmd/trickster/main.go -config docs/developer/environment/trickster-config/trickster.yaml --log-level debug`
- verified `http://127.0.0.1:8480/es1/_cat/indices?v` includes `trickster-dev-logs`
- verified direct Elasticsearch and Trickster-backed `_count` both return 252 documents for `trickster-dev-logs`
- verified a date histogram query through `http://127.0.0.1:8480/es1/trickster-dev-logs/_search`
- verified Kibana status is `available`
- verified `kibana_seed` exits 0 and creates the data view, saved search, and dashboard

GitHub checks are green on the updated head as well.

houyuwushang:
I took another pass at the DPC boundary before review and found a correctness issue: a partial final date-histogram bucket could be treated as complete and then reused from cache.

The update admits only complete, aligned buckets from a single required timestamp range. Unsupported histogram shapes use the exact-request OPC path; incomplete Elasticsearch responses are rejected and not cached. I also preserved aggregation metadata and reconstructed `hits.total`.

I rebased the branch onto current `main`. The seeded Kibana dashboard now includes a log-volume histogram that exercises this path, and the PR description reflects the supported query shape.

jranson:
@houyuwushang we're ready to start reviewing this PR 🎉 Would you be able to resolve the merge conflicts?

houyuwushang:
Resolved the conflicts against current `main`. I also adapted Elasticsearch's extent rewrite to the error-returning backend interface that landed after this branch was opened.

jranson:
@houyuwushang thanks for your work on this. I have done a review of this feature end-to-end and here are the findings:

1 - when issue #7 was created, over 8 years ago, ElasticSearch did not have an internal delta-based query cache (similar to Trickster's); but it does have that today. That means that the need/justification for an ElasticSearch origin is significantly narrowed since any implementation won't actually accelerate ElasticSearch, but instead just replaces the function of the comparable internal ES delta cache; so trickster in the request path only shifts where the acceleration happens.

2 - ElasticSearch is fronted by Kibana the overwhelming majority of the time (the "ELK Stack"). Modern Kibana versions use the '_async_search' API, which is not cacheable since it's a streaming endpoint that is accessed by unique search IDs rather than the underlying query. This means that even if Trickster did provide a true Delta acceleration, the vast majority of users would not even realize that since they would be going through the non-opc+non-dpc-able _async_search endpoint in Kibana anyway.

3 - For the very small percentage of users who use ElasticSearch with Grafana (instead of with Kibana), while Grafana uses the cachable `_msearch` endpoint;  the current implementation also does not work, since the Grafana ES plugin sends the to/from stamps based on the current millisecond epoch, rather than to/from being bucket-aligned. The PR states that partial edge buckets fall back to the OPC. So virtually every Grafana request to ElasticSearch via Trickster falls back to OPC with a 0% hit rate since, since they are not bucket-aligned and it's highly unlikely that multiple users are refreshing their dashboards at the exact same millisecond to realize a cache hit.

4 - right now the PR uses the OPC for the default / catch-all, which means everything is considered cacheable for at least one minute. This doesn't work in practice because some ElasticSearch APIs must always provide real-time data that can't be stale. Evidence of this is in how the dev environment is struggling to work end-to-end right now.

Given these facts, I recommend that we abandon the ElasticSearch support on the basis that there is not much we can do to improve its performance; and accordingly close this PR without merging and close #7.  Let me know if you agree.

houyuwushang:
That makes sense. After considering the current Kibana and Grafana request paths, I agree the provider would rarely deliver useful acceleration, while the catch-all cache behavior adds correctness risk. I'm okay with closing this PR and #7.

Issue #1018, opened 2026-05-28
delta proxy returns HTTP 200 on a proxy-error, should be 5xx


crandles:
Reported via #1016. When DPC collapses an upstream or internal failure into `LookupStatusProxyError`, the client gets HTTP 200 with an empty body instead of a 5xx.

The error result is built in `buildErrorResult` ([deltaproxycache.go:287-294](https://github.com/trickstercache/trickster/blob/1e7d0fe6805fd11b82d799279f881db42b04f9af/pkg/proxy/engines/deltaproxycache.go#L287-L294)) and written at [deltaproxycache.go:511-517](https://github.com/trickstercache/trickster/blob/1e7d0fe6805fd11b82d799279f881db42b04f9af/pkg/proxy/engines/deltaproxycache.go#L511-L517) via `Respond(w, result.statusCode, ...)`, so it emits whatever `doc.StatusCode` happened to be. Worth confirming where the 200 comes from (the recovered Splice panic vs. an empty/zero-status doc) and making proxy errors surface as 5xx.

jranson:
fixed in #1053

PR #1053, opened 2026-07-07
fix: return 500 on internal failures

jranson:
## Description
This fixes the remaining cases outlined in #1018 - when a 200 response is received from a backend but trickster internally fails to handle the request (e.g., unmarshal failure). Other cases where a 200 was being returned instead of 5xx are already addressed. With this change, instead of proxying the 200 response to the caller on internal failures, the response code is now changed to 500 Internal Server Error before responding. When the internal failure is on one or more shards of a TSM operation, the TSM handles the failure(s) the same as if the failure came from the remote origin.  Related tests are updated with the new expectation.

## Type of Change

- - [x] Bug fix
- - [ ] New feature
- - [ ] Optimization
- - [x] Test coverage
- - [ ] Documentation
- - [ ] Infrastructure

## AI Disclosure

### DIFF
--- a/pkg/proxy/engines/deltaproxycache.go
+++ b/pkg/proxy/engines/deltaproxycache.go
@@ -62,6 +62,13 @@ const (
 	errorBodyCap = 1 << 20
 )
 
+func dpcProxyErrorStatusCode(statusCode int) int {
+	if statusCode == 0 || (statusCode >= http.StatusOK && statusCode < http.StatusMultipleChoices) {
+		return http.StatusInternalServerError
+	}
+	return statusCode
+}
+
 // fetchFastForward executes a fast-forward request and merges the result into rts.
 // Returns the fast-forward status string ("off", "hit", "miss", or "err").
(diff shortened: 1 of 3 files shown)


houyuwushang (commented):
One small question: should `dpcProxyErrorStatusCode` also normalize 3xx statuses? `fetchExtents` treats any `resp.StatusCode != http.StatusOK` as a failed extent, so an all-3xx DPC failure would still be recorded as `proxy-error` but respond with that 3xx instead of a 5xx. If preserving upstream redirects/statuses is intentional here, this is fine; I just wanted to check because #1018 frames `proxy-error` responses as needing a 5xx.

jranson:
Currently none of the supported backends that use DPC (prometheus, influxdb, clickhouse) return 3xx responses to a Trickster request (no redirects, and Trickster does not send `If-Modified-Since` headers to time series backends), so this should not be an issue.

LimitlessEarth (approved):

PR #1057, opened 2026-07-13
fix: close demand probe response bodies

houyuwushang:
## Description

The on-demand healthcheck path copied an upstream response body to the caller but never closed it. That left response-body resources owned by the transport open after each demand probe and could prevent normal connection cleanup.

This closes a non-nil upstream body after forwarding completes. The regression test uses a tracking `ReadCloser` to verify both that the body is closed and that its content is still returned to the caller.

Validation:

- `go test ./pkg/backends/healthcheck/... ./pkg/proxy/handlers/health/... -count=1`
- `go test -race ./pkg/backends/healthcheck -run '^(TestDemandProbe|TestDemandProbeClosesUpstreamBody)$' -count=20`
- `go tool golangci-lint run --timeout 5m ./pkg/backends/healthcheck/...`
- `go test -run '^$' ./...`

Reviewer: @jranson

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Optimization
- [x] Test coverage
- [ ] Documentation
- [ ] Infrastructure

## AI Disclosure

### DIFF
--- a/pkg/backends/healthcheck/target.go
+++ b/pkg/backends/healthcheck/target.go
@@ -315,6 +315,9 @@ func (t *target) notifyStatus(st int32, detail string) {
 func (t *target) demandProbe(w http.ResponseWriter) {
 	r := t.baseRequest.Clone(context.Background())
 	resp, err := t.httpClient.Do(r)
+	if resp != nil && resp.Body != nil {
+		defer resp.Body.Close()
+	}
 	h := w.Header()
 	if err != nil {
 		if t.status != nil && t.status.Get() != StatusUnchecked {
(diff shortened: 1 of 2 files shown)


jranson (changes requested):
@houyuwushang this is a valid defect, thanks for finding! However, the fix is incomplete. It is possible (and likely) for an error response to include a body, but this fix only closes the body of successful responses since a non-nil `err` would cause `demandProbe` to exit prior to the deferred body close. So error response bodies would remain unclosed.  A complete fix would be to do this immediately after the `t.httpClient.Do` call on L317 and before the `err` nil check on L319:
```
if resp != nil && resp.Body != nil {
    defer resp.Body.Close()
}
```

houyuwushang:
Moved the body close immediately after `Do` and guarded both `resp` and `resp.Body`, so a response returned alongside an error is closed before the error path exits.

jranson (approved):

PR #1060, opened 2026-07-14
fix: batch high-cardinality tsm rank merges

### LINKED ISSUE #1056: TSM rank rewrite and merge exceed the caller timeout for a topk query
Disclaimer: Used AI agent to diagnose the issue and write this report

## Summary

With Trickster 2.0.4 (`91f547433698f221094e03def7d9503c49d9d358`), a TSM ALB query can wait until the caller’s timeout and return 502 when the fanout contains a mix of successful and non-2xx downstream responses.

The same PromQL succeeds when sent directly to healthy pool members. Downstream responses complete quickly, indicating that most of the delay occurs in Trickster’s fanout/gather path.

## Setup

- Trickster 2.0.4
- Prometheus-compatible TSM ALB
- Instant PromQL query
- Large multi-backend pool
- Mixed downstream results, including HTTP 200 and HTTP 401

## Observed behavior

- The TSM request produces no response headers before the client’s 90-second timeout.
- The caller receives 502 (CORRECTION) the observed 502 was returned by Grafana after its 90-second datasource timeout, not directly by Trickster. Our direct Trickster request timed out after 110 seconds without receiving an HTTP response. The earlier non-2xx member results came from a test that omitted the required org header and are unrelated to the reproduced issue.
- (CORRECTION: please disregard this line which was my mistake in testing) In one reproduction, a failing member returned HTTP 401 with `no org id` in approximately 0.2 seconds, while a working member returned HTTP 200 in approximately 0.34 seconds.
- Despite these downstream responses completing quickly, the TSM request waited until the client timeout.
- Representative Trickster logs include:

  - `vector unmarshaling error`
  - `tsm gather merge failure: no timerange query`
  - `tsm gather failed due to nil func`

## Expected behavior

A failed or non-2xx pool member should not hold the entire TSM request until the client deadline.

If at least one member returns usable data, Trickster should return a bounded partial response with warnings. If no member contributes usable data, it should return an error promptly.

## Related issue

This appears related to #1040 / #1041, but that fix is already included in the tested commit and seems focused on proxy-only/cache-range responses. The remaining case appears to involve instant queries with mixed successful and non-2xx member responses.

houyuwushang:
## Description

Fixes #1056.

TSM removes an outer `topk`/`bottomk` before fanout so the rank is computed from globally merged values. That exposes each member's full result, and the gather path was then merging every response inside its fanout callback under one accumulator lock. With high-cardinality responses, the lock serialized the merges while the series lookup and tag ordering were rebuilt for every member.

This change keeps the global rank semantics and removes that bottleneck:

- fanout callbacks decode and store contributions by member slot; after fanout, TSM invokes an optional backend-supplied `BatchMergeFunc` carried through `request.Resources`, or falls back to each configured `MergeFunc` in slot order;
- Prometheus opts into the batch path for query and query-range responses, where series from all members share one hash-index pass and one tag sort while preserving the configured strategy and dedup tolerance;
- tag sort keys are computed once per series;
- rank finalization keeps only `k` candidates per timestamp/group in a bounded heap and computes label tie-break keys lazily;
- non-DataSet merge functions retain the existing fallback behavior and panic isolation.

The member-side query remains unranked and the original rank operation is still applied after the global merge. This does not use the local-`topk` diagnostic shortcut from the issue.

A 64-member, 1,000-series-per-member benchmark on the same host measured:

| Series distribution | Sequential merge | Batch merge |
| --- | ---: | ---: |
| Disjoint | 2.16 s | 107 ms |
| Fully overlapping | 110 ms | 31 ms |

For a 64,000-series `topk(20)` result, bounded rank finalization measured about 14 ms with 7.7 KB allocated.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Optimization
- [x] Test coverage
- [ ] Documentation
- [ ] Infrastructure

## AI Disclosure

### DIFF
--- a/pkg/backends/prometheus/tsm_rank_finalize_test.go
+++ b/pkg/backends/prometheus/tsm_rank_finalize_test.go
@@ -17,6 +17,12 @@
 package prometheus
 
 import (
+	"cmp"
+	"math"
+	"reflect"
+	"slices"
+	"strconv"
+	"strings"
 	"testing"
 
 	"github.com/trickstercache/trickster/v2/pkg/timeseries/dataset"
@@ -113,6 +119,61 @@ func TestFinalizeTSMMergeRankAggregation(t *testing.T) {
 	})
 }
 
+func TestRankCandidateHeapMatchesFullSort(t *testing.T) {
+	values := []float64{4, 1, math.NaN(), 9, 9, -2, math.Inf(1), math.Inf(-1), 4, 7}
+	candidates := make([]rankCandidate, len(values))
+	for i, value := range values {
+		candidates[i] = rankCandidate{
+			series: &dataset.Series{Header: dataset.SeriesHeader{Tags: dataset.Tags{
+				"rank": []string{"c", "a", "nan", "b", "a", "z", "inf", "ninf", "c", "d"}[i],
+			}}},
+			value: value,
+			order: i,
+		}
+	}
+
+	for _, operator := range []string{rankOperatorTopK, rankOperatorBottomK} {
+		for _, limit := range []int{0, 1, 3, len(candidates), len(candidates) + 2} {
+			name := operator + "/" + strconv.Itoa(limit)
+			t.Run(name, func(t *testing.T) {
+				wantCandidates := append([]rankCandidate(nil), candidates...)
+				sortRankCandidatesReference(wantCandidates, operator)
+				wantCandidates = wantCandidates[:min(limit, len(wantCandidates))]
+
+				h := &rankCandidateHeap{operator: operator}
+				for _, candidate := range candidates {
+					h.consider(candidate, limit)
+				}
+				gotCandidates := append([]rankCandidate(nil), h.items...)
+				sortRankCandidatesReference(gotCandidates, operator)
+
+				orders := func(input []rankCandidate) []int {
+					out := make([]int, len(input))
+					for i := range input {
+						out[i] = input[i].order
+					}
+					return out
+				}
+				if got, want := orders(gotCandidates), orders(wantCandidates); !reflect.DeepEqual(got, want) {
+					t.Fatalf("selected orders got %v want %v", got, want)
+				}
+			})
+		}
+	}
+}
+
+func sortRankCandidatesReference(candidates []rankCandidate, operator string) {
+	slices.SortStableFunc(candidates, func(a, b rankCandidate) int {
+		if c := compareRankCandidateValues(a, b, operator); c != 0 {
+			return c
+		}
+		if c := strings.Compare(a.series.Header.Tags.JSON(), b.series.Header.Tags.JSON()); c != 0 {
+			return c
+		}
+		return cmp.Compare(a.order, b.order)
+	})
+}
+
 func rankDataSet(series ...*dataset.Series) *dataset.DataSet {
 	return &dataset.DataSet{
 		Results: dataset.Results{{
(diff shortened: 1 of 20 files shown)


pkg/timeseries/dataset/dataset_batch_merge_test.go
```diff
@@ -0,0 +1,217 @@
...
+				b.StopTimer()
+				inputs := benchmarkMergeFixture(64, 1000, overlap)
+				b.StartTimer()
+				for _, next := range inputs[1:] {
+					inputs[0].MergeWithStrategy(true, int(MergeStrategySum), next)
+				}
+			}
+		})
+		b.Run(name+"/batch", func(b *testing.B) {
+			for range b.N {
```
jranson: should use use `for b.Loop()` instead of `for range b.N`

pkg/timeseries/dataset/dataset_batch_merge_test.go
```diff
@@ -0,0 +1,217 @@
...
+}
+
+func BenchmarkFanoutDataSetMerge(b *testing.B) {
+	for _, overlap := range []bool{false, true} {
+		name := "disjoint"
+		if overlap {
+			name = "overlap"
+		}
+		b.Run(name+"/sequential", func(b *testing.B) {
+			for range b.N {
```
jranson: should use use `for b.Loop()` instead of `for range b.N`

pkg/backends/prometheus/tsm_rank_finalize_test.go
```diff
@@ -195,3 +256,17 @@ func equalStringSlicesByKey(a, b map[string][]string) bool {
 	}
 	return true
 }
+
+func BenchmarkFinalizeTSMMergeTopK(b *testing.B) {
+	for range b.N {
```
jranson: should use use `for b.Loop()` instead of `for range b.N`

jranson (changes requested):
@houyuwushang see recommendations for go1.24+ benchmark semantics (iterating `b.N` vs `b.Loop()`).

For the actual batch merge capability, right now in this change the prometheus-compatible batch merge function is declared inside of the TSM package, which would require additional refactoring if we ever want to expand TSM to providers other than Prometheus. Also, since the batch function bypasses the configured `merge.MergeFunc` altogether, I think it might be good to handle this similarly to how the `merge.MergeFunc` feature is done:
- `merge` package also defines a `BatchMergeFunc` type and owns the batch merge semantics
- `request.Resources` carries a new BatchMergeFunc member alongside `MergeFunc`
- TSM-compatible Backend packages (currently only `prometheus`) declare their BatchMergeFunc-compatible func + assign to `rsc` in their handlers

Something like this might work to separate the concerns:

in pkg/proxy/response/merge:
```
type BatchItem struct {
    Data   any
    Member int
}

type BatchMergeFunc func(
    accumulator *Accumulator,
    items []BatchItem,
) (handled bool, err error)
```

A `handled == false` response means the inputs are not batch-compatible, the accumulator was not mutated, and the sequential functions should be used as a fallback in case the runtime data is not actually batch-compatible.


in pkg/proxy/request/resources.go:
```
type Resources struct {
   ...
    MergeFunc merge.MergeFunc
    // add:
    BatchMergeFunc merge.BatchMergeFunc
   ...
}
```

in pkg/backends/alb/mech/tsm/time_series_merge.go
```
type gatherContribution struct {
	data      any
	mergeFunc merge.MergeFunc
	batchMergeFunc merge.BatchMergeFunc  // Add here
	member    int
}
```

houyuwushang:
I moved the batch path behind a backend-supplied `BatchMergeFunc`. The `merge` package now owns the contract and timeseries implementation, `request.Resources` carries it, and the Prometheus query handlers opt in. TSM falls back to each original `MergeFunc` in slot order if any contribution lacks the capability or the batch function returns `handled == false`.

pkg/timeseries/dataset/dataset_batch_merge_test.go
houyuwushang: Updated to use `for b.Loop()`.

pkg/timeseries/dataset/dataset_batch_merge_test.go
houyuwushang: Updated to use `for b.Loop()`.

pkg/backends/prometheus/tsm_rank_finalize_test.go
houyuwushang: Updated to use `for b.Loop()`.

jranson (approved):

PR #1064, opened 2026-07-15
fix: merge aggregations through prometheus sort wrappers

### LINKED ISSUE #1063: Support aggregation-aware TSM merge through sort() / sort_desc() wrappers
When Trickster is used for Prometheus federation with Time Series Merge (TSM) enabled, merge strategy selection appears to depend on the outermost PromQL aggregation operator.

For example, this can receive an aggregation-aware merge:

```promql
count by () (
  group by () (
    last_over_time(metric[...])
  )
)
```

However, dashboard queries often wrap the aggregation in an ordering function:

```promql
sort_desc(
  count by () (
    group by () (
      last_over_time(metric[...])
    )
  )
)
```

In this case, Trickster appears to treat `sort_desc` as the outermost expression and fall back to deduplication rather than detecting the nested `count`.

That can produce incorrect federated results when shard responses have colliding label sets, because the counts are not summed before sorting.

`sort_desc(topk(...))` and `sort_desc(bottomk(...))` already receive special handling. Could Trickster generalize this behavior for `sort()` and `sort_desc()` wrappers around mergeable aggregations?

## Expected behavior

1. Detect an outer `sort()` or `sort_desc()` wrapper.
2. Inspect its inner expression to determine the appropriate merge strategy, such as `sum`, `count`, `avg`, `min`, or `max`.
3. Merge the inner results across backends using that strategy.
4. Apply the final sort after the merge.

This would preserve PromQL semantics while allowing common Grafana ordering patterns to work correctly in federated TSM deployments.

_This issue was drafted with assistance from an AI agent._

houyuwushang:
## Description

Fixes #1063.

TSM previously classified PromQL from the outer expression, so queries such as `sort_desc(count by (...) (...))` fell back to deduplication. Keeping the wrapper on each fanout request also meant no final ordering was applied to the globally merged vector.

This change:

- recognizes outer `sort()` and `sort_desc()` wrappers, including nested wrappers and valid whitespace;
- classifies the wrapped aggregation so `sum`, `count`, `avg`, `min`, and `max` retain their existing merge behavior and unsupported aggregators retain their warning;
- sends the inner aggregation to fanout members, including the sum/count rewrite used for weighted averages;
- applies the requested ordering after the global merge for instant vectors, including native histograms ordered by observation sum;
- leaves range-query ordering and non-aggregation sort expressions unchanged.

The scatter/gather regression test uses two shards with colliding label sets and verifies that their counts are summed before the final descending sort.

Validation:

- `go test ./pkg/backends/prometheus/... ./pkg/backends/alb/... ./pkg/proxy/... -count=1`
- `go test -race ./pkg/backends/prometheus/promql ./pkg/backends/prometheus ./pkg/backends/alb/mech/tsm -run '^Test(ParseSortWrapper|ParseRankAggregation.*|ClassifyMerge|RewriteForTSMMerge.*|RewriteForWeightedAvg.*|FinalizeTSMMerge.*|ServeStandardMergesThroughSortWrapper)$' -count=10`
- `go vet ./pkg/backends/prometheus/... ./pkg/backends/alb/...`
- `go tool golangci-lint run --timeout 5m ./pkg/backends/prometheus/... ./pkg/backends/alb/mech/tsm`
- `go test -run '^$' ./...`
- Linux/amd64 integration test binary cross-compile

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Optimization
- [x] Test coverage
- [ ] Documentation
- [ ] Infrastructure

## AI Disclosure

### DIFF
--- a/pkg/backends/alb/mech/tsm/time_series_merge_finalizer_test.go
+++ b/pkg/backends/alb/mech/tsm/time_series_merge_finalizer_test.go
@@ -17,6 +17,7 @@
 package tsm
 
 import (
+	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
@@ -26,11 +27,15 @@ import (
 
 	"github.com/trickstercache/trickster/v2/pkg/backends/alb/pool"
 	"github.com/trickstercache/trickster/v2/pkg/backends/healthcheck"
+	bo "github.com/trickstercache/trickster/v2/pkg/backends/options"
+	"github.com/trickstercache/trickster/v2/pkg/backends/prometheus"
 	"github.com/trickstercache/trickster/v2/pkg/observability/logging/logger"
 	"github.com/trickstercache/trickster/v2/pkg/proxy/params"
 	"github.com/trickstercache/trickster/v2/pkg/proxy/request"
+	"github.com/trickstercache/trickster/v2/pkg/proxy/response/merge"
 	"github.com/trickstercache/trickster/v2/pkg/timeseries"
 	"github.com/trickstercache/trickster/v2/pkg/timeseries/dataset"
+	"github.com/trickstercache/trickster/v2/pkg/timeseries/epoch"
 )
 
 type finalizerStubBackend struct {
@@ -71,6 +76,12 @@ type rankRewriteFinalizerStubBackend struct {
 	innerQuery string
 }
 
+type prometheusSortBackend struct {
+	*prometheus.Client
+}
+
+func (b *prometheusSortBackend) Configuration() *bo.Options { return nil }
+
 func (b *rankRewriteFinalizerStubBackend) ClassifyMerge(string) (int, bool, string) {
 	return int(dataset.MergeStrategySum), false, ""
 }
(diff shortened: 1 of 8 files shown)


jranson (changes requested):
@houyuwushang 2 issues to fix up:

1. Per the [PromQL Documentation](https://prometheus.io/docs/prometheus/latest/querying/functions/#sort), "Histogram samples in the input vector are ignored silently."   Therefore, we need to check if each `*dataset.Series` is a histogram and, if so, exclude those series from the sorted results.

You can use a check like this and handle only non-histograms (without having to use `json.Unmarshal` repeatedly):

```golang
const histogramFieldName = "histogram"

func isHistogramSeries(series *dataset.Series) bool {
	return series != nil &&
		len(series.Header.ValueFieldsList) > 0 &&
		series.Header.ValueFieldsList[0].Name == histogramFieldName
}
```

2. `sortPointValue` is repeatedly called via `slices.SortStableFunc`, which is inefficient because it repeatedly parses the same parts of the series in an `O(n log n)` fashion. You should be able to precompute all series' values and tags in `sortSeriesIfInstant` and completely eliminate `sortPointValue` like this (includes fix for item 1):

```golang
type sortItem struct {
	series *dataset.Series
	value  float64
	tags   string
}

func sortSeriesIfInstant(
	seriesList dataset.SeriesList,
	descending bool,
) dataset.SeriesList {
	// PromQL sort()/sort_desc() ignore histogram samples. Filter using
	// the series metadata rather than attempting to recognize JSON values.
	filtered := seriesList[:0]
	for _, series := range seriesList {
		if series == nil || isHistogramSeries(series) {
			continue
		}
		filtered = append(filtered, series)
	}
	seriesList = filtered

	if len(seriesList) < 2 {
		return seriesList
	}

	// Preserve the existing instant-vector safety check.
	if len(seriesList[0].Points) != 1 {
		return seriesList
	}
	epoch := seriesList[0].Points[0].Epoch
	for _, series := range seriesList {
		if len(series.Points) != 1 || series.Points[0].Epoch != epoch {
			return seriesList
		}
	}

	// Parse each value and label set once instead of once per comparison.
	items := make([]sortItem, 0, len(seriesList))
	for _, series := range seriesList {
		value, ok := rankPointValue(series.Points[0])
		if !ok {
			// This should not happen for a valid Prometheus float sample.
			// Keeping it at the end matches the current behavior.
			value = math.NaN()
		}
		items = append(items, sortItem{
			series: series,
			value:  value,
			tags:   series.Header.Tags.JSON(),
		})
	}

	operator := rankOperatorBottomK
	if descending {
		operator = rankOperatorTopK
	}

	slices.SortStableFunc(items, func(a, b sortItem) int {
		if rankValueLess(a.value, b.value, operator) {
			return -1
		}
		if rankValueLess(b.value, a.value, operator) {
			return 1
		}
		return strings.Compare(a.tags, b.tags)
	})

	for i := range items {
		seriesList[i] = items[i].series
	}
	return seriesList
}
```

houyuwushang:
Updated this to drop histogram series based on `ValueFieldsList` before sorting, and to precompute each remaining series' value and tags once.

jranson:
@houyuwushang one more thing to fix up here. Examining how the Prometheus [engine](https://raw.githubusercontent.com/prometheus/prometheus/main/promql/engine.go) [handles](https://raw.githubusercontent.com/prometheus/prometheus/main/promql/functions.go) `sort()` and `sort_desc()` with range queries, it still currently still sends each step value's vectors to the sort function, parses the floats, and strips the histograms; even though it does not perform the sort operation on the remaining series. It also injects a warning about sort being ineffective for range queries. So we need to duplicate that functionality by separating `sortSeriesIfInstant` into separate `filterHistogramSeries` and `sortInstantSeries` functions. Then in `finalizeSortWrapper`, conditionally call those and inject the warning.


```golang
const sortInRangeQueryWarning =
	"PromQL warning: sort is ineffective for range queries " +
		"since results are always ordered by labels"

func finalizeSortWrapper(ds *dataset.DataSet, descending bool) {
	isRangeQuery := ds.Step() > 0

	ds.UpdateLock.Lock()
	defer ds.UpdateLock.Unlock()

	if isRangeQuery {
		appendWarningOnce(ds, sortInRangeQueryWarning)
	}

	for _, result := range ds.Results {
		if result == nil {
			continue
		}

		// sort() and sort_desc() omit histogram samples in both instant
		// and range evaluations.
		result.SeriesList = filterHistogramSeries(result.SeriesList)

		// Range-query output has fixed label ordering, so do not reorder it.
		if !isRangeQuery {
			sortInstantSeries(result.SeriesList, descending)
		}
	}
}

func appendWarningOnce(ds *dataset.DataSet, warning string) {
	if !slices.Contains(ds.Warnings, warning) {
		ds.Warnings = append(ds.Warnings, warning)
	}
}
```

houyuwushang:
Updated the range-query path to filter histogram series without reordering and add the PromQL warning once. I also covered `sort(topk(...))`, which goes through the rank finalizer instead of `finalizeSortWrapper`.

jranson (approved):

Issue #1008, opened 2026-05-18
make CORS opt-in per backend now that user_router replaces credentials


crandles:
trickster currently sets `Access-Control-Allow-Origin: *` on every response unconditionally in `pkg/proxy/headers/forwarding.go` `AddResponseHeaders` (line 336). that was defensible when caches only held public read-only telemetry, but the new user_router credential-replacement path (`pkg/backends/alb/mech/ur/user_router.go`) lets cached responses carry privileged upstream data, so any browser origin can issue credentialed GETs and read the body.

CORS should be opt-in per backend/path, or scoped to a configured allowlist, rather than wildcard-on-by-default.

jranson:
we will need to be careful with the implementation here. Since Trickster is a reverse proxy, we can't just accept+proxy any origin's remotely-provided CORS headers, since a permitted origin may not (and _should not_) have awareness of the subdomains used by any reverse proxies (e.g., Trickster instances) between the caller and the origin. For example, if we preserved the origin's headers which send `Access-Control-Allow-Origin: https://api.internal.example.com`, the browser CORS check would fail if the Trickster domain is `trickster.example.com`. That is why the `*` is used by default.  So we'd need to provide an easy way for operators, on a per-backend-basis, to either opt to preserve all of the origin's CORS headers unaltered, preserve the origin's headers while overriding specific ones (e.g., replace the domain in the `Access-Control-Allow-Origin` header) or otherwise overwrite all CORS headers (either with our existing permissive `*` or a custom collection).

houyuwushang:
@jranson I traced the response paths and think this can use one policy consistently for proxied, OPC, and DPC responses. Would this config shape match what you had in mind?

```yaml
cors:
  mode: merge # preserve, merge, or replace
  headers:
    Access-Control-Allow-Origin: https://trickster.example.com
```

`preserve` would leave the origin's `Access-Control-*` headers alone, `merge` would preserve them and apply the configured overrides, and `replace` would remove them all before applying the configured map. An empty map in `replace` mode would disable CORS. A path-level block could override the backend-level policy.

There are two compatibility choices I don't want to guess at: should an omitted `cors` block retain today's `*` behavior for 2.x, and for an ALB/user_router should the client-facing backend's policy win over the selected member's policy? My inclination is yes to both. `preserve` and `merge` would also vary cache keys by the request `Origin` so an origin-specific response header is not replayed to another origin.

jranson:
completed in #1066

PR #1066, opened 2026-07-16
feat: configure backend CORS policies

houyuwushang:
## Description

Related to #1008.

Adds per-backend and per-path CORS response policies with `preserve`, `merge`, and `replace` modes. The client-facing route keeps ownership of the policy across ALB, user-router, and rule dispatch, and policies that preserve origin headers partition cache keys by the request `Origin`.

To avoid changing existing 2.x behavior, an omitted `cors` block still applies the legacy `Access-Control-Allow-Origin: *` behavior. CORS can be deliberately disabled with `mode: disable`, which rejects any `headers` block.

CodeQL validation also identified request-derived values reflected by the purge handlers. Those status and error responses now escape backend names and purge targets before passing through response-writer middleware.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [x] Test coverage
- [x] Documentation
- [ ] Infrastructure

## AI Disclosure

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ Trickster is a fully-featured HTTP Reverse Proxy Cache for HTTP applications lik
 * [Supports TLS](./docs/tls.md) and HTTP/2 for frontend termination and backend origination
 * Offers several options for a [caching layer](./docs/caches.md), including in-memory, filesystem, Redis and bbolt
 * [Highly customizable](./docs/configuring.md), using simple yaml configuration settings, [down to the HTTP Path](./docs/paths.md)
+* Per-backend and per-path [CORS response policies](./docs/cors.md)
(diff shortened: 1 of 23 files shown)


jranson (changes requested):
@houyuwushang thanks for this contribution.  One small change for disabling CORS altogether: I'd prefer we support an explicit `mode: disable` that functions like `mode: replace` with an empty `headers: {}` block. `Options.Validate()` should fail if a user configures `mode: disable` but also configures a `headers` block. That makes the intent deliberate and obvious to anyone who might be reviewing a trickster config while lacking full awareness of this new feature.

houyuwushang:
I added `mode: disable` with the same header-removal behavior as an empty `replace` policy. It rejects any configured `headers` block, including `headers: {}`, and the docs and examples now use the explicit mode.

jranson (approved):

PR #1067, opened 2026-07-17
feat: add automatic config reload polling

### LINKED ISSUE #904: [feature request] Can Trickster detect and reload configs automatically?
Hello Trickster team,

Currently, Trickster provides an endpoint ( http://8484/trcikster/config/reload ) for configs reloading and this is working good.

In our environment, we run Trickster in K8s with 10 (may be more in future) pods deployment. We have implemented a helper app that does the heavy lifting of updating configmaps (for example, a new member joining an ALB backend) and invoke the reload endpoint at every pod. 

This supposed to be a simple task, however, in reality, there are failing factors within K8s runtime. For example, pods may not come up at the same time when helper pokes the kube-api.

Ideally, if Trickster can watch for configs (file contents) change (at file system level) and call reload internally, it will be a great help at our side not to worry if any pods are missing. Also this is a more "native" approach than getting an external app to do it one by one. 

Thanks, 
Keith

houyuwushang:
## Description

Fixes #904.

Adds opt-in configuration-source polling through `mgmt.auto_reload_interval`. The watcher compares the full effective-source fingerprint before invoking the existing graceful reload path, so changes to a file, included fragments, or a configured directory can trigger a reload. Unchanged polls do not increment reload-attempt metrics or engage the reload rate limiter. A successful reload publishes the new configuration back to the watcher, allowing the interval to be changed or disabled without restarting Trickster.

The default remains `0s` (disabled), and negative intervals are rejected. Tests cover Kubernetes ConfigMap projected-volume symlink swaps, same-timestamp content changes, and additions or removals in the multi-source configuration paths introduced by #908.

Validation:

- focused config-source and auto-reload tests repeated 20 times
- race-enabled config-source, management, and daemon reload tests
- affected-package `go vet` and incremental `golangci-lint`
- Windows binary build and the config suite excluding the existing platform-dependent golden test

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [x] Test coverage
- [x] Documentation
- [ ] Infrastructure

## AI Disclosure

### DIFF
--- a/examples/conf/example.full.yaml
+++ b/examples/conf/example.full.yaml
@@ -765,8 +765,11 @@ backends:
 #   reload_drain_timeout: 30s
 #   # reload_rate_limit specifies the rate limit timeout duration to apply to the HTTP reload interface.
 #   # The reload interface is disabled for this duration of time whenever a config reload request is
-#   # made that fails because the underlying config file is unmodified. default is 3
+#   # made that fails because the underlying config file is unmodified. default is 3s
 #   reload_rate_limit: 3s
+#   # auto_reload_interval controls how often Trickster checks the configuration file for changes.
+#   # A zero value disables automatic reloads. default is 0s
+#   auto_reload_interval: 10s
 
 #   # config_handler_path provides the HTTP path to view a read-only printout of the running configuration
 #   # which can be reached at http://your-trickster-endpoint:port/$config_handler_path
--- a/examples/conf/example.full.yaml
+++ b/examples/conf/example.full.yaml
@@ -770,9 +770,9 @@ backends:
 #   reload_drain_timeout: 30s
 #   # reload_rate_limit specifies the rate limit timeout duration to apply to the HTTP reload interface.
 #   # The reload interface is disabled for this duration of time whenever a config reload request is
-#   # made that fails because the underlying config file is unmodified. default is 3s
+#   # made that fails because the effective configuration sources are unchanged. default is 3s
 #   reload_rate_limit: 3s
-#   # auto_reload_interval controls how often Trickster checks the configuration file for changes.
+#   # auto_reload_interval controls how often Trickster checks the effective configuration sources for changes.
 #   # A zero value disables automatic reloads. default is 0s
 #   auto_reload_interval: 10s
 
--- a/examples/conf/example.full.yaml
+++ b/examples/conf/example.full.yaml
@@ -770,8 +770,11 @@ backends:
 #   reload_drain_timeout: 30s
 #   # reload_rate_limit specifies the rate limit timeout duration to apply to the HTTP reload interface.
 #   # The reload interface is disabled for this duration of time whenever a config reload request is
-#   # made that fails because the underlying config file is unmodified. default is 3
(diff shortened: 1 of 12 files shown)


examples/conf/example.full.yaml
```diff
@@ -747,18 +747,21 @@ backends:
 #   # empty by default, listening on all interfaces
 #   listen_address: ''
 
-#   # reload_handler_path defines the HTTP path where the Reload interface is available.
+#   # handler_path defines the HTTP path where the Reload interface is available.
```
jranson: `reload_handler_path`, `reload_drain_timeout`, `reload_rate_limit` are the intended key names. The defect you found is actually in the struct tags in pkg/config/mgmt/mgmt.go.  As part of this patch, please adjust the struct tags to include the `reload_` prefix for clarity and revert the key changes here. Also please add the `reload_` prefix to the keys in deploy/kube/configmap.yaml. Thank you!

pkg/config/config.go
```diff
@@ -297,19 +319,18 @@ func (c *Config) IsStale() bool {
...
-		return false
-	}
-	return !t.Equal(c.Main.configLastModified)
+	return c.hasConfigChanged()
 }
 
 // CheckAndMarkReloadInProgress checks if the config is stale and
 // marks it as being reloaded to prevent duplicate reloads.
 func (c *Config) CheckAndMarkReloadInProgress() bool {
+	if c == nil || c.Main == nil {
```
jranson: you can add the check for `|| c.Main.configFilePath == ""` here to return false before claiming the lock, then just check for `time.Now().Before(c.Main.configRateLimitTime)` after claiming the lock.

jranson (changes requested):

pkg/config/config.go
houyuwushang: Moved the empty config path check into the early return and left only the rate-limit check under `stalenessCheckLock`.

examples/conf/example.full.yaml
houyuwushang: Updated the struct tags to `reload_handler_path`, `reload_drain_timeout`, and `reload_rate_limit`, and restored those names in both examples. I also updated the config fixtures and reload-storm config so they continue exercising the intended keys and non-default values.

houyuwushang:
@jranson, could you re-review this when you get a chance? The requested configuration-key and empty-path changes are addressed, and the branch is rebased onto current `main`.

jranson:
We're targeting this for the v2.1.0 release, so I'll give it a final review once v2.0.5 release (which includes the new aggregation items from #1068) lands

houyuwushang:
@jranson v2.0.5 is out now. I rebased this onto current `main` and resolved the release-line conflicts, so it should be ready for the final review you mentioned.

jranson:
@houyuwushang we're ready to re-review this once the merge conflicts are resolved

houyuwushang:
Resolved the conflicts against current main. The polling check now uses #908's configuration-source fingerprint, so changes to included files and config directories are covered too.

jranson (approved):

Issue #1072, opened 2026-07-17
support promql limitk in ALB TSM


jranson:
# Support first-visited `limitk` compatibility in ALB TSM fanout

Parent: #1068

## Summary

Add global PromQL `limitk(k, v)` handling to ALB Time Series Merge (TSM)
using the selection algorithm in the current Prometheus evaluator: for every
aggregation group and evaluation timestamp, retain the first `k` input samples
encountered while visiting the inner expression's result matrix.

Prometheus documentation describes the subset as deterministic
pseudo-random, but the current implementation does not assign a label-hash
score or rank all candidates. It scans the input matrix in its existing order,
appends samples until each group has `k`, and then may stop scanning early.

Prometheus requests selector series from storage without requiring them to be
sorted. Consequently, there is no public canonical visitation order that a
federated query layer can reconstruct to guarantee the identical label set a
separate global Prometheus deployment would have selected. Trickster will
reproduce the current first-visited algorithm over a stable, documented TSM
merged-series order. It will guarantee cardinality, grouping, repeatability for
the same merged input/order, and independence from goroutine completion order,
but will not claim identical selected-series identity across different storage
engines or physical shard layouts.

`limitk` is experimental in Prometheus. This compatibility contract should be
revisited if Prometheus defines a canonical selection order or changes its
implementation.

## Compatibility contract

- For each `(result, timestamp, aggregation group)`, scan the globally merged
  inner vector in stable TSM series order.
- Retain the first `min(k, group cardinality)` samples present at that
  timestamp.
- Use `by` and `without` only to form selection groups; selected samples retain
  their complete original PromQL labels and values.
- Include both float and native-histogram samples.
- Do not rank by value or by label hash.
- Do not promise the same selected label set as a separate global Prometheus
  instance whose storage layer supplies a different input order.
- Do guarantee that asynchronous fanout completion does not affect the merged
  order or selected result.

## Scope

- Support non-negative literal `k` values accepted by the target Prometheus
  compatibility version.
- Support `by (...)` and `without (...)`, before or after the aggregation
  arguments.
- Support instant and range queries.
- Preserve float samples, native-histogram samples, original labels, and sample
  values.
- Preserve common outer `sort` and `sort_desc` wrappers.
- Rewrite and globally merge the inner expression before selection whenever
  it contains an operation already supported by the TSM planner.
- Retain a warning for unsupported scalar parameter expressions or inner
  expressions whose cross-shard semantics TSM cannot evaluate correctly.

[...]

Issue #1069, opened 2026-07-17
support promql limit_ratio in ALB TSM


jranson:
# Support semantically correct `limit_ratio` in ALB TSM fanout

Parent: #1068

## Summary

Add semantically correct PromQL `limit_ratio(r, v)` support to the ALB Time
Series Merge (TSM) mechanism.

For a fixed ratio, Prometheus selects each input sample independently using a
deterministic hash of its complete label set. This selection is distributive
across shards: selecting on each shard and then unioning/deduplicating the
results is equivalent to selecting from the combined input vector. However,
the inner expression must itself have correct federated semantics. For example,
`limit_ratio(0.1, sum by (service) (...))` must globally merge the inner `sum`
before applying `limit_ratio`.

## Scope

- Support literal ratios in the full PromQL range `[-1, 1]`, including the
  complementary selection semantics of negative ratios.
- Support both `by (...)` and `without (...)` grouping forms, before or after
  the aggregation arguments.
- Support instant and range queries.
- Preserve float and native-histogram samples selected by the operator.
- Preserve common outer `sort` and `sort_desc` wrappers with the same range
  query behavior as Prometheus.
- Remove the current inaccurate-result warning for supported expressions.
- Retain a clear fallback warning for parameter expressions or nested query
  shapes that the TSM planner cannot evaluate globally.

## Implementation guidance

- [ ] Add a parsed aggregation specification for `limit_ratio` containing the
      ratio, inner query, grouping, and optional sort wrapper.
- [ ] Add a fast path for a fixed ratio over a shard-local inner expression:
      allow each Prometheus shard to apply `limit_ratio`, then use normal TSM
      union/deduplication.
- [ ] For an inner aggregation that TSM can merge, rewrite the request to the
      inner expression, perform its global TSM merge, and apply `limit_ratio`
      once in the finalizer.
- [ ] Make the finalizer use a Prometheus-compatible label hash and threshold:
      select `hash/maxUint64 < r` for non-negative `r`, and
      `hash/maxUint64 >= 1+r` for negative `r`.
- [ ] Hash the labels visible to PromQL after any configured TSM-only injected
      labels have been handled; do not let routing-only labels change the
      sample set.
- [ ] Treat unsupported or shard-dependent scalar parameter expressions
      conservatively. Do not silently evaluate different ratios per shard.
- [ ] Ensure partial fanout failure remains visible as an incomplete-global-
      result warning.

## Acceptance criteria

[...]

PR #1075, opened 2026-07-18
refactor: introduce TSM merge plans

### LINKED ISSUE #1073: refactor TSM w/ MergePlan
# Introduce `MergePlan` and refactor ALB TSM query execution

## Summary

Introduce a shared TSM query-planning model before adding support for the new PromQL outer aggregators in #1068. Replace the current classification tuple and weighted-average special case with a plan that describes fanout variants, per-variant accumulation, completeness requirements, reduction, finalization, and unsupported-query warnings.

Migrate the existing `avg -> sum + count` implementation to the generic plan executor as its first multi-variant user. Existing single-query merge behavior for deduplication, `sum`, `count`, `count_values`, `min`, `max`, and current `topk`/`bottomk` finalization must continue to work through single-variant plans.

This code is new and the package has no external consumers. No compatibility adapter or parallel legacy interface is required. Once the new path is covered and equivalent, remove `needsDualQuery`, `RewriteForWeightedAvg`, and the dedicated `serveWeightedAvg` execution path.

This issue is a prerequisite for the four operator-support sub-issues under #1068 and should merge to `main` before work on those implementations begins.

## Goals

- Give TSM one planning model for single-query and multi-query operations.
- Make fanout variant execution, contribution pairing, reduction, failure handling, metrics, capture limits, and finalization reusable by future aggregators.
- Use weighted `avg` to validate the plan abstraction without changing its successful-result semantics.
- Strengthen partial-failure correctness by preventing incomplete variant sets from contributing as though they were complete.
- Remove operator-specific dispatch flags and weighted-average fanout code.

## Universal planning requirement

Every time-series query handled by TSM must be represented by a `MergePlan`. Plans are not limited to rewritten, multi-query, or finalizable aggregations. This includes:

- queries with no outer aggregation, represented by one primary variant using deduplication;
- simple outer aggregations such as `sum`, `count`, `count_values`, `min`, and `max`, represented by one primary variant using the corresponding merge strategy;
- `group`, represented by one primary deduplication variant;
- existing rank/sort queries, represented by a rewritten primary variant plus the existing finalizer;
- weighted `avg`, represented by its sum and count variants;
- unsupported aggregations, represented by a deduplication fallback plan with an explicit inaccurate-result warning.

There must not be a parallel tuple/boolean classification path for "simple" queries. Planning is the uniform control plane for variant construction, merge strategy, completeness, reduction, finalization, warnings, response authority, and safe fast-path eligibility.

Uniform planning does not require every plan to use the full multi-variant executor. Execution may retain optimized data paths beneath the plan:

- a one-variant plan may use the existing standard fanout implementation;
- an unchanged one-variant request need not be cloned unnecessarily;
- a genuinely single-member pool may proxy directly when the plan requires no rewriting, injected-label stripping, reduction, finalization, or warning;
- a degraded multi-member pool must still use the planned response path so its incomplete-result warning is preserved.

Plan construction and validation, rather than separate request-time operator branches, must decide whether these optimizations are safe.

## Suggested planning model

The exact names and public shape are not prescribed, but the plan must express the following concepts:

```go
type MergePlan struct {
	// One request per logical query variant. Most plans have one; weighted avg
	// has sum and count variants.
	Variants []QueryVariant

	// Combines variant accumulators or per-member variant state after fanout.
	Reduction ReductionSpec

	// Optional operation applied after reduction/global accumulation.
	Finalizer FinalizerSpec

	// Determines whether a member may contribute if a required variant failed.
	Completeness CompletenessPolicy

	// Warning attached when the expression cannot be planned exactly.
	UnsupportedWarning string
}

type QueryVariant struct {
	// Stable observability name, such as "avg-sum" or "avg-count".
	Name string

	// A request rewritten by the provider and safe for concurrent fanout.
	Request *http.Request

	// Dataset strategy used to accumulate this variant's contributions.
	MergeStrategy int
}
```

An internal plan type, replacement backend-provider contract, reducer interface, or equivalent representation is acceptable. The important requirement is that the executor not switch on PromQL operator names or grow new booleans for future multi-query plans.

Every valid plan must satisfy these invariants:

- [ ] It contains at least one query variant.
- [ ] Variant names are unique and stable for observability.
- [ ] Every variant declares its merge strategy.
- [ ] Exactly one variant is designated as response authority.
- [ ] Standard reduction accepts the supported one-variant shape.
- [ ] Multi-variant reductions declare their member-completeness policy.
- [ ] An unsupported fallback plan carries a non-empty warning.
- [ ] Any single-member bypass is explicitly safe for that complete plan.

The plan should be created after the original request and query are known so the provider remains responsible for protocol-specific GET/POST rewriting and safe request cloning.

### Standard single-query plan

```text
variants:
  - primary: original or finalizer-rewritten request merge
  strategy: classified dedup/sum/min/max strategy
reduction:
  standard TSM accumulation
completeness:
  one usable primary contribution per participating member
finalizer:
  existing provider finalizer, when applicable
```

The existing standard fanout implementation may remain as an internal single-variant fast path, but it must be selected from a `MergePlan`, not from a separate classification/dispatch model.

### Weighted-average plan

```text
variants:
  - avg-sum:
      request: outer avg rewritten to sum
      merge strategy: sum
  - avg-count:
      request: outer avg rewritten to count
      merge strategy: sum
reduction:
  pair accumulated sum and count series/timestamps, then divide
completeness:
  require both variants from a member before either contribution from that member enters the global reduction
response authority:
  avg-sum variant, preserving current envelope/header behavior
finalizer:
  existing outer rank/sort finalizer, when applicable
```

### Plan construction

- [ ] Replace `ClassifyMerge(query) (strategy, needsDualQuery, warning)` with a plan-producing provider contract.
- [ ] Produce a plan for every TSM time-series request; no-aggregation, supported single-variant aggregation, and unsupported fallback queries must not bypass planning.
- [ ] Represent all existing supported merge strategies as single-variant plans where no additional query variant is required.
- [ ] Represent weighted `avg` as the two-variant plan described above.
- [ ] Move provider-specific request rewriting into plan construction or a provider-owned plan preparation step.
- [ ] Preserve the original query separately from rewritten variant queries so metadata, warnings, series pairing, and finalizers continue to use the correct logical statement.
- [ ] Ensure every variant request has an independently readable request body and no concurrently mutable request state.

### Generic execution

- [ ] Add a generic plan executor that can fan out one or more variants under the existing query concurrency and capture limits.
- [ ] Preserve deterministic member indexing for every variant.
- [ ] Collect contributions by `(variant, member)` before deciding which contributions are eligible for reduction.
- [ ] Apply the plan's completeness policy before merging contributions. For weighted `avg`, discard both contributions from a member unless both its sum and count variants are usable.
- [ ] Accumulate each variant with its configured merge strategy.
- [ ] Run the plan reducer only after all eligible variant contributions have been accumulated.
- [ ] Invoke the existing provider finalizer exactly once, after reduction.
- [ ] Keep cancellation checks around fanout, accumulation, reduction, and finalization.
- [ ] Preserve the all-fanout-failed `502` behavior and existing empty-result handling.

### Response and failure behavior

- [ ] Let the plan identify the response-authority variant used for status, headers, and response marshaling when variants differ.
- [ ] Preserve winner-header and multi-valued-header behavior from the current standard and weighted-average paths.
- [ ] Preserve partial-hit status reporting for mixed successful and failed members.
- [ ] Emit a clear warning naming the member and missing/failed variant when a completeness policy excludes that member.
- [ ] Never combine sum data from one member set with count data from a different member set and present it as a complete weighted average.
- [ ] Preserve degraded-pool warnings and the genuinely single-member fast path where it remains semantically safe.

### Observability and resource controls

- [ ] Preserve `ALBFanoutAttempts` and `ALBFanoutFailures` variant labels, including `avg-sum` and `avg-count`.
- [ ] Preserve panic recovery and failure-reason accounting for every variant and member.
- [ ] Preserve per-response and per-fanout capture-budget enforcement.
- [ ] Document whether a multi-variant plan's total capture budget is shared or applied independently per variant; do not change the current behavior accidentally during the refactor.
- [ ] Preserve configured query concurrency limits and avoid unbounded `variant count * member count` goroutine creation.

### Cleanup

- [ ] Remove the `needsDualQuery` flag and all associated branches.
- [ ] Remove `RewriteForWeightedAvg` from the provider contract and Prometheus backend.
- [ ] Remove or generalize `serveWeightedAvg`; no weighted-average-specific fanout executor should remain.
- [ ] Update comments that describe TSM as having a special dual-query average path.
- [ ] Keep weighted-average arithmetic helpers that remain useful as reducer implementation details.

## Acceptance criteria

- [ ] Every TSM time-series request is represented by a validated plan, including no-aggregation dedup, simple supported aggregations, rank/sort rewrites, weighted `avg`, and unsupported warning fallbacks.
- [ ] No legacy or parallel classification tuple controls request execution outside the plan.
- [ ] Existing successful `avg`, `sum`, `count`, `count_values`, `min`, `max`, dedup, `topk`, `bottomk`, and sort-wrapper tests continue to pass.
- [ ] Weighted `avg` results remain mathematically correct for unequal shard cardinalities in instant and range queries.
- [ ] GET and POST weighted-average rewriting remains equivalent.
- [ ] An `avg` member missing either its sum or count contribution is excluded as a pair and produces an explicit partial-result warning.
- [ ] Variant and member completion order cannot change the reduced result.
- [ ] Existing injected-label stripping, sparse timestamp pairing, unmatched series warnings, headers, status aggregation, cancellation, and capture limits remain covered.
- [ ] `needsDualQuery`, `RewriteForWeightedAvg`, and the dedicated weighted-average dispatch path no longer exist.
- [ ] Adding a future multi-variant operator does not require duplicating the fanout executor or adding an operator-specific dispatch boolean.

## Tests

- [ ] Convert existing classification tests to assert complete plan contents, including variant names, rewritten queries, strategies, reduction, and completeness policy.
- [ ] Retain the current weighted-average tests as characterization tests for successful results, GET/POST rewriting, grouping, wrappers, and sparse timestamps.
- [ ] Add plan validation tests for zero variants, duplicate variant names, missing response authority, invalid strategies, and incompatible reduction configuration.
- [ ] Add generic executor tests for one and multiple variants.
- [ ] Add member-level completeness tests covering missing sum, missing count, parse failure, transport failure, panic, cancellation, and capture-limit failure.
- [ ] Add deterministic-reduction tests with permuted variant completion and member completion order.
- [ ] Add response-authority, winner-header, multi-valued-header, mixed-status, all-failed, and empty-result tests.
- [ ] Run race-enabled tests for the TSM mechanism and Prometheus provider.
- [ ] Benchmark the plan executor against the current standard and weighted-average paths to detect material allocation or latency regressions.

## Code pointers

- [`pkg/backends/tsm_merge_provider.go`](https://github.com/trickstercache/trickster/blob/main/pkg/backends/tsm_merge_provider.go) defines the current classification and weighted-average rewrite contract.
- [`pkg/backends/prometheus/tsm_provider.go`](https://github.com/trickstercache/trickster/blob/main/pkg/backends/prometheus/tsm_provider.go) classifies PromQL queries and rewrites standard-finalizer and weighted-`avg` requests.
- [`pkg/backends/alb/mech/tsm/time_series_merge.go`](https://github.com/trickstercache/trickster/blob/main/pkg/backends/alb/mech/tsm/time_series_merge.go) contains standard fanout, `serveWeightedAvg`, contribution collection, deterministic batch merging, response selection, failure reporting, and finalizer invocation.
- [`pkg/timeseries/dataset/dataset.go`](https://github.com/trickstercache/trickster/blob/main/pkg/timeseries/dataset/dataset.go) contains `FinalizeWeightedAvg` and the dataset merge behavior used by the current reducer.
- [`pkg/backends/alb/mech/tsm/time_series_merge_serve_weighted_avg_test.go`](https://github.com/trickstercache/trickster/blob/main/pkg/backends/alb/mech/tsm/time_series_merge_serve_weighted_avg_test.go) provides weighted-average execution characterization coverage.
- [`pkg/backends/alb/mech/tsm/time_series_merge_weighted_avg_test.go`](https://github.com/trickstercache/trickster/blob/main/pkg/backends/alb/mech/tsm/time_series_merge_weighted_avg_test.go) covers weighted-average finalization and warning behavior.

## Non-goals

- Implementing `stddev`, `stdvar`, `quantile`, `limit_ratio`, or `limitk` in this issue. Those changes will be implemented later in separate Pull Requests.
- Changing successful PromQL aggregation semantics.
- Building a general recursive distributed PromQL planner in this refactor.
- Preserving the old provider interface or weighted-average executor after all in-repository callers and tests have migrated.
- Broad performance redesigns unrelated to making plan execution reusable.

houyuwushang:
## Description

Closes #1073.

This introduces one planning model for every TSM time-series request:

- replace the classification tuple and separate request rewriters with a validated `TSMMergePlan` covering variants, reduction, completeness, response authority, finalization, warnings, and single-member bypass safety
- keep the existing one-variant fanout as a plan-selected fast path, while multi-variant plans use a generic executor that collects by variant and member before accumulation
- migrate weighted `avg` to the generic executor and exclude both sides of a member when either `avg-sum` or `avg-count` is unusable, so reduction never combines different member sets
- share the configured query concurrency limit across variants while retaining independent per-variant fanout capture budgets and metric labels
- preserve response-authority headers/status, partial-hit reporting, cancellation, injected-label stripping, deterministic member order, finalizer ordering, empty responses, and all-authority-failed `502` behavior
- remove the legacy dual-query flag, weighted-average rewrite method, and dedicated weighted-average executor

The new operators from #1069-#1072 are intentionally not included here.

## Type of Change

- - [ ] Bug fix
- - [x] New feature
- - [ ] Optimization
- - [x] Test coverage
- - [ ] Documentation
- - [ ] Infrastructure

## AI Disclosure

## Tests

- affected package tests and race tests for `backends`, Prometheus planning, TSM, and fanout
- `golangci-lint` with the repository configuration
- multi-variant failure coverage for status, parse, transport, panic, cancellation, capture limits, degraded pools, response authority, injected labels, and completion-order determinism
- plan validation coverage and standard/two-variant executor benchmarks

### DIFF
--- a/pkg/backends/alb/mech/tsm/merge_plan.go
+++ b/pkg/backends/alb/mech/tsm/merge_plan.go
@@ -319,7 +319,8 @@ func (h *handler) collectPlanResult(
 		}
 	}
 	result.contrib = contribution
-	result.failed = contribution == nil
+	result.failed = contribution == nil || result.statusCode < http.StatusOK ||
+		result.statusCode >= http.StatusMultipleChoices
 	execution.contributions[member] = contribution
 	execution.results[member] = result
 }
--- a/pkg/backends/alb/mech/tsm/merge_plan.go
+++ b/pkg/backends/alb/mech/tsm/merge_plan.go
@@ -1,0 +1,465 @@
+/*
+ * Copyright 2018 The Trickster Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package tsm
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/http"
+	"strconv"
+
+	"github.com/trickstercache/trickster/v2/pkg/backends"
+	"github.com/trickstercache/trickster/v2/pkg/backends/alb/mech/fanout"
+	"github.com/trickstercache/trickster/v2/pkg/backends/alb/names"
+	"github.com/trickstercache/trickster/v2/pkg/backends/alb/pool"
+	"github.com/trickstercache/trickster/v2/pkg/encoding"
+	"github.com/trickstercache/trickster/v2/pkg/observability/logging"
+	"github.com/trickstercache/trickster/v2/pkg/observability/logging/logger"
+	"github.com/trickstercache/trickster/v2/pkg/observability/metrics"
+	"github.com/trickstercache/trickster/v2/pkg/proxy/handlers/trickster/failures"
+	"github.com/trickstercache/trickster/v2/pkg/proxy/headers"
+	"github.com/trickstercache/trickster/v2/pkg/proxy/request"
+	"github.com/trickstercache/trickster/v2/pkg/proxy/response/merge"
+	"github.com/trickstercache/trickster/v2/pkg/timeseries/dataset"
+	"golang.org/x/sync/errgroup"
+)
+
+func defaultTSMMergePlan(r *http.Request, query string) *backends.TSMMergePlan {
+	return &backends.TSMMergePlan{
+		OriginalQuery: query,
+		Variants: []backends.TSMQueryVariant{{
+			Name:              "primary",
+			Request:           r,
+			MergeStrategy:     int(dataset.MergeStrategyDedup),
+			ResponseAuthority: true,
+		}},
+		Reduction: backends.TSMReductionSpec{
+			Kind:          backends.TSMReductionStandard,
+			InputVariants: []string{"primary"},
+		},
+		Completeness:            backends.TSMCompletenessResponseAuthority,
+		AllowSingleMemberBypass: true,
+	}
+}
+
+func planNeedsLabelStripping(plan *backends.TSMMergePlan) bool {
+	if plan == nil || len(plan.Variants) > 1 {
+		return plan != nil
+	}
+	for _, variant := range plan.Variants {
+		if dataset.MergeStrategy(variant.MergeStrategy) != dataset.MergeStrategyDedup {
+			return true
+		}
+	}
+	return false
+}
+
+func (h *handler) servePlan(
+	w http.ResponseWriter,
+	r *http.Request,
+	hl pool.Targets,
+	rsc *request.Resources,
+	plan *backends.TSMMergePlan,
+	stripKeys []string,
+	finalizer mergeFinalizer,
+	warnMsg string,
+) {
+	if plan.Reduction.Kind == backends.TSMReductionStandard {
+		variant := plan.Variants[0]
+		if !plan.Finalizer.Enabled {
+			finalizer = nil
+		}
+		h.serveStandard(w, variant.Request, hl, rsc,
+			dataset.MergeStrategy(variant.MergeStrategy), stripKeys,
+			plan.Finalizer.Query, finalizer, warnMsg)
+		return
+	}
+	h.serveMultiVariantPlan(w, r, hl, rsc, plan, stripKeys, finalizer, warnMsg)
+}
+
+type planVariantExecution struct {
+	results       []gatherResult
+	contributions []*gatherContribution
+	fanoutResults []fanout.Result
+}
+
+// serveMultiVariantPlan applies one shared concurrency bound across variants.
+// MaxFanoutCaptureBytes remains independent per variant, matching the previous
+// weighted-average behavior; MaxCaptureBytes still applies to every response.
+func (h *handler) serveMultiVariantPlan(
+	w http.ResponseWriter,
+	r *http.Request,
+	hl pool.Targets,
+	rsc *request.Resources,
+	plan *backends.TSMMergePlan,
+	stripKeys []string,
+	finalizer mergeFinalizer,
+	warnMsg string,
+) {
+	parentCtx := r.Context()
+	memberCount := len(hl)
+	executions := make([]planVariantExecution, len(plan.Variants))
+	limiter := fanout.NewConcurrencyLimiter(
+		h.tsmOptions.ConcurrencyOptions.GetQueryConcurrencyLimit(),
+	)
+	dedupToleranceNanos := h.dedupToleranceNanos()
+
+	var eg errgroup.Group
+	for variantIndex := range plan.Variants {
+		variant := plan.Variants[variantIndex]
+		executions[variantIndex] = planVariantExecution{
+			results:       make([]gatherResult, memberCount),
+			contributions: make([]*gatherContribution, memberCount),
+		}
+
+		primed, err := fanout.PrimeBody(variant.Request)
+		if err != nil {
+			logger.Warn("tsm plan variant body preparation failure", logging.Pairs{
+				"variant": variant.Name, "error": err,
+			})
+			failures.HandleBadGateway(w, r)
+			return
+		}
+		plan.Variants[variantIndex].Request = primed
+
+		eg.Go(func() error {
+			execution := &executions[variantIndex]
+			fanoutResults, err := fanout.All(parentCtx, primed, hl, fanout.Config{
+				Mechanism:             names.MechanismTSM,
+				Variant:               variant.Name,
+				ConcurrencyLimiter:    limiter,
+				MaxCaptureBytes:       h.maxCaptureBytes,
+				MaxFanoutCaptureBytes: h.maxFanoutCaptureBytes,
+				Resources: func(int) *request.Resources {
+					return &request.Resources{
+						IsMergeMember:         true,
+						TSReqestOptions:       rsc.TSReqestOptions,
+						TSMergeStrategy:       variant.MergeStrategy,
+						TSDedupToleranceNanos: dedupToleranceNanos,
+					}
+				},
+				OnResult: func(member int, fr *fanout.Result) {
+					h.collectPlanResult(parentCtx, execution, member, fr, stripKeys, variant.Name)
+				},
+			})
+			execution.fanoutResults = fanoutResults
+			return err
+		})
+	}
+	if err := eg.Wait(); err != nil && parentCtx.Err() == nil {
+		logger.Warn("tsm plan gather failure", logging.Pairs{"error": err})
+	}
+	if parentCtx.Err() != nil {
+		return
+	}
+
+	authorityIndex, _ := plan.ResponseAuthority()
+	if allFanoutFailed(executions[authorityIndex].fanoutResults) {
+		failures.HandleBadGateway(w, r)
+		return
+	}
+
+	var responseSeed *dataset.DataSet
+	if !hasCompletePlanMember(plan, executions, memberCount) {
+		responseSeed = emptyResponseSeed(executions[authorityIndex].contributions)
+		if responseSeed == nil {
+			for variantIndex := range executions {
+				if variantIndex == authorityIndex {
+					continue
+				}
+				if responseSeed = emptyResponseSeed(executions[variantIndex].contributions); responseSeed != nil {
+					break
+				}
+			}
+		}
+	}
+	warnings, hasPlanFailure := applyPlanCompleteness(plan, executions, memberCount)
+	accumulators := make([]*merge.Accumulator, len(plan.Variants))
+	for variantIndex := range plan.Variants {
+		accumulators[variantIndex] = merge.NewAccumulator()
+		failedMembers := mergeGatherContributions(parentCtx, accumulators[variantIndex],
+			executions[variantIndex].contributions)
+		if len(failedMembers) > 0 {
+			for _, member := range failedMembers {
+				metrics.ALBFanoutFailures.WithLabelValues(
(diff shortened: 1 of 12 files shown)


pkg/backends/alb/mech/tsm/merge_plan.go
```diff
@@ -0,0 +1,465 @@
...
+			result.failed = true
+			execution.results[member] = result
+			return
+		}
+		if len(body) > 0 {
+			contribution = prepareGatherContribution(ctx, rsc, body, member, stripKeys)
+		}
+	}
+	result.contrib = contribution
+	result.failed = contribution == nil
```
jranson: because prometheus returns a valid response envelope even on 5xx errors (e.g, a 503), currently error status responses can potentially decode to a non-nil dataset and be included in the reduction, which could impact weighted averages and other operations. To protect against that, here you can check for a valid status code (2xx only) and consider the result failed if the code is not 2xx:

```
result.failed = contribution == nil || !result.statusCode >= 200 && result.statusCode < 300
```

If the result failed but has a non-nil DataSet, you can keep the decoded contribution temporarily so it can supply an error-envelope `responseSeed` and  `applyPlanCompleteness` will remove it from reduction.

It would be good to add a regression test using a non-nil error DataSet or Prometheus error envelope to verify this.

pkg/backends/tsm_merge_provider.go
```diff
@@ -16,32 +16,184 @@
...
+			return fmt.Errorf("tsm merge plan variant %q has no request", variant.Name)
+		}
+		if _, ok := requestPointers[variant.Request]; ok {
+			return fmt.Errorf("tsm merge plan variants share request %q", variant.Name)
+		}
+		requestPointers[variant.Request] = struct{}{}
+		// dataset.MergeStrategy currently defines values 0 through 5. Keep the
+		// shallow backends package independent of dataset while rejecting values
+		// that the merge layer would otherwise silently treat as deduplication.
+		if variant.MergeStrategy < 0 || variant.MergeStrategy > 5 {
```
jranson: It would be good to define in dataset/merge_strategy.go a const for `MaxMergeStrategyValue = MergeStrategyCount`, and use `dataset.MaxMergeStrategyValue` instead of a literal 5 here. that way if we add more merge strategies later, we can update the MaxMergeStrategyValue const assignment to the new largest MergeStrategy* const value without having to modify the actual validation logic.

pkg/backends/tsm_merge_provider_test.go
```diff
@@ -0,0 +1,136 @@
...
+			},
+			{
+				Name:          "avg-count",
+				Request:       testTSMRequest("count"),
+				MergeStrategy: 1,
+			},
+		},
+		Reduction: TSMReductionSpec{
+			Kind:          TSMReductionWeightedAverage,
+			InputVariants: []string{"avg-sum", "avg-count"},
```
jranson: While this works because `reducePlan` interprets the positions implicitly (e.g., `countAccumulator := accumulators[indexes[plan.Reduction.InputVariants[1]]]`), a plan like `InputVariants: []string{"avg-count", "avg-sum"}` (where the variants were inadvertently reversed) would currently pass validation even though the ordering is incorrect.

At minimum, we should document the positional contract in TSMReductionSpec and validate that weighted-average inputs use the expected variant strategies in the correct order. But the InputVariants slice will become even more brittle when we need to use three-input variance reductions for forthcoming reductions like stddev. One way to ensure the InputVariants positions are properly configured would be to add and use helper functions (e.g.,  `func TSMReductionWeightedAverageVariants() []string { return []string{"avg-sum", "avg-count"} }`)  anywhere the InputVariants slices need to be defined (in both production code and tests).  Also, it might be a good idea to define and use consts for the variant names like `avg-sum` instead of using the literals throughout the codebase, to avoid typos and other human error.

jranson (changes requested):
@houyuwushang  this refactor is solid and largely meets the requested criteria. just a few minor things to fix up.

pkg/backends/tsm_merge_provider_test.go
houyuwushang: Added weighted-average variant constants and TSMReductionWeightedAverageVariants(), and used them in provider and test plan construction. Validation now rejects reversed inputs and requires both variants to use MergeStrategySum.

pkg/backends/tsm_merge_provider.go
houyuwushang: Added dataset.MaxMergeStrategyValue and switched plan validation to use it. The new dependency exposed a test-only cycle through the SQL test package importing the broad testutil package, so I localized the two fixed time values used there.

pkg/backends/alb/mech/tsm/merge_plan.go
houyuwushang: Handled in 9d2d04db. Non-2xx results are now marked failed after decoding, so the DataSet remains available for response seeding while completeness removes that member from every reduction input. I added coverage for both a mixed pool and a no-complete-member 503 response.

pkg/backends/prometheus/tsm_provider.go
```diff
@@ -33,14 +34,112 @@ import (
...
+		switch agg {
+		case "sum", "count", "count_values":
+			strategy = int(dataset.MergeStrategySum)
+		case "avg":
+			return weightedAveragePlan(r, query, fanoutQuery, finalizer)
+		case "min":
+			strategy = int(dataset.MergeStrategyMin)
+		case "max":
+			strategy = int(dataset.MergeStrategyMax)
+		case "stddev", "stdvar", "quantile", "topk", "bottomk", "limitk", "limit_ratio":
```
jranson, line 66: should `topk` and `bottomk` be included in the unsupported case?

jranson: it looks like these are just for the inner aggregators, so the warning is appropriate.

jranson (approved):

PR #1077, opened 2026-07-18
feat: support limit_ratio in ALB TSM

### LINKED ISSUE #1069: support promql limit_ratio in ALB TSM

houyuwushang:
## Description

Closes #1069.

Adds literal `limit_ratio` planning and finalization for ALB TSM. Shard-local expressions keep the deterministic per-member fast path, while compatible inner aggregations are merged globally before applying the ratio once. Inner `sum` and `avg` retain an explicit fallback warning until TSM can reduce native histograms without losing shard contributions. The finalizer follows Prometheus label hashing and threshold behavior, strips configured routing labels before selection, preserves histogram/range semantics, and keeps explicit warnings for query shapes that cannot be merged safely.

Coverage includes parser and hash boundaries, positive/negative complements, GET and POST rewrites, supported nested aggregation plans, multi-shard native-histogram fallbacks for `sum` and `avg`, HA duplicates, pool ordering, partial fanout, and capture limits.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [x] Test coverage
- [ ] Documentation
- [ ] Infrastructure

## AI Disclosure

### DIFF
--- a/pkg/backends/prometheus/tsm_provider.go
+++ b/pkg/backends/prometheus/tsm_provider.go
@@ -42,6 +42,9 @@ func (c *Client) PlanTSMMerge(r *http.Request, query string) (*merge.TSMMergePla
 	if r == nil {
 		return nil, errors.New("cannot plan a nil request")
 	}
+	if spec, found := promql.ParseLimitRatioAggregation(query); found {
+		return c.planLimitRatio(r, query, spec)
+	}
 	fanoutQuery, rewritten := tsmInnerQuery(query)
 	finalizer := tsmFinalizer(query)
 
@@ -59,7 +62,7 @@ func (c *Client) PlanTSMMerge(r *http.Request, query string) (*merge.TSMMergePla
 		case aggregation.Sum, aggregation.Count, aggregation.CountValues:
 			strategy = int(merge.StrategySum)
 		case aggregation.Average:
-			return weightedAveragePlan(r, query, fanoutQuery, finalizer)
+			return weightedAveragePlan(r, query, fanoutQuery, finalizer, false)
 		case aggregation.Minimum:
 			strategy = int(merge.StrategyMin)
 		case aggregation.Maximum:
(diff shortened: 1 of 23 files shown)


pkg/backends/prometheus/tsm_provider.go
```diff
@@ -101,8 +104,110 @@ func (c *Client) PlanTSMMerge(r *http.Request, query string) (*merge.TSMMergePla
...
+	finalizer := merge.TSMFinalizerSpec{}
+	if spec.SortSet {
+		finalizer = merge.TSMFinalizerSpec{Enabled: true, Query: query}
+	}
+
+	if agg, aggregationInput, found := promql.CompleteOuterAggregation(spec.InnerQuery); found {
+		candidateStrategy := int(merge.StrategyDedup)
+		weightedAverage := false
+		switch agg {
+		case aggregation.Sum, aggregation.Count, aggregation.CountValues:
```
jranson, line 123: This marks inner `sum` and `avg` plans as exact, but TSM’s numeric merger does not aggregate native histograms: `aggregateValues` treats both histogram JSON values as non-numeric and keeps the first shard, while `FinalizeWeightedAvg` parses the histogram as NaN and replaces it with `"NaN"`. Thus queries such as `limit_ratio(-1, sum by (job) (native_histogram))` lose shard contributions, and the corresponding `avg` form can produce a malformed histogram value without an unsupported warning. Prometheus requires `sum` to add histograms and `avg` to add then divide them. Either implement histogram-aware reduction for these plans or retain an explicit unsupported fallback until it exists, with multi-shard histogram tests. Reference documentation: https://prometheus.io/docs/specs/native_histograms/

jranson (changes requested):
One compatibility finding that we need to address, otherwise looks good.

pkg/backends/prometheus/tsm_provider.go
houyuwushang: Good catch. I changed `sum` and `avg` back to the explicit unsupported path for now, since handling them correctly needs histogram-aware reduction. I left `count`/`count_values`/`min`/`max`/`group` on the exact plans and added multi-shard histogram cases for both operators, including the `NaN` regression.

jranson: Routing every inner `sum` and `avg` to the warning fallback avoids histogram corruption, but it also removes the required exact path for ordinary float samples. For example, `limit_ratio(-1, sum by (service) (requests))` now runs the complete aggregation independently on each shard and deduplicates the conflicting shard totals instead of returning their global sum. The supplied guidance explicitly requires globally merging supported inner aggregations and includes this nested-sum case in the tests. Because sample type is only known at execution time, satisfying both float and native-histogram semantics requires histogram-aware reduction (or equivalent runtime handling), rather than classifying all `sum`/`avg` expressions as unsupported.

houyuwushang: You're right - the fallback fixed the histogram case by breaking the float path. I moved that decision into the runtime merge: float sum/avg keep the exact reducers, native histograms merge count/sum/buckets (and avg divides the merged histogram), and mixed samples are dropped with the PromQL warning. The cross-shard nested-sum case is covered now.

jranson (approved):

PR #1080, opened 2026-07-19
feat: support pooled stddev and stdvar in ALB TSM

### LINKED ISSUE #1070: support promql stddev and stdvar in ALB TSM
# Support pooled `stddev` and `stdvar` in ALB TSM fanout

Parent: #1068

## Summary

Add semantically correct PromQL population variance and standard deviation for
federated ALB TSM queries. Independently calculated shard variances must not be
deduplicated or averaged. TSM must merge sufficient statistics for each
aggregation group and timestamp.

Use a numerically stable pooled-variance state containing count (`n`), mean
(`mean`), and the sum of squared deviations from the mean (`M2`). A shard can
provide this state using `count`, `avg`, and `stdvar`, where
`M2 = count * stdvar`.

## Scope

- Support `stdvar` and `stddev` over float samples.
- Match PromQL population variance semantics (`M2 / n`), not sample variance.
- Ignore native-histogram inputs consistently with Prometheus.
- Support `by (...)` and `without (...)`, in both accepted syntax positions.
- Surface partial multi-query or multi-shard failure clearly.

## Implementation guidance

- [ ] Generalize the weighted-`avg` dual-query path into a multi-variant TSM
      plan capable of issuing `count`, `avg`, and `stdvar` variants.
- [ ] Rewrite only the intended outer aggregation and preserve its inner
      expression and grouping exactly.
- [ ] Run the three variants concurrently, subject to existing fanout and
      capture concurrency limits.
- [ ] Pair `count`, `avg`, and `stdvar` results by shard, result, labels, and
      timestamp. Include a shard only when all required statistics are usable.
- [ ] Convert each shard result to `(n, mean, M2)` with
      `M2 = n * shardVariance`.
- [ ] Merge states in deterministic pool order using the Chan/Welford formula:

      ```text
      delta = meanB - meanA
      n     = nA + nB
      mean  = meanA + delta * nB / n
      M2    = M2A + M2B + delta*delta * nA*nB / n
      ```

- [ ] Finalize `stdvar` as `M2/n` and `stddev` as `sqrt(M2/n)`.
- [ ] Do not use `E[x^2] - E[x]^2`; it is vulnerable to catastrophic
      cancellation.
- [ ] Apply configured injected-label stripping consistently to all three
      variants before pairing and pooling.
- [ ] Preserve Prometheus behavior for empty groups, singleton groups, `NaN`,
      and infinities.
- [ ] If any required variant is incomplete, do not combine statistics from
      different shard sets; warn or fail according to the TSM partial-result
      policy.

## Acceptance criteria

- [ ] Results match a global Prometheus-compatible reference within an
      explicitly documented floating-point tolerance.
- [ ] Unequal shard cardinalities produce the true pooled population variance,
      not an average of shard variances.
- [ ] Large-offset/small-variance inputs remain numerically stable.
- [ ] Singleton groups return zero variance and zero standard deviation.
- [ ] `by` and `without` work for instant and range queries.
- [ ] Histogram samples are ignored consistently and do not corrupt float
      aggregation groups.
- [ ] Supported queries no longer include the inaccurate-result warning.
- [ ] Missing count, mean, or variance data cannot silently produce a complete-
      looking result.

## Tests

- [ ] Parser and rewrite tests for both operators, both grouping forms, GET and
      POST requests, and sort wrappers.
- [ ] Pooled-state unit tests against a direct Welford calculation.
- [ ] Tests for unequal shard sizes, zero variance, singleton groups, negative
      values, very large values, `NaN`, and infinities.
- [ ] Instant and range tests with sparse/misaligned timestamps.
- [ ] Pairing tests for missing series, missing timestamps, and one failed
      variant on one shard.
- [ ] HA replica deduplication and injected-label stripping tests.
- [ ] Benchmarks covering many shards, groups, and range-query steps.

## Prometheus references

- [`stddev` operator documentation](https://prometheus.io/docs/prometheus/latest/querying/operators/#stddev)
  defines population standard deviation, float-only behavior, and histogram
  handling.
- [`stdvar` operator documentation](https://prometheus.io/docs/prometheus/latest/querying/operators/#stdvar)
  defines population variance and its float/histogram behavior.
  contains the `STDVAR`/`STDDEV` Welford update and final population
  calculation. It is the compatibility reference for `NaN`, infinity,
  histogram, grouping, and output-label behavior.
- [Prometheus aggregation operator documentation](https://prometheus.io/docs/prometheus/latest/querying/operators/#aggregation-operators)
  defines the shared `by`/`without` grammar and label-retention rules.
- [Prometheus query-range semantics](https://prometheus.io/docs/prometheus/latest/querying/basics/#range-queries)
  explains that a range query repeatedly evaluates the expression at its
  configured steps; the pooled state must therefore be independent per
  timestamp.

## Non-goals

- Bit-for-bit equality with a single Prometheus execution. Pooling changes the
  floating-point operation order, although it preserves the mathematical
  population statistic.
- Supporting a nested inner expression whose own cross-shard semantics cannot
  be planned by TSM.

houyuwushang:
## Description

Closes #1070.

Adds exact ALB TSM planning for outer `stdvar` and `stddev`. Each pool member returns float-only `count`, `avg`, and `stdvar` variants; TSM pairs them by member, result, labels, and timestamp, then combines `(n, mean, M2)` states in deterministic pool order with the Chan/Welford formula. Incomplete triples are excluded with partial-result warnings, injected labels are stripped before pairing, and response extents are merged only from included members.

With the replica-group support from #1088 now on `main`, pooled variance first coalesces replicas within each group and then combines the independent group contributions. This preserves HA deduplication without dropping distinct shards.

The PromQL rewrite supports `by` and `without` in both syntax positions, preserves metric metadata labels, handles instant/range queries and outer `sort` wrappers, and ignores native histograms through a float-only input filter. Compatible nested aggregations are finalized globally. Nested `sum` and `avg` now reuse the exact histogram-aware sum and weighted-average plans before the outer variance finalizer runs.

Coverage includes unequal shard sizes, a documented `1e-12` relative tolerance against direct Welford calculation, large offsets, singleton and special values, sparse timestamps, partial variants, replica groups, metadata grouping, GET/POST rewrites, request cancellation, extents, and multi-shard serving benchmarks.

Validation after rebasing onto current `main`:

- `go test ./pkg/backends/prometheus/... ./pkg/backends/alb/mech/tsm ./pkg/timeseries/dataset ./pkg/timeseries/merge -count=20`
- `go test -race ./pkg/backends/prometheus/... ./pkg/backends/alb/mech/tsm ./pkg/timeseries/dataset ./pkg/timeseries/merge -count=5`
- `go test -run '^$' ./...`
- `go vet ./pkg/backends/prometheus/... ./pkg/backends/alb/mech/tsm ./pkg/timeseries/dataset ./pkg/timeseries/merge`

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [x] Test coverage
- [ ] Documentation
- [ ] Infrastructure

## AI Disclosure

### DIFF
--- a/pkg/backends/alb/mech/tsm/merge_plan.go
+++ b/pkg/backends/alb/mech/tsm/merge_plan.go
@@ -195,9 +195,13 @@ func (h *handler) serveMultiVariantPlan(
 		}
 	}
 	applyPlanCompleteness(plan, executions, memberCount)
-	applyPlanPointCompleteness(plan, executions, memberCount)
+	pointCompleteness := applyPlanPointCompleteness(plan, executions, memberCount)
 	var warnings []string
-	var hasPlanFailure bool
+	if plan.Reduction.Kind == tsmerge.TSMReductionPooledVariance {
+		warnings = pooledVarianceCompletenessWarnings(
+			pointCompleteness, hl, configured)
+	}
+	hasPlanFailure := len(warnings) > 0
 	for variantIndex := range plan.Variants {
 		logical, groupWarnings, groupFailure := coalesceReplicaContributions(
 			parentCtx, hl, configured, executions[variantIndex].contributions,
@@ -205,7 +209,7 @@ func (h *handler) serveMultiVariantPlan(
 		executions[variantIndex].contributions = logical
 		warnings = append(warnings, groupWarnings...)
 		if variantIndex == authorityIndex {
-			hasPlanFailure = groupFailure
+			hasPlanFailure = hasPlanFailure || groupFailure
 		}
 	}
 	var responseAccumulator *merge.Accumulator
--- a/pkg/backends/alb/mech/tsm/merge_plan.go
+++ b/pkg/backends/alb/mech/tsm/merge_plan.go
@@ -198,7 +198,6 @@ func (h *handler) serveMultiVariantPlan(
 	applyPlanPointCompleteness(plan, executions, memberCount)
 	var warnings []string
 	var hasPlanFailure bool
-	accumulators := make([]*merge.Accumulator, len(plan.Variants))
 	for variantIndex := range plan.Variants {
 		logical, groupWarnings, groupFailure := coalesceReplicaContributions(
 			parentCtx, hl, configured, executions[variantIndex].contributions,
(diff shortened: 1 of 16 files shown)


jranson:
@houyuwushang if you can rebase this back to the base repo's `main` branch and resolve the merge conflicts here, I can start the review.

jranson:
@houyuwushang currently there is an issue in this PR (and likely other places) where it is impossible to know if data from a given shard is a replica of another shard vs separate/disjointed data. The HA/deduplicate determination is made based on whether data at a given timestamp is identical. So when multiple shards have identical data at a given timestamp, those are currently de-duplicated even when the shards are not actually HA replicas and just coincidentally have identical values.

We need to be able to support HA and non-HA shards as appropriate. This is a bigger problem than just this PR. I will be opening a separate PR shortly to address this, which will allow operators to indicate a replica group id for TSM pool backends, which will allow features like this to determine, based on the source backend replica ID, whether or not identical data in multiple shards should be treated as HA and de-duplicated vs treated as disjointed and merged. Once that PR is merged, we'll need to update this PR to properly handle both use cases.

Update: Once PR #1088 is merged, we can revisit this one for corresponding modifications.

houyuwushang:
Got it. I'll wait for #1088 to merge, then rebase this and update pooled variance so replicas are deduplicated only within a replica group, while identical states from different groups are still pooled separately. I'll make sure the fallback and incomplete-member cases follow the same grouping semantics.

jranson:
@houyuwushang with #1088 merged, this and the other aggregation issues are now ready for re-work

jranson (approved):

PR #1082, opened 2026-07-20
support prom scalars in ALB Time Series Merge

### LINKED ISSUE #1081: ALB fanout can return `NaN` for Prometheus scalar queries
_Note: This issue draft was generated with AI assistance and reviewed against observed query responses and the relevant Trickster source paths._

## Summary

When a Prometheus query whose top-level result is a scalar (for example, `scalar(count(...))`) is sent through an ALB-backed Trickster backend, Trickster can return `NaN` even though one or more ALB members return a valid scalar value.

The issue is particularly visible for an instant query. A range query can also contain a large number of duplicate or otherwise non-equivalent scalar fragments instead of one scalar value per evaluation timestamp.

## Observed behavior

For an ALB-backed Mimir frontend query, the upstream members returned a mix of valid scalar values and `NaN` values.

- Trickster instant query: `NaN`.
- Trickster range query: 3,920 returned values for the same window; 547 were finite and the rest were `NaN`.

This makes the ALB result unsuitable for callers that expect Prometheus scalar semantics: a scalar aggregation should yield one result for an instant query and one value per requested step for a range query.

## Why `NaN` is returned

The ALB backend fans the request out to multiple origins. Each origin evaluates the scalar expression independently. Origins with no matching input series may legitimately return a scalar `NaN`, while other origins return finite values.

Trickster recognizes `resultType: "scalar"` and converts it to an internal vector-like series representation for processing. However, the ALB merge/selection path appears to treat those independent per-origin scalar results as ordinary compatible time-series fragments. That permits a `NaN` member response to be selected for an instant query, and permits many member fragments to be retained in a range response. The final response therefore does not represent a single global evaluation of the scalar expression.

## Expected behavior

Trickster should not select an invalid scalar (`NaN`) when at least one healthy ALB member returned a finite scalar, and it should not expose per-origin scalar fragments as though they were a single Prometheus scalar result.

At minimum, the response behavior should be deterministic and documented for scalar results through ALB backends.

## Suggested fix

Treat `resultType: "scalar"` as a distinct ALB merge case rather than normal series data:

- Preserve scalar result type and provenance through the fanout/merge pipeline.
- For instant queries, exclude `NaN` candidates when finite candidates exist, then apply an explicit and documented selection policy (for example, the ALB mechanism's chosen origin).
- For range queries, return at most one scalar per requested evaluation timestamp; do not concatenate independent origin scalar samples into one matrix.
- If global aggregation across origins is intended, re-evaluate or aggregate with semantics appropriate for the PromQL expression. Do not infer global `count` semantics merely by merging independently evaluated scalar outputs.
- Add ALB integration tests covering mixed finite/`NaN` scalar responses for both `/query` and `/query_range`.

## Relevant implementation area

Prometheus scalar responses are recognized in `pkg/backends/prometheus/model/timeseries.go`, where scalar results are transformed into a vector-like internal representation. The ALB merge path needs to retain the scalar distinction when selecting or combining member responses.

jranson:
## Description
Resolves #1081 by carrying the data type (vector vs scalar) in the internal dataset type and adding a `scalar` merge strategy to TSM.

## Type of Change

- - [x] Bug fix
- - [ ] New feature
- - [ ] Optimization
- - [x] Test coverage
- - [ ] Documentation
- - [ ] Infrastructure

## AI Disclosure


pkg/backends/prometheus/promql/scalar.go
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+package promql
+
+import "strings"
+
+// IsScalarCall reports whether query is a top-level call to PromQL's scalar
+// conversion function. TSM uses this before fanout so range-query responses,
+// which Prometheus encodes as matrices, retain scalar selection semantics.
+func IsScalarCall(query string) bool {
```
houyuwushang: Parentheses do not change a PromQL expression's result type, but this only recognizes the literal `scalar(...)` shape. I reproduced `(scalar(count(up)))` selecting `dedup` in `PlanTSMMerge`; with the default exact-epoch dedup, a later member's `NaN` can then replace an earlier finite value. Could this classify the expression result (or at least unwrap grouping) instead? Scalar expressions such as `scalar(count(up)) + 1` otherwise have the same gap.

pkg/backends/prometheus/model/vector.go
```diff
@@ -17,16 +17,121 @@
...
+func MergeAndWriteVectorBatchMergeFunc() merge.BatchMergeFunc {
+	standard := merge.TimeseriesBatchMergeFunc()
+	return func(accum *merge.Accumulator, items []merge.BatchItem) (bool, error) {
+		if len(items) == 0 {
+			return false, nil
+		}
+		candidates := make([]*dataset.DataSet, len(items))
+		for i, item := range items {
+			ds, ok := item.Data.(*dataset.DataSet)
+			if !ok || ds == nil || ds.SourceResultType != string(Scalar) {
```
houyuwushang: Prometheus error envelopes unmarshal to a non-nil `DataSet` with an empty `SourceResultType`, so one failed member sends the whole batch through the standard vector merge. I reproduced member 0 returning a 500 error envelope and member 1 returning scalar `42`: TSM returned HTTP 200 with `resultType:"vector"` and the failed member's error fields. Since mixed 2xx/non-2xx fanout is treated as a partial success, can this keep/select scalar candidates instead of falling back because of the error dataset?

houyuwushang (commented):
I found two cases where a scalar request still falls back to the previous dedup/vector behavior. Details inline.

houyuwushang (commented):
The two cases from my previous review are fixed. I found one remaining scalar-classification gap inline.

pkg/backends/prometheus/promql/scalar.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	if inner, ok := unwrapUnaryFunction(q, "scalar"); ok {
+		return strings.TrimSpace(inner) != ""
+	}
+	for _, name := range []string{"time", "pi"} {
+		if inner, ok := unwrapUnaryFunction(q, name); ok {
+			return strings.TrimSpace(inner) == ""
+		}
+	}
+	if _, err := strconv.ParseFloat(q, 64); err == nil {
```
houyuwushang: PromQL accepts hexadecimal and duration literals as scalars. With Prometheus v0.313.1, `scalar(count(up)) + 0x2` and `scalar(count(up)) + 1h` both parse as `ValueTypeScalar`; here `IsScalarExpression` returns false, so `PlanTSMMerge` selects dedup (0) rather than scalar (6). A later member's `NaN` can therefore replace an earlier finite range value, reopening #1081 for these valid expressions. Could this literal check cover PromQL's accepted numeric/duration forms and add the two planner cases?

LimitlessEarth (commented):
I have surface reviewed the entire change(code change and tests look in order but I have not tested it), but looked more closely at the commit since houyuwushang reviewed. LGTM

PR #1083, opened 2026-07-21
feat: support exact quantile in ALB TSM

### LINKED ISSUE #1071: support promql quantile in ALB TSM
# Support exact `quantile` finalization in ALB TSM fanout

Parent: #1068

## Summary

Add semantically correct PromQL `quantile(phi, v)` support by evaluating the
outer quantile once over the globally merged inner vector.

Per-shard quantiles cannot be combined into an exact global quantile, and no
fixed-size exact summary exists for arbitrary input values. Exact support must
therefore make the relevant inner float samples available to the TSM
finalizer. This can substantially increase response size and memory use and
must remain subject to existing capture limits.

## Scope

- Support literal quantile parameters, including Prometheus edge-case behavior
  for `NaN`, values below zero, and values above one.
- Support `by (...)` and `without (...)`, before or after the arguments.
- Ignore native-histogram samples consistently with Prometheus aggregation
  semantics.
- Remove the inaccurate-result warning only for expressions whose inner query
  TSM can merge correctly.

## Implementation guidance

- [ ] Parse an outer `quantile` into its parameter, inner query, grouping, and
      optional sort wrapper.
- [ ] Rewrite fanout requests to the inner expression.
- [ ] Classify and merge the rewritten inner expression before quantile
      finalization, so cases such as
      `quantile(0.9, sum by (service, instance) (...))` use globally merged
      inner values.
- [ ] Group float samples by result, timestamp, and PromQL `by`/`without`
      label set.
- [ ] Remove the metric name and construct output labels exactly as a PromQL
      aggregation does.
- [ ] Implement Prometheus's exact value quantile rule: sort values, compute
      rank `phi * (N - 1)`, and linearly interpolate between the adjacent
      values.
- [ ] Match Prometheus ordering and special handling for `NaN`, `-Inf`, and
      `+Inf`; ignore histogram samples.
- [ ] Initially prefer a clear sort-based implementation. Consider selection
      of the two adjacent ranks only after profiling demonstrates a need.
- [ ] Where practical, finalize one timestamp at a time to bound temporary
      memory. Avoid a permanent `(group, timestamp) -> all values` copy in
      addition to the already merged dataset.
- [ ] Enforce per-member and total fanout capture limits and return a clear
      failure rather than degrading to shard quantiles or deduplication.
- [ ] Preserve incomplete-fanout warnings because a quantile over a subset is
      not the requested global quantile.

## Acceptance criteria

- [ ] Supported queries match a single global Prometheus-compatible query
      layer for the same samples.
- [ ] Interpolated quantiles match for even and odd input cardinalities.
- [ ] Empty groups, singleton groups, duplicate values, `NaN`, and infinities
      match Prometheus behavior.
- [ ] `by` and `without` work for instant and range queries, including sparse
      timestamps.
- [ ] Histogram samples are ignored without affecting float results.
- [ ] Outer `sort` and `sort_desc` behave correctly for instant queries and
      preserve the existing range-query warning behavior.
- [ ] Capture-limit and partial-fanout failures never return a silently
      approximate or apparently complete quantile.
- [ ] Supported queries no longer include the inaccurate-result warning.

## Tests

- [ ] Parser/rewrite tests for prefix/postfix grouping, GET/POST, wrappers,
      malformed queries, and unsupported scalar expressions.
- [ ] Quantile algorithm table tests copied from Prometheus-compatible edge
      cases.
- [ ] Multi-shard tests where every shard has a very different distribution,
      proving that shard-quantile merging is not used.
- [ ] Instant and range tests with multiple groups and sparse points.
- [ ] Histogram, `NaN`, infinity, and duplicate-value tests.
- [ ] Capture-budget and partial-failure tests.
- [ ] Benchmarks for high-cardinality vectors and long range queries.

## Prometheus references

- [`quantile` operator documentation](https://prometheus.io/docs/prometheus/latest/querying/operators/#quantile)
  defines float-only operation, `NaN` ordering, out-of-range parameter
  behavior, grouping, and histogram handling.
- [Prometheus value-quantile implementation (`quantile.go`)](https://raw.githubusercontent.com/prometheus/prometheus/main/promql/quantile.go)
  contains the exact sort, `phi * (N - 1)` rank, adjacent-value interpolation,
  and special-case logic that the TSM finalizer should reproduce.
  shows how values are collected per group and timestamp, how histogram
  samples are ignored, and where the value-quantile helper is invoked.
- [Prometheus aggregation operator documentation](https://prometheus.io/docs/prometheus/latest/querying/operators/#aggregation-operators)
  defines the shared `by`/`without` grammar and output-label rules.
- [Prometheus query-range semantics](https://prometheus.io/docs/prometheus/latest/querying/basics/#range-queries)
  describes range queries as repeated expression evaluation, requiring an
  independent global quantile at every step.

## Non-goals

- Approximate sketches such as t-digest, KLL, or merging shard-local
  quantiles. Those may be useful as a separately configured feature but do not
  preserve exact PromQL semantics.
- Guaranteeing bounded network volume independent of the inner-vector
  cardinality; that is mathematically incompatible with exact arbitrary
  quantiles.
- Supporting an inner expression whose own cross-shard operations cannot be
  planned by TSM.

houyuwushang:
## Description

Closes #1071.

Adds exact global `quantile(phi, v)` handling to ALB TSM. Supported queries fan out the inner vector, apply the correct merge strategy to supported inner aggregations, ignore native histograms in the outer quantile, and calculate the Prometheus sort-and-interpolate quantile once per group and timestamp. Nested `sum` and `avg` reuse the histogram-aware exact reducers already on `main` before quantile finalization. Finalizer input detection also recognizes the weighted-sum statement retained after an inner `avg`; the shared path covers nested `stddev` and `stdvar` as well.

The finalizer sweeps one timestamp at a time with a cursor heap, so it does not retain a second all-values copy for the full range. Unsupported cross-shard expression shapes keep the existing warning fallback, and sorted fallbacks are not aggregated twice.

The parser accepts the same literal scalar forms as PromQL, including hexadecimal integers and duration literals, while leaving non-literal scalar expressions on the established fallback. Prefix/postfix grouping, GET/POST requests, sparse ranges, NaN, infinities, out-of-range values, and outer sort wrappers are covered.

After #1080 merged, this branch was replayed onto current `main`. The request-flow coverage now uses explicit #1088 replica groups and data where failing to coalesce an HA replica changes the expected median, while independent shards remain part of the global distribution. Partial fanout warnings and injected-label stripping remain covered.

Validation:

- `go test ./pkg/backends/prometheus/... ./pkg/backends/alb/mech/tsm -count=20`
- `go test -race ./pkg/backends/prometheus/... ./pkg/backends/alb/mech/tsm -count=5`
- `go test -run '^$' ./...`
- `go vet ./pkg/backends/prometheus/... ./pkg/backends/alb/mech/tsm ./pkg/timeseries/...`

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [x] Test coverage
- [x] Documentation
- [ ] Infrastructure

## AI Disclosure


pkg/backends/prometheus/tsm_quantile_finalize.go
```diff
@@ -0,0 +1,202 @@
...
+
+func (h *quantileCursorHeap) Pop() any {
+	old := *h
+	last := len(old) - 1
+	value := old[last]
+	*h = old[:last]
+	return value
+}
+
+func finalizeQuantileAggregation(ds *dataset.DataSet, spec promql.QuantileAggregation) {
```
jranson, line 71: It looks like this function is not running for queries like `quantile(..., avg(...))` so the quantile never happens, because the rewritten `sum(..))` queries are still being carried.

An easy fix would be in the Prometheus finalization layer: when determining whether the dataset contains the inner query, recognize that a completed weighted-average reduction still carries the rewritten sum(...) statement. For an inner outer-avg, `dataSetContainsQueryStatement` can check both:
 - The logical inner query: avg by (...) (...)
 - Its weighted-average sum variant: sum by (...) (...)

Then quantile finalization proceeds over the already-reduced averages. This should be implemented as a shared Prometheus helper because nested `stddev`/`stdvar` use the same statement-detection logic and may have the same metadata mismatch (it's fine to fix those aggregators in this PR).

pkg/backends/prometheus/tsm_quantile_finalize_test.go
```diff
@@ -0,0 +1,270 @@
...
+		want   float64
+	}{
+		{name: "empty", phi: 0.5, want: math.NaN()},
+		{name: "nan parameter", phi: math.NaN(), values: []float64{1}, want: math.NaN()},
+		{name: "below range", phi: -0.1, values: []float64{1}, want: math.Inf(-1)},
+		{name: "above range", phi: 1.1, values: []float64{1}, want: math.Inf(1)},
+		{name: "singleton", phi: 0.5, values: []float64{7}, want: 7},
+		{name: "odd", phi: 0.5, values: []float64{5, 1, 3}, want: 3},
+		{name: "even interpolation", phi: 0.5, values: []float64{4, 1, 3, 2}, want: 2.5},
+		{name: "upper interpolation", phi: 0.9, values: []float64{0, 10, 20, 30},
```
jranson, line 44: depending on the version of go and the chipset, the function may return just '27' instead of '27.000000000000004', so we need to make sure it is tolerant via an appropriate floating point comparison rather than exact equality match.

jranson (changes requested):
@houyuwushang just a few minor fixes inline

pkg/backends/prometheus/tsm_quantile_finalize.go
houyuwushang: Fixed. The finalizer now accepts both the logical `avg(...)` statement and the weighted `sum(...)` statement retained by the reducer. I moved that check into the shared Prometheus path and added regressions for `quantile(avg(...))`, `stddev(avg(...))`, and `stdvar(avg(...))`.

pkg/backends/prometheus/tsm_quantile_finalize_test.go
houyuwushang: Fixed. Finite results now use a relative tolerance; NaN and infinities are still handled explicitly.

jranson (approved):

PR #1084, opened 2026-07-21
feat: support global PromQL limitk in ALB TSM

### LINKED ISSUE #1072: support promql limitk in ALB TSM

houyuwushang:
Closes #1072.

## Description

Adds literal `limitk` planning and global finalization for ALB TSM. Supported inner expressions are fanned out without the outer selector, merged with the existing TSM strategy, and then limited once per timestamp and aggregation group in stable TSM label order.

The finalizer treats float and native-histogram representations with the same labels as one logical candidate, preserves selected labels and values, and keeps selection independent of fanout completion and pool order. Unsupported scalar parameters and inner expressions retain the existing inaccurate-result fallback. The ALB documentation records the stable-order contract and why selected labels can differ from another Prometheus storage order.

After #1083 merged, this branch was replayed onto current `main`. It now uses the shared finalizer-input detection for nested weighted averages, so the retained `sum(...)` statement does not prevent `limitk(avg(...))` from being finalized.

Validation:

- `go test ./pkg/backends/prometheus/... ./pkg/backends/alb/mech/tsm -count=20`
- `go test -race ./pkg/backends/prometheus/... ./pkg/backends/alb/mech/tsm -count=5`
- `go vet ./pkg/backends/prometheus/... ./pkg/backends/alb/mech/tsm ./pkg/timeseries/...`
- `go test -run '^$' ./...`

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [x] Test coverage
- [x] Documentation
- [ ] Infrastructure

## AI Disclosure


jranson:
@houyuwushang I can review this once merge conflicts against main are resolved. Thank you!

houyuwushang:
@jranson Resolved against current main; the branch is ready for review.

pkg/backends/alb/mech/tsm/time_series_merge_limitk_test.go
```diff
@@ -0,0 +1,307 @@
...
+}
+
+func newLimitKPool(specs []limitKMemberSpec, recorder *queryRecorder) pool.Pool {
+	targets := make(pool.Targets, len(specs))
+	for i, spec := range specs {
+		status := &healthcheck.Status{}
+		status.Set(healthcheck.StatusPassing)
+		backend := &pooledVarianceBackend{
+			stripKeysStubBackend: stripKeysStubBackend{
+				cfg: &bo.Options{Prometheus: &prop.Options{
```
jranson: here in the options pointer, it is recommended to set the `Name`, and provide a `ReplicaGroup` value.

jranson (changes requested):
@houyuwushang limitk implementation looks good. One minor testing gap is that none of the unit tests for limitk actually exercise replica groups, since a ReplicaGroup value is never set on the backends created in the `newLimitKPool` function, and the `spec.replica` field is really serving to populate the backendName. That means every backend defaults to being in its own single-member replica group named by the backend, bypassing any replica group collapsing for limitk. I recommend adding a `replicaGroup` field to `limitKMemberSpec` struct to allow you to layout different arrangements of 1 or more backends across 2 or more replica groups in a series of replica group topology tests, then verify the results properly collapse replica groups before merging data across replica groups. consider renaming `spec.replica` to `spec.backendName` for clarity.

pkg/backends/alb/mech/tsm/time_series_merge_limitk_test.go
houyuwushang: Good catch. I split the fixture into separate `backendName` and `replicaGroup` fields and now set both on each backend. The new topology cases verify that same-group replicas are collapsed before independent shards are summed, including mixed member ordering and a failed-member fallback.

jranson (approved):

PR #1088, opened 2026-07-24
feat: support TSM HA, Federation and Hybrid via Replica Groups

### LINKED ISSUE #1087: tsm: support configurable HA replica shards vs disjointed shards
Trickster currently lacks the ability to declare/determine whether two or more ALB TSM pool members are HA replicas of each other. To fill this gap, TSM merge heuristics currently de-duplicate any points at a given timestamp where the values are identical. This logic can incorrectly de-duplicate data points that are from disjointed/non-HA shards that are coincidentally identical, thus misrepresenting the actual data in the final merged output.

To solve this, we need to extend backend options to support a `replica_group` field. Then shards with the same `replica_group` can be de-duplicated before the merge is performed.

jranson:
## Description

This adds a `replica_group` designation feature to TimeSeriesMerge-compatible backends to allow Trickster to differentiate between replica and disjoint shards to support HA, Federation and hybrid HA+Federation ALB configurations. A TSM pool can contain any number of backends across various Replica Groups, and the TSM operation will de-duplicate shard data within replica groups before merging/reducing across unique replica groups per-timestamp. For backwards compatibility, backends that do not have an explicit Replica Group defined will automatically use the backend name for the replica group to indicate a single-member unique/disjoint/non-HA shard. 

The previous de-duplication mechanism assumed that any identical data at a given timestamp on different shards were de-duplicatable, even when they were only coincidentally identical and not specifically HA replicas.

Also fixes a test-harness-only data race in progressive_collapse_forwarder_test.go

Fixes #1087 

## Type of Change

- - [x] Bug fix
- - [x] New feature
- - [ ] Optimization
- - [x] Test coverage
- - [x] Documentation
- - [ ] Infrastructure

## AI Disclosure


pkg/backends/alb/mech/tsm/merge_plan.go
```diff
@@ -333,9 +348,7 @@ func applyPlanCompleteness(
 	plan *tsmerge.TSMMergePlan,
 	executions []planVariantExecution,
 	memberCount int,
-) ([]string, bool) {
-	var warnings []string
-	var hasFailure bool
+) {
 	for member := range memberCount {
```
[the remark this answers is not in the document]
jranson: This is incorrect guidance based on older versions of Golang. Go 1.25 and above support ranging over ints. Feedback rejected.

pkg/backends/alb/client.go
```diff
@@ -233,6 +238,40 @@ func (c *Client) ValidateAndStartPool(clients backends.Backends, hcs healthcheck
...
+	if visited.Contains(name) {
+		return fmt.Errorf("%w: cycle encountered at backend %q",
+			alberr.ErrInvalidTimeSeriesMergeProvider, name)
+	}
+	client, ok := clients[name]
+	if !ok || client == nil || client.Configuration() == nil {
+		return alberr.NewErrInvalidPoolMemberName("", name)
+	}
+	cfg := client.Configuration()
+	if providers.IsSupportedTimeSeriesMergeProvider(cfg.Provider) {
```
houyuwushang, line 279: Accepting a nested ALB here makes it look TSM-compatible, but the outer TSM still gets its plan and finalizer from `hl[0].Backend()`. That is the immediate ALB, which does not implement `TSMMergeProvider`, so it falls back to the generic dedup plan. I reproduced this with two RR-over-Prometheus members: `sum(requests)` returned `1` instead of `2`. We either need to expose the terminal provider's planner through a compatible ALB wrapper or keep this topology rejected.

pkg/backends/options/options.go
```diff
@@ -591,6 +595,14 @@ func (l Lookup) Initialize() error {
 // any values that were set during YAML unmarshaling
 func (o *Options) Initialize(name string) error {
 	o.Name = name
+	o.ReplicaGroup = strings.TrimSpace(o.ReplicaGroup)
+	if !providers.IsSupportedTimeSeriesMergeProvider(o.Provider) &&
```
houyuwushang, line 601: `pool.NewTarget` captures the immediate member's `ReplicaGroup`, so this prevents compatible nested ALB wrappers from sharing a replica group. A nested RR-over-Prometheus backend with `replica_group: shard-a` fails during initialization with this error. If nested ALBs are intentionally supported by `validateTSMPoolMemberProvider`, this validation needs to be topology-aware as well; otherwise HA plus nested ALB cannot be represented.

pkg/backends/alb/mech/tsm/time_series_merge.go
```diff
@@ -686,12 +700,22 @@ func (h *handler) serveStandard(
...
-		contributions[i] = results[i].contrib
+	contributions := usableGatherContributions(results)
+	var configuredTargets pool.Targets
+	if len(configured) > 0 {
+		configuredTargets = configured[0]
+	}
+	logical := contributions
+	var groupWarnings []string
+	var groupFailure bool
+	if mergeStrategy != tsmerge.StrategyScalar {
```
houyuwushang, line 778: Skipping contribution coalescing for scalar values preserves first-non-NaN selection, but skipping the grouped result/status view leaks physical failures into the logical status. With two members in one replica group (primary 500, replica scalar 42), the body is correct but `X-Trickster-Result` is still `status=phit`. A covered replica failure should not make the response partial. We can leave scalar contributions uncoalesced while still using `coalesceReplicaResults` for status and headers.

houyuwushang (changes requested):
I found correctness gaps in nested ALB compatibility and scalar replica fallback.

integration/alb_nested_test.go
```diff
@@ -135,6 +136,106 @@ func TestALBNestedPoolAvailable(t *testing.T) {
...
+  default:
+    port: %d
+  metrics:
+    port: %d
+  mgmt:
+    port: %d
+backends:
+  prom-a1:
+    provider: prometheus
+    origin_url: %s
```
houyuwushang, line 172: Could this integration case also give each terminal Prometheus backend a different injected label and assert that it is removed before merging? I reproduced `sum by (job) (requests)` returning three series (`route=a1`, `route=a2`, and `route=b`, each `1`) instead of one `{job="api"}` series with value `2`. The outer TSM collects strip keys only from these immediate ALB wrappers, so it never sees `prometheus.labels` on their terminal backends.

houyuwushang (changes requested):
The three original cases are fixed. I found one remaining nested-ALB merge issue with injected Prometheus labels.

houyuwushang (changes requested):
The nested injected-label case is fixed on this head. One scope issue remains: the branch now differs from `main` across 146 files. The merge added `tasks/lessons.md` and `tasks/todo.md`, and `e8d489b` adds `hack/check-imports` plus repository-wide import formatting. Those changes are unrelated to replica groups and make the release diff much harder to review. Could they be dropped from this branch or moved to a separate PR so #1088 contains only the replica-group work?

jranson:
@houyuwushang this is due to the merge from #1090. I can remove the tasks directory that was created from that PR, but the main changes were adjusting imports. You can just check out de4bb6c4386c5bfd9836e5a4299eb6004bcf9819 to add more comments regarding the core changes for replica groups, but the PR will land with the import cleanup in e8d489b137ecd42a82b20ccd3ebad2024193e0e9

houyuwushang (changes requested):
The `tasks` files are gone and the replica-group changes look correct. One issue remains in the retained import checker: running its documented command, `go run ./hack/check-imports`, from the repository root exits 1 and reports 11 files under `integration/`, including `integration/alb_nested_test.go`. The walker enters the nested integration module, but those files still group external imports before local Trickster imports. Could the integration imports be formatted to match the checker, or should the walker skip nested modules? As written, the new checker does not pass on this head.

jranson:
The import checker is not currently part of the official project checks via the Makefile or CI workflow, so it is not intended to be ran by AI agents. Right now I'm focusing on `./pkg` since that is what other projects would import, so that is more important for correctness.

houyuwushang (approved):

LimitlessEarth (approved):

PR #1102, opened 2026-08-05
feat: cache InfluxDB Prometheus remote reads

### LINKED ISSUE #272: Trickster with influxdb prometheus endpoint needs post_uri
There needs to be a post_uri addon unless i mis-understand how this functions. 

The schema for using the influxdb URL  build into influxdb is:

`http://myinflux.server.com:8086/api/v1/prom/read?db=somedatabase&u=someuser&p=somepassword`

The only way i could do this was using nginx.

houyuwushang:
## Description

Fixes #272.

Adds DeltaProxyCache support for InfluxDB 1.x `POST /api/v1/prom/read` requests using the single-query `SAMPLES` shape supported by InfluxDB. Unsupported request shapes remain on the existing proxy path, preserving the origin's status and response body.

The implementation canonicalizes label matchers for cache identity, keeps InfluxDB query and authorization inputs isolated, maps InfluxDB's exclusive wire end to 1 ms cache extents, and uses `hints.step_ms` only for retention and backfill policies rather than as guaranteed sample spacing. Snappy decoding is bounded, and the supported scope is documented.

The original revision was validated on Linux against InfluxDB 1.12.4. Extending a cached request from three to four samples fetched only the missing final millisecond; the next request was a full hit, and the four-sample response remained available after stopping InfluxDB. A two-query request retained InfluxDB's 400 response and body.

### Review follow-up (2026-09-14)

The follow-up uses the existing klauspost Snappy package, requests Snappy without an extra HTTP compression layer, and reapplies wire headers after DPC's buffer marshal. No-positive-hint requests take the proxy path. The logical policy step survives serialized cache entries and drives LRU counts without changing raw-sample extent precision.

Local validation for this revision:

```text
go test -p 4 -count=1 ./pkg/backends/influxdb/... ./pkg/timeseries/... ./pkg/proxy/engines ./pkg/backends/prometheus/... ./pkg/backends/alb/... ./pkg/backends/clickhouse/... ./pkg/backends/graphite/...
go test -race -count=1 ./pkg/backends/influxdb/... ./pkg/timeseries/... ./pkg/proxy/engines
go vet ./pkg/backends/influxdb/... ./pkg/timeseries/... ./pkg/proxy/engines
go build -p 4 ./...
```

The focused regression cases were also repeated 10 times. They cover gzip-wrapped or missing-encoding origin responses, no-cache reads, hour-long retention with memory and filesystem caches, changed hints, and legacy MsgPack entries. Lint and import checks were run, and both Go modules were tidied.

A fresh live InfluxDB run for this revision is still pending. Full Windows test compilation also hits existing `syscall.Kill` references in daemon and integration tests; the same errors reproduce at the pre-fix head `782027f7`.

## Type of Change

- - [ ] Bug fix
- - [x] New feature
- - [ ] Optimization
- - [x] Test coverage
- - [x] Documentation
- - [ ] Infrastructure

## AI Disclosure


pkg/backends/influxdb/promremote/snappy.go
```diff
@@ -0,0 +1,83 @@
...
+	"errors"
+	"fmt"
+	"io"
+	"math"
+	"net/http"
+
+	bo "github.com/trickstercache/trickster/v2/pkg/backends/options"
+	"github.com/trickstercache/trickster/v2/pkg/proxy/request"
+
+	"github.com/golang/snappy"
```
jranson: should use "github.com/klauspost/compress/snappy" here. We already import the klauspost compress module and its snappy lib is more performant than the golang snappy.

pkg/backends/influxdb/handler_promremote_test.go
```diff
@@ -0,0 +1,267 @@
...
+	"time"
+
+	"github.com/trickstercache/trickster/v2/pkg/backends/influxdb/promremote"
+	"github.com/trickstercache/trickster/v2/pkg/backends/influxdb/promremote/prompb"
+	"github.com/trickstercache/trickster/v2/pkg/backends/providers"
+	"github.com/trickstercache/trickster/v2/pkg/proxy/headers"
+	"github.com/trickstercache/trickster/v2/pkg/proxy/request"
+	tu "github.com/trickstercache/trickster/v2/pkg/testutil"
+
+	"github.com/golang/snappy"
```
jranson: also use the alternate snappy import here

pkg/backends/influxdb/promremote/promremote.go
```diff
@@ -0,0 +1,220 @@
...
+	"strings"
+	"time"
+
+	"github.com/trickstercache/trickster/v2/pkg/backends/influxdb/iofmt"
+	"github.com/trickstercache/trickster/v2/pkg/backends/influxdb/promremote/prompb"
+	"github.com/trickstercache/trickster/v2/pkg/proxy/headers"
+	"github.com/trickstercache/trickster/v2/pkg/proxy/request"
+	"github.com/trickstercache/trickster/v2/pkg/timeseries"
+
+	"github.com/golang/snappy"
```
jranson: should use "github.com/klauspost/compress/snappy" here. We already import the klauspost compress module and its snappy lib is more performant than the golang snappy.

pkg/backends/influxdb/promremote/promremote_test.go
```diff
@@ -0,0 +1,288 @@
...
+	"time"
+
+	"github.com/trickstercache/trickster/v2/pkg/backends/influxdb/iofmt"
+	"github.com/trickstercache/trickster/v2/pkg/backends/influxdb/promremote/prompb"
+	bo "github.com/trickstercache/trickster/v2/pkg/backends/options"
+	"github.com/trickstercache/trickster/v2/pkg/proxy/headers"
+	"github.com/trickstercache/trickster/v2/pkg/proxy/request"
+	"github.com/trickstercache/trickster/v2/pkg/timeseries"
+
+	"github.com/golang/snappy"
```
jranson: swap import here too

pkg/backends/influxdb/promremote/serialization.go
```diff
@@ -0,0 +1,244 @@
...
+	"slices"
+	"time"
+
+	"github.com/trickstercache/trickster/v2/pkg/backends/influxdb/promremote/prompb"
+	"github.com/trickstercache/trickster/v2/pkg/proxy/headers"
+	"github.com/trickstercache/trickster/v2/pkg/timeseries"
+	"github.com/trickstercache/trickster/v2/pkg/timeseries/dataset"
+	"github.com/trickstercache/trickster/v2/pkg/timeseries/epoch"
+
+	"github.com/golang/snappy"
```
jranson: swap import for klauspost

pkg/backends/influxdb/promremote/serialization_test.go
```diff
@@ -0,0 +1,226 @@
...
+	"testing"
+	"time"
+
+	"github.com/trickstercache/trickster/v2/pkg/backends/influxdb/promremote/prompb"
+	"github.com/trickstercache/trickster/v2/pkg/proxy/headers"
+	"github.com/trickstercache/trickster/v2/pkg/timeseries"
+	"github.com/trickstercache/trickster/v2/pkg/timeseries/dataset"
+	"github.com/trickstercache/trickster/v2/pkg/timeseries/epoch"
+
+	"github.com/golang/snappy"
```
jranson: use klauspost/compress/snappy

jranson (commented):
@houyuwushang, thanks for getting this one done.

I went ahead and resolved the merge conflicts + got it on par with main + resolved some CodeQL security advisories that were not part of your change but still first detected in this PR. All checks are passing now.

I left some feedback inline about switching to a more performant snappy encoder/decoder that we already import anyway. Otherwise, a few things here:

## Performance - Gzip-of-Snappy in DPC

InfluxDB registers /api/v1/prom/read with Gzipped: true (influxdb@v1.13.0/services/httpd/handler.go:289), and its gzipFilter wraps the writer whenever Accept-Encoding contains gzip. Trickster's DPC still builds its upstream profile from all supported web encodings (deltaproxycache.go:636), and the HTTP proxy still stamps that value onto the upstream request (httpproxy.go:281). So with a real backend, every miss/partial-hit fetch is gzip(snappy(protobuf)): InfluxDB spends CPU deflating an already-snappy payload, Trickster spends CPU inflating it, and both buffers live in memory. Prometheus talking to InfluxDB directly sends Accept-Encoding: snappy and pays none of that. For multi-MB responses this is tens of ms per fetch.

getDecoderReader (deltaproxycache.go:739) does unwrap the gzip, so it works — but nothing in the PR or tests accounts for it. Fix: give the backend a way to opt out of upstream encoding negotiation for pre-compressed payloads (e.g., a RequestOptions field the DPC honors when building the fetch profile in fetchExtents/fetchTimeseries, setting Accept-Encoding: snappy or identity).

## Conformance - Responses can drop Content-Encoding header

serialization.go:191 only sets Content-Type/Content-Encoding when writer is an http.ResponseWriter. The DPC still marshals into a bytes.Buffer and writes wireBody later (deltaproxycache.go:516), so the headers actually sent are doc.SafeHeaderClone() — i.e., whatever the origin returned. Combined with the Gzip-of-Snappy issue, getDecoderReader deletes Content-Encoding after gunzipping, so real deployments will respond with no Content-Encoding: snappy. Prometheus tolerates this (it snappy-decodes any application/x-protobuf body unconditionally), but TestPromRemoteReadDeltaCache's header assertion passes only because the fake origin supplied that header. Fixing Gzip-of-Snappy makes the upstream header survive; otherwise the headers need to be applied somewhere the DPC honors.

## Gaps in 1ms per-Step Mode

* `oldest` mode without `hints.step_ms` (Prometheus sends StepMs=0 for instant queries): retention = 1024 × 1ms or ~1s, because CachePolicyStep() still falls back to Step and the DPC (L263) uses it for the retention window. So every request is a miss plus a pointless cache read/unmarshal/merge/marshal/write — so even slower than a plain proxy. The PR already proxies no-hint requests when BackfillTolerancePoints > 0 (promremote.go:140-145); the same decision (or a const default policy step) should apply to `oldest` case.

* `lru` mode: DataSet.Step()/TimestampCount/DefaultSizeCropper (dataset.go L535) still use the 1 ms step, and PolicyStep is msg:"-" with no PolicyStepNS twin, so the PolicyStep doesn't survive the cache round-trip. A 1h cached range counts as 3.6M "timestamps" against a default TimeseriesRetentionFactor of 1024 → every write evicts nearly everything on nearly every reap. We should persist the PolicyStep across MsgPack so that caches like File and Redis that serialize the DataSet work correctly on cache hits and partial hits against an LRU cache.

jranson (changes requested):
(see previous comment)

houyuwushang:
@jranson I missed the gzip layer in the original fixture. The follow-up switches to klauspost/snappy and carries the encoding override through partial-hit fetches too. DPC now applies the Snappy response headers after the buffer marshal.

For retention, I chose passthrough without a positive hint rather than inventing a cadence for raw samples. The policy step is now persisted and used for LRU counts, while extent calculations stay at 1 ms.

A fresh live InfluxDB run for this revision is still pending; the new regression tests use a controlled HTTP transport and memory/filesystem caches.

jranson (approved):

PR #1104, opened 2026-08-12
feat: expose rule regex captures to rewriters

### LINKED ISSUE #598: Request Rewrite Example
For something like:
```
rules:
  example-user-router:
    # default route is reader cluster
    next_route: example-reader-cluster

    input_source: path
    input_type: string
    operation: rmatch      # perform regex match against the path to see if it matches 'writer
    operation_arg: '^.*%7Bmylabel%3D%22([a-z0-9]{3})%22.*$'
    cases:
      rmatch-true:
        matches: [ 'true' ] # rmatch returns true when the input matches the regex; update next_route
        next_route: example-writer-cluster
```

Trying to take the capture group above and add into the hostname as a prefix.  Basically like a relabel_config in prometheus.  I didn't see an examples on how to do it.  Most of the examples for rewrites involve fixed values like here: https://github.com/trickstercache/trickster/blob/main/docs/request_rewriters.md#header-set

Looking through the source: https://github.com/trickstercache/trickster/blob/0cf4524824c6916aade49dbd99ef360e61426289/pkg/proxy/request/rewriter/rewrite_instructions_test.go#L71

Looks like that uses a ${trickster} var.  Don't see where it originated.

houyuwushang:
## Description

Adds regex capture expansion to Rule Router request rewriters. An `rmatch` rule now exposes `${0}`, numbered submatches such as `${1}`, and named submatches such as `${tenant}` to the matched case rewriter and that rule's egress rewriter. Captures are request-scoped and cleared before the next route.

The change also makes explicit scheme, host, hostname, and port rewrites apply to the final upstream URL. Because a captured hostname can select a different origin, the final rewritten authority participates in cache identity to prevent responses from different origins sharing an entry.

Existing rules without token-bearing rewriters do not calculate capture maps. Undefined tokens remain unchanged, optional unmatched captures expand to an empty string, and expansion is non-recursive. The documentation includes the decoded-path behavior from the original example and warns that client-derived authorities require restrictive regular expressions because configured credentials follow the selected origin.

Fixes #598.

### Validation

- `go test -count=20 ./pkg/backends/rule ./pkg/proxy/request/rewriter`
- `go test -race -count=5 ./pkg/backends/rule ./pkg/proxy/request/rewriter`
- `go test -count=1 ./pkg/backends/... ./pkg/proxy/...`
- `go vet ./pkg/backends/... ./pkg/proxy/...`
- `go tool golangci-lint run --timeout 5m -c .golangci.yml ./pkg/backends/rule/... ./pkg/proxy/request/rewriter/... ./pkg/proxy/urls/... ./pkg/proxy/engines/...`
- Linux/amd64 compile-only checks for the changed Rule, rewriter, URL, and engine packages

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Optimization
- [x] Test coverage
- [x] Documentation
- [ ] Infrastructure

## AI Disclosure

### DIFF
--- a/pkg/backends/rule/rule.go
+++ b/pkg/backends/rule/rule.go
@@ -18,6 +18,8 @@ package rule
 
 import (
 	"net/http"
+	"regexp"
+	"strconv"
 
 	"github.com/trickstercache/trickster/v2/pkg/proxy/context"
 	"github.com/trickstercache/trickster/v2/pkg/proxy/handlers/trickster/failures"
@@ -57,8 +59,10 @@ type rule struct {
 
 	cases caseList
 
-	extractionArg string
-	operationArg  string
+	extractionArg    string
+	operationArg     string
+	regex            *regexp.Regexp
+	hasCaptureTokens bool
 
 	defaultRedirectURL  string
 	defaultRedirectCode int
@@ -85,6 +89,7 @@ type evaluatorFunc func(*http.Request) (http.Handler, *http.Request, error)
 var badRequestHandler = http.HandlerFunc(failures.HandleBadRequestResponse)
 
 func (r *rule) EvaluateOpArg(hr *http.Request) (http.Handler, *http.Request, error) {
+	hr = rewriter.WithoutTokens(hr)
 	currentHops, maxHops := context.Hops(hr.Context())
 	if r.maxRuleExecutions < maxHops {
 		maxHops = r.maxRuleExecutions
@@ -100,13 +105,21 @@ func (r *rule) EvaluateOpArg(hr *http.Request) (http.Handler, *http.Request, err
 	}
 
 	h := r.defaultRouter
-	res := r.operationFunc(r.extractionFunc(hr, r.extractionArg),
-		r.operationArg, r.negateOpResult)
+	input := r.extractionFunc(hr, r.extractionArg)
+	res := r.operationFunc(input, r.operationArg, r.negateOpResult)
+	var captureTokens map[string]string
+	if r.regex != nil && r.hasCaptureTokens {
+		matches := r.regex.FindStringSubmatch(input)
+		captureTokens = regexCaptureTokens(r.regex, matches)
+	}
 	var nonDefault bool
 
 	for _, c := range r.cases {
 		if c.matchValue == res {
 			nonDefault = true
+			if len(captureTokens) > 0 {
+				hr = rewriter.WithTokens(hr, captureTokens)
+			}
 			h, hr = handleMatchedCase(c, hr)
 		}
 	}
@@ -125,9 +138,28 @@ func (r *rule) EvaluateOpArg(hr *http.Request) (http.Handler, *http.Request, err
 			r.defaultRedirectCode, r.defaultRedirectURL))
 	}
 
+	hr = rewriter.WithoutTokens(hr)
 	hr = hr.WithContext(context.WithHops(hr.Context(), currentHops+1, maxHops))
 
 	return h, hr, nil
+}
+
+func regexCaptureTokens(re *regexp.Regexp, matches []string) map[string]string {
+	if len(matches) == 0 {
+		return nil
+	}
+	tokens := make(map[string]string, len(matches)*2)
+	names := re.SubexpNames()
+	for i, match := range matches {
+		tokens[strconv.Itoa(i)] = match
+		if i >= len(names) || names[i] == "" {
+			continue
+		}
+		if _, ok := tokens[names[i]]; !ok {
+			tokens[names[i]] = match
+		}
+	}
+	return tokens
 }
 
 func (r *rule) EvaluateCaseArg(hr *http.Request) (http.Handler, *http.Request, error) {
--- a/pkg/backends/rule/rule.go
+++ b/pkg/backends/rule/rule.go
@@ -106,11 +106,17 @@ func (r *rule) EvaluateOpArg(hr *http.Request) (http.Handler, *http.Request, err
 
 	h := r.defaultRouter
 	input := r.extractionFunc(hr, r.extractionArg)
-	res := r.operationFunc(input, r.operationArg, r.negateOpResult)
 	var captureTokens map[string]string
+	var res string
 	if r.regex != nil && r.hasCaptureTokens {
 		matches := r.regex.FindStringSubmatch(input)
+		res = falseValue
+		if len(matches) > 0 {
+			res = trueValue
+		}
 		captureTokens = regexCaptureTokens(r.regex, matches)
+	} else {
+		res = r.operationFunc(input, r.operationArg, r.negateOpResult)
 	}
 	var nonDefault bool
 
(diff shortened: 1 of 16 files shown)


pkg/backends/rule/rule.go
```diff
@@ -100,13 +105,21 @@ func (r *rule) EvaluateOpArg(hr *http.Request) (http.Handler, *http.Request, err
 	}
 
 	h := r.defaultRouter
-	res := r.operationFunc(r.extractionFunc(hr, r.extractionArg),
-		r.operationArg, r.negateOpResult)
+	input := r.extractionFunc(hr, r.extractionArg)
+	res := r.operationFunc(input, r.operationArg, r.negateOpResult)
+	var captureTokens map[string]string
+	if r.regex != nil && r.hasCaptureTokens {
+		matches := r.regex.FindStringSubmatch(input)
```
jranson, line 112: this implementation, by using `FindStringSubmatch()`, executes the identical regex evaluation performed by `r.operationFunc())` 4 lines above on L108. We should evaluate the regex only once and reuse the original evaluation as needed. 

This should work:
```
	h := r.defaultRouter
	input := r.extractionFunc(hr, r.extractionArg)
	var captureTokens map[string]string
	var res string
	if r.regex != nil && r.hasCaptureTokens {
		matches := r.regex.FindStringSubmatch(input)
		res = falseValue
		if len(matches) > 0 {
			res = trueValue
		}
		captureTokens = regexCaptureTokens(r.regex, matches)
	} else {
		res = r.operationFunc(input, r.operationArg, r.negateOpResult)
	}
	var nonDefault bool
```

pkg/proxy/urls/upstream_rewrites.go
```diff
@@ -0,0 +1,132 @@
...
+	"net/http"
+	"net/url"
+	"slices"
+	"strings"
+)
+
+type upstreamURLComponent uint8
+
+const (
+	upstreamScheme upstreamURLComponent = iota
```
jranson: recommend using a start value of 1 here and bitshift the rest. then you can use 'bitwise AND' comparisons instead of switch statements when checking the flags down in the apply function.

`upstreamScheme upstreamURLComponent = 1 << iota`

pkg/proxy/urls/upstream_rewrites.go
```diff
@@ -0,0 +1,132 @@
...
+func SetUpstreamPort(r *http.Request, port string) {
+	setUpstreamURLRewrite(r, upstreamPort, port)
+}
+
+func setUpstreamURLRewrite(r *http.Request, component upstreamURLComponent, value string) {
+	if r == nil {
+		return
+	}
+	rewrites, _ := r.Context().Value(upstreamURLRewritesKey{}).([]upstreamURLRewrite)
+	rewrites = append(slices.Clone(rewrites), upstreamURLRewrite{
```
jranson: While there are likely to be very few calls to `setUpstreamURLRewrite` for the same request; in situations where there could be many (e.g., a complex config with many rules), the entire history of rewrites is cloned to a new slice, then the next rewrite rule is appended to the clone, which is finally saved to the context - all happening once per call to `setUpstreamURLRewrite`.  Then `applyUpstreamURLRewrites` iterates through all of those rewrite directions again, replaying the full sequence to reach the final rewritten URL state. Let's make this more performant.

Since rewrites only mutate URL Scheme, Host, Hostname and Port, you should be able to track the latest change for each of those 4 url components in `setUpstreamURLRewrite` via `upstreamURLRewrite` rather than compiling and replaying a change history of unbounded length.  The `components` (now-plural) field tracks which of the 4 components are included in the final `apply` operation (using bitwise logic to hold all 4 flags in this single value) while `scheme`, `host`, `hostname`, and `port` track the final value for each component that is to be changed.

```
type upstreamURLRewrite struct {
	components                   upstreamURLComponent
	scheme, host, hostname, port string
}
```

Then in `setUpstreamURLRewrite`, the scheme, host, hostname or port (as indicated by `component`) is set to the incoming value:

```
func setUpstreamURLRewrite(r *http.Request, component upstreamURLComponent, value string) {
	if r == nil {
		return
	}
	rewrites, _ := r.Context().Value(upstreamURLRewritesKey{}).(upstreamURLRewrite)
	switch component {
	case upstreamScheme:
		rewrites.scheme = value
	case upstreamHost:
		rewrites.host = value
		rewrites.hostname = ""
		rewrites.port = ""
		rewrites.components &^= upstreamHostname | upstreamPort
	case upstreamHostname:
		rewrites.hostname = value
	case upstreamPort:
		rewrites.port = value
	}
	rewrites.components |= component
	*r = *r.WithContext(context.WithValue(r.Context(), upstreamURLRewritesKey{}, rewrites))
}
```

Then for applying, you can use a dedicated `apply` function on rewrites, and call that as needed (and `apply` uses bitwise AND to determine if a given url component value changes  🚀):
```
func (rewrites upstreamURLRewrites) apply(u *url.URL) {
	if rewrites.components&upstreamScheme != 0 {
		u.Scheme = rewrites.scheme
	}
	if rewrites.components&upstreamHost != 0 {
		u.Host = rewrites.host
	}
	if rewrites.components&upstreamHostname != 0 {
		u.Host = replaceHostname(u.Host, rewrites.hostname)
	}
	if rewrites.components&upstreamPort != 0 {
		u.Host = replacePort(u.Host, rewrites.port)
	}
}

func applyUpstreamURLRewrites(r *http.Request, u *url.URL) {
	if r == nil || u == nil {
		return
	}
	rewrites, _ := r.Context().Value(upstreamURLRewritesKey{}).(upstreamURLRewrites)
	rewrites.apply(u)
}

func UpstreamURLRewriteCacheKey(r *http.Request, base *url.URL) string {
	if r == nil || base == nil {
		return ""
	}
	rewrites, _ := r.Context().Value(upstreamURLRewritesKey{}).(upstreamURLRewrites)
	if rewrites.components == 0 {
		return ""
	}
	rewritten := Clone(base)
	rewrites.apply(rewritten)
	if rewritten.Scheme == base.Scheme && rewritten.Host == base.Host {
		return ""
	}
	return "\x00upstream=" + rewritten.Scheme + "://" + rewritten.Host + "\x00"
}
```

jranson (changes requested):
@houyuwushang just 2 small performance items to fix up

pkg/backends/rule/rule.go
houyuwushang: The capture-token path now derives the rule result from the same `FindStringSubmatch` result; `operationFunc` is only called on the no-capture fast path.

pkg/proxy/urls/upstream_rewrites.go
houyuwushang: Changed these to `1 << iota`; the apply path now checks the combined component mask with bitwise AND.

pkg/proxy/urls/upstream_rewrites.go
houyuwushang: Replaced the slice history with one `upstreamURLRewrite` value in the request context. A host rewrite clears earlier hostname/port flags, while later component rewrites update that final state before apply and cache-key derivation.

jranson (approved):

PR #623, opened 2022-12-04
MySQL parsing implemented, plus minor fixes in some related files

jnichols-git:
Signed-off-by: jakenichols2719 <jnichols2719@protonmail.com>

PR #1109, opened 2026-08-19
feat: add MySQL Support

jranson:
## Description

This adds:
* MySQL wire protocol listener
* MySQL Query parsing via [Vitess](https://github.com/vitessio/vitess/tree/main/go/vt/sqlparser)
* Re-use of the generic sqlanalyzer to assess a parsed MySQL query's cacheability level
* Timeseries caching of supported MySQL queries (e.g., from Grafana dashboards and similar queries outlined in the new docs)
* Fallback Object caching of non-Timeseries MySQL SELECT queries
* NativeListener adapter that can extend to work for a future ClickHouse native listener (#981), etc.
* Updated docs, developer env, data seeding, example configs, etc. for MySQL

This replaces #623

## Type of Change

- - [ ] Bug fix
- - [x] New feature
- - [ ] Optimization
- - [x] Test coverage
- - [x] Documentation
- - [ ] Infrastructure

## AI Disclosure


pkg/backends/mysql/protocol.go
```diff
@@ -0,0 +1,1390 @@
...
+	cacheStatus := cachestatus.LookupStatusProxyOnly
+	if err != nil {
+		cacheStatus = cachestatus.LookupStatusProxyError
+	}
+	h.observeCache(analysis.Mode, cacheStatus, points, time.Since(proxyStarted))
+	return err
+}
+
+func hasMultipleStatements(query string) (bool, error) {
+	pieces, err := defaultAnalyzer.(*Analyzer).parser.SplitStatementToPieces(query)
```
houyuwushang: [P1] Guard `SplitStatementToPieces` against parser panics

This call can panic on malformed short input instead of returning an error. The PR's fuzz target reproduces `string("\"\\;0")` with `slice bounds out of range [:5] with length 4` inside Vitess `parser.go:322`. Vitess recovers at the listener goroutine and closes the client, but a client can repeatedly force connection drops and panic logging. Catch/avoid the panic and return the documented parse error, and keep a regression seed for this case.

pkg/backends/mysql/protocol.go
```diff
@@ -0,0 +1,1390 @@
...
+		metrics.MySQLConnectionErrors.WithLabelValues(h.config.BackendName, "upstream_connect").Inc()
+		return sqlerror.NewSQLError(sqlerror.CRServerGone, sqlerror.SSNetError,
+			"Trickster could not connect to the MySQL origin")
+	}
+	if h.closed.Load() {
+		conn.Close()
+		return sqlerror.NewSQLError(sqlerror.CRServerGone, sqlerror.SSUnknownSQLState,
+			"Trickster MySQL listener is shutting down")
+	}
+	session.conn = conn
```
houyuwushang, line 1014: [P1] Replay tracked session state after reconnect

`updateSessionState` records `USE` and the cache-safe `time_zone`, while `discardUpstream` clears `session.conn`. A later `connectSession` creates a fresh origin connection from only `h.config.Upstream` and never replays that state. The next origin query can therefore run in the configured default database/time zone while `queryCacheKey` still identifies it with the client's selected database/time zone (`pkg/backends/mysql/cache.go:505-521`). Restore the tracked state before accepting a query, or invalidate/disable the affected cache identity when the origin is rebuilt.

pkg/backends/alb/mech/ur/user_router.go
```diff
@@ -110,54 +102,81 @@ func (h *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	}
+	if input.Username == "" || opts == nil || len(opts.Users) == 0 {
+		return useDefault()
+	}
+	mapping, ok := opts.Users[input.Username]
+	if !ok || mapping == nil {
+		return useDefault()
+	}
+	target, ok := routes[input.Username]
+	if !ok || !target.Available() {
```
houyuwushang, line 151: [P1] Do not fall back for a mapped unavailable terminal

When a username has an explicit route but its terminal is unhealthy/unavailable, this branch calls `useDefault()` and can send that user to another tenant's default backend. This contradicts the native MySQL contract in `docs/mysql.md:368-370`, which requires a bounded availability error and no silent reroute. Return an unavailable/no-route decision here so the native authenticator rejects the session instead of selecting the default tenant.

pkg/backends/mysql/native_listener.go
```diff
@@ -0,0 +1,285 @@
...
+	result := make([]string, 0, len(seen))
+	for name := range seen {
+		result = append(result, name)
+	}
+	slices.Sort(result)
+	return result
+}
+
+func routedRestartKey(c *config.Config, router *bo.Options) string {
+	parts := []any{router.ALBOptions.UserRouter, router.RequireTLS, router.TLS}
```
houyuwushang: [P1] Include listener-facing credentials in the routed restart key

`nativeProtocolConfig` builds `DownstreamUsers` from the router's authenticator, but `routedRestartKey` only includes the User Router options, TLS, and terminal backend keys. A change to `router.AuthOptions.Users` or the contents of `UsersFile` therefore leaves the listener descriptor unchanged, so reload can keep the old authenticator and old password. Include the resolved credential map and/or a content hash of the users file in this key.

pkg/backends/mysql/cache.go
```diff
@@ -0,0 +1,1051 @@
...
+	return 0, fmt.Errorf("unsupported MySQL time unit %d", unit)
+}
+
+func rowGroupKey(row []sqltypes.Value, indexes []int) string {
+	if len(indexes) == 0 {
+		return ""
+	}
+	var builder strings.Builder
+	for _, index := range indexes {
+		value := row[index].Raw()
```
houyuwushang: [P1] Preserve SQL NULL versus empty-string group identity

`sqltypes.Value.Raw()` returns nil for SQL NULL and an empty slice for an empty string; both are encoded here as `0:`. `mergeResults` then uses that value in its `(epoch, group)` map key and overwrites the later row when two groups differ only by NULL versus empty string. Encode a null marker (and preferably the value type) so DPC merging cannot drop one of the groups.

houyuwushang (commented):
I reviewed this against commit bc306c6c90f186ab62f40c365fd24943be5d88c9. The comments below are blocking correctness issues: native routing can cross tenant boundaries when a mapped terminal is unavailable, reconnects can execute in a different database than the cache identity records, and the protocol/cache paths lose MySQL semantics in several cases. I also included reproducible parser and cache-classification issues.

pkg/backends/mysql/protocol.go
```diff
@@ -0,0 +1,1390 @@
...
+	return err
+}
+
+func hasMultipleStatements(query string) (bool, error) {
+	pieces, err := defaultAnalyzer.(*Analyzer).parser.SplitStatementToPieces(query)
+	return len(pieces) > 1, err
+}
+
+func unsupportedTextFeature(query string) string {
+	trimmed := vtparser.StripLeadingComments(query)
```
houyuwushang: [P1] Reject executable comments anywhere in the statement

This only detects `/*!...*/` after stripping leading comments. An embedded executable comment is still passed to the analyzer; for example, a normal DPC-shaped query with `SELECT /*!80000 SQL_NO_CACHE */ ...` is classified as `CacheModeDelta` in the current code. MySQL executes the directive, so the proxy can cache a query whose semantics explicitly request a different execution policy. Scan the complete statement and reject versioned executable comments before analysis, not just the leading form.

pkg/backends/mysql/parsing.go
```diff
@@ -0,0 +1,1002 @@
...
+	}, stmt)
+	return found
+}
+
+func isNondeterministic(stmt sqlparser.SQLNode) bool {
+	unsafe := false
+	names := map[string]struct{}{
+		"connection_id": {}, "current_timestamp": {}, "curdate": {}, "curtime": {},
+		"database": {}, "found_rows": {}, "get_lock": {}, "is_free_lock": {},
+		"is_used_lock": {}, "last_insert_id": {}, "now": {}, "rand": {},
```
houyuwushang: [P1] Treat `SLEEP` and `RANDOM_BYTES` as non-cacheable

The volatile-function list omits both built-ins. DPC-shaped queries containing `SLEEP(1)` or `RANDOM_BYTES(16)` currently fall through to `CacheModeObject`, and `executeObject` stores the result for later clients. Add both names to the nondeterministic set (and add classification tests) so these values are always proxied.

pkg/backends/mysql/protocol.go
```diff
@@ -0,0 +1,1390 @@
...
+		if err == nil {
+			h.updateSessionState(session, query)
+		}
+		return err
+	}
+	var result *sqltypes.Result
+	var warnings uint16
+	err := h.runOriginQuery(session, upstream, func() error {
+		var executeErr error
+		result, warnings, executeErr = upstream.ExecuteFetchWithWarningCount(query, 1, true)
```
houyuwushang, line 1394: [P1] Preserve upstream status flags and warnings in downstream responses

The non-streaming path receives `result.StatusFlags` and a warning count, but only the warning count is saved; Vitess builds the downstream OK packet from `c.StatusFlags`, not from the result. The streaming path has the same problem because `ExecuteStreamFetch`/`FetchNext` discard the final EOF/OK status and warning fields. This loses semantics such as `SERVER_STATUS_IN_TRANS` after `BEGIN` and upstream SELECT warnings. Copy the upstream flags/warnings into the downstream connection/session before invoking the callback, and cover both DML and result-set paths.

pkg/backends/mysql/protocol.go
```diff
@@ -0,0 +1,1390 @@
...
+		ctx, span = h.config.Tracer.Start(session.traceContext, "mysql.origin.connect",
+			trace.WithAttributes(attribute.String("trickster.backend", h.config.BackendName)))
+		defer span.End()
+	}
+	if h.config.ConnectTimeout > 0 {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(ctx, h.config.ConnectTimeout)
+		defer cancel()
+	}
+	conn, err := vtmysql.Connect(ctx, &h.config.Upstream)
```
houyuwushang: [P2] Propagate `CLIENT_FOUND_ROWS` to the origin connection

The downstream capability is available as `c.Capabilities`, but every origin connection is created from the static `h.config.Upstream` params. Without copying `CapabilityClientFoundRows` into those params, UPDATE/DELETE affected-row semantics change for clients that negotiated this flag. Set the upstream flag per downstream session before connecting (and test both capability settings).

pkg/backends/mysql/protocol.go
```diff
@@ -0,0 +1,1536 @@
...
+		target.setControl(c.ConnectionID, control)
+	}
+	c.ClientData = &routedConnection{target: target}
+	target.NewConnection(c)
+	target.ConnectionReady(c)
+}
+
+func (h *routedProtocolHandler) target(c *vtmysql.Conn) (*protocolHandler, error) {
+	if c != nil {
+		if routed, ok := c.ClientData.(*routedConnection); ok && routed.target != nil {
```
houyuwushang, line 716: [P1] Handle the initial database before `ConnectionReady` in routed mode.

Vitess invokes `ComQuery("use ...")` for `CLIENT_CONNECT_WITH_DB` after authentication but before `ConnectionReady`. At that point `ClientData` is still a `backends.RouteDecision`, while `target()` only accepts `*routedConnection`, so a routed client whose normal DSN includes a database fails the handshake with `CRServerGone`. Following the real callback order (`NewConnection`, auth decision, `ComQuery("use `analytics`")`) reproduces `Trickster has no MySQL route for this connection`. The selected target and its session need to be activated before this initial `USE`, without initializing them again in `ConnectionReady`.

pkg/backends/mysql/protocol.go
```diff
@@ -0,0 +1,1536 @@
...
+			"Trickster MySQL listener has no session")
+	}
+	session.mtx.Lock()
+	defer session.mtx.Unlock()
+	if !session.upstreamParamsReady {
+		session.upstream = h.config.Upstream
+		session.upstreamParamsReady = true
+	}
+	h.applyDownstreamCapabilitiesLocked(session, session.downstream)
+	if session.conn != nil {
```
houyuwushang, line 969: [P1] Do not silently reconnect a session whose state cannot be replayed.

`runOriginQuery` currently discards the upstream on every query error, including an ordinary server SQL error. The next command reaches this branch with `conn == nil` and reconnects even when `session.inTx` or `session.cacheUnsafe` is set, but only the database and time zone are replayed. For example, `BEGIN; <statement returning a SQL error>; UPDATE ...` runs the `UPDATE` on a fresh autocommit connection. Temporary tables, locks, and user variables are lost in the same way. Please either retain a still-usable connection for server errors or close/refuse the downstream session when unreplayable state would be lost.

pkg/backends/mysql/parsing.go
```diff
@@ -0,0 +1,1023 @@
...
+		return sqlanalyzer.Analysis{
+			Mode: mode, Reason: sqlanalyzer.ReasonInvalidSQL,
+			Err: fmt.Errorf("%w: %w", ErrInvalidSQL, parseErr),
+		}
+	}
+	selectStmt, ok := stmt.(*sqlparser.Select)
+	if !ok {
+		mode := sqlanalyzer.CacheModeNone
+		if _, readOnly := stmt.(sqlparser.SelectStatement); readOnly {
+			mode = sqlanalyzer.CacheModeObject
```
houyuwushang: [P1] Fail closed for invalid and non-`*Select` read-only ASTs instead of granting OPC by default.

Both this branch and the `parseErr` branch above skip the determinism/session-state scan. `SELECT 1 UNION ALL SELECT RAND()` and `SELECT 1 UNION ALL SELECT @session_value` currently return `mode=object, reason=unsupported_statement`; a parser-rejected statement beginning with `SELECT` also returns `object`. That can reuse a random or session-local result, and an origin-valid parser-skew query can change state without making the session cache-unsafe. These forms need `CacheModeNone` unless the complete AST has passed the same safety proof as a plain `SELECT`.

pkg/backends/mysql/protocol.go
```diff
@@ -0,0 +1,1536 @@
...
+	return nil
+}
+
+func (h *protocolHandler) applyDownstreamCapabilitiesLocked(session *upstreamSession,
+	downstream *vtmysql.Conn,
+) {
+	if session == nil || downstream == nil {
+		return
+	}
+	session.upstream.Flags &^= uint64(vtmysql.CapabilityClientFoundRows)
```
houyuwushang, line 1030: [P1] Preserve the downstream handshake collation in both the upstream connection and cache identity.

`Conn.CharacterSet` is negotiated MySQL session state, but this copies only `CLIENT_FOUND_ROWS`; the upstream keeps its configured charset, and `queryCacheKey` also omits the downstream collation. The same query bytes can therefore be interpreted differently under latin1 and utf8mb4 while sharing one cached result. A minimal check leaves an utf8mb4 upstream at collation 255 after applying a latin1 downstream at collation 8, and both connections produce the same cache key. Please propagate the effective collation before connecting and include it in the cache identity/reset behavior.

pkg/backends/mysql/cache.go
```diff
@@ -0,0 +1,1281 @@
...
+		}
+	}
+	slices.SortStableFunc(rows, func(a, b keyedRow) int {
+		if a.epoch < b.epoch {
+			return -1
+		}
+		if a.epoch > b.epoch {
+			return 1
+		}
+		return strings.Compare(a.group, b.group)
```
houyuwushang: [P2] Do not use the collision-safe serialized identity as the SQL sort key.

`rowGroupKey` now places the byte length before the raw value, and this comparison consequently changes `ORDER BY time, metric`. With a binary `VARBINARY` group at one timestamp, rows `"z"` and `"aa"` are emitted as `z, aa`, although MySQL binary ascending order is `aa, z`. The identity is suitable for deduplication, but ordering needs a separate comparison that preserves the field's MySQL ordering semantics (with a regression covering different-length group values).

houyuwushang (commented):
I rechecked the current head against my earlier comments; those items are addressed. I found five additional correctness gaps in the new routing, session, analyzer, and DPC paths below. Each has a small reproducer and should be handled before merge.

pkg/backends/mysql/protocol.go
```diff
@@ -0,0 +1,1728 @@
...
+func (h *protocolHandler) proxyQuery(session *upstreamSession, query string, parsed parsedQuery,
+	callback func(*sqltypes.Result) error,
+) error {
+	if err := h.connectSession(session); err != nil {
+		return err
+	}
+	session.mtx.Lock()
+	upstream := session.conn
+	session.mtx.Unlock()
+	if returnsResultSet(parsed.statementType) {
```
houyuwushang: [P1] Dispatch from the actual response shape

The statement type does not safely predict the origin packet shape. Vitess returns StmtUnknown for a WITH ... SELECT even though Parse returns *sqlparser.Select, so a two-row CTE reaches ExecuteFetchWithWarningCount(query, 1, ...) and fails with Row count exceeded 1. The opposite edge also loses data: SELECT ... INTO is classified as SELECT but returns an OK packet, and this streaming path replaces its OK metadata with an empty result. Please classify from the parsed AST where possible and preserve the actual OK/result-set response shape.

pkg/backends/mysql/cache.go
```diff
@@ -0,0 +1,1609 @@
...
+	}
+	return 0, fmt.Errorf("unsupported MySQL time unit %d", unit)
+}
+
+// rowGroupKey builds a collision-safe identity for a row's group columns. It is
+// used only to deduplicate rows across delta parts. It deliberately cannot
+// order them: the length prefix that makes it collision-safe would sort binary
+// "z" before "aa", and raw bytes ignore a text field's collation. Ordering is
+// groupComparator's job.
+func rowGroupKey(row []sqltypes.Value, indexes []int) string {
```
houyuwushang: [P1] Deduplicate groups using MySQL equality

Raw bytes are not a MySQL group identity for collated text. Under utf8mb4_0900_ai_ci, the comparator considers "a" and "A" equal. If overlapping parts contain the same epoch with those two representatives, rowGroupKey differs and mergeResults emits both instead of the newer part replacing the older one. A focused probe gets comparator=0 and merged groups [a A]. Deduplication needs the same typed/collation equality as sorting, or DPC should fall back when it cannot provide that identity.

docs/developer/mysql-release-contract.md
```diff
@@ -0,0 +1,194 @@
...
+## Release compliance and approval
+
+`make check-third-party-licenses` regenerates the linked dependency notice tree
+and explicitly verifies the Vitess license used by the native protocol/parser.
+Release archives include that tree. Container images install the project
+`LICENSE`, `NOTICE`, and third-party tree under `/licenses`; RPMs install the
+same material under the platform license directory.
+
+Maintainer approval is recorded using
+[`mysql-release-approval.md`](mysql-release-approval.md). The checked-in file
```
houyuwushang: [P2] Add the required release approval record

This link currently has no target in the PR head: docs/developer contains mysql-benchmarks.md and mysql-release-contract.md, but no mysql-release-approval.md. Since this paragraph says the checked-in pending template is the release record, please add it or adjust the contract before merge.

houyuwushang (commented):
The five items from my previous review are addressed. I found two remaining correctness problems and one missing release artifact below.

docs/developer/mysql-release-contract.md
jranson: this document does not exist, removing reference.

pkg/backends/mysql/protocol.go
```diff
@@ -0,0 +1,1744 @@
...
+	// streaming API recognizes that packet but does not expose its affected-row,
+	// insert-ID, or session metadata, so dispatch from the parsed shape whenever
+	// it is available.
+	switch statement := parsed.statement.(type) {
+	case *vtparser.Select:
+		return statement.Into == nil
+	case *vtparser.Union:
+		return statement.Into == nil
+	}
+	switch parsed.statementType {
```
houyuwushang: [P1] Treat standalone `VALUES` as a row-producing statement

[MySQL 8.4 defines `VALUES`](https://dev.mysql.com/doc/refman/8.4/en/values.html) as a standalone statement that returns one or more rows. Vitess parses `VALUES ROW(1), ROW(2)` as `*sqlparser.ValuesStatement`, but `ASTToStatementType` leaves it `StmtUnknown`; this function consequently returns false and the query reaches `ExecuteFetchWithWarningCount(..., 1, ...)`, which fails with `Row count exceeded 1`.

This is the same response-shape edge as the CTE case. Please classify parsed table-value statements as reads, send them through the result-set path, and keep a successful `VALUES` from marking the session stateful. A two-row proxy regression should cover both.

houyuwushang (commented):
The three items from my last review are addressed. I found one remaining response-shape case in the new dispatch logic below.

pkg/backends/mysql/protocol.go
```diff
@@ -0,0 +1,1864 @@
...
+			return responseShapeUnsupported, "unclassified statements"
+		}
+	default:
+		return responseShapeOK, ""
+	}
+}
+
+func unsupportedTextFeature(query string) string {
+	trimmed := strings.TrimSpace(vtparser.StripLeadingComments(query))
+	if strings.HasPrefix(trimmed, "/*!") {
```
houyuwushang: [P1] Finish rejecting embedded executable comments

The cache guard fixed the earlier DPC case, but this check still only rejects a version comment when it starts the statement. Under [MySQL 8.4 executable-comment semantics](https://dev.mysql.com/doc/refman/8.4/en/comments.html), `SELECT 1 /*!80400 INTO @answer */` executes its `INTO` clause; Vitess ignores that 8.4-gated body here and classifies the query as rows. A lifecycle probe modeling the origin's OK packet returned an empty zero-metadata result downstream, and `SELECT /*!80400 @tenant := 1, */ 1` also left the session non-stateful. This is the hidden shape/state case the release contract says is rejected.

Please detect executable-comment tokens anywhere in the statement before dispatch, without treating `/*!` inside a quoted string as a comment, and cover both `80400` cases.

LimitlessEarth (approved):

PR #1128, opened 2026-09-03
refactor: move empty extent trimming to ExtentList

### LINKED ISSUE #1126: Move trimEmptyExtents logic to pkg/timeseries/extent_list.go
### Summary
Right now, in `deltaproxycache.go`, the function `fetchExtents` (lines 753–894) calls `trimEmptyExtents` to trim the extent list (i.e., deleting empty elements from the list). For performance reasons, we prefer to over-allocate a slice, so some elements inside the slice are empty while others contain elements.

This is information leakage: the deltaproxycache module should not care about modifying the internal state of the extent list. 

### Proposed Solution
The idea is to move the `trimEmptyExtents` function inside `pkg/timeseries/extent_list.go` and create a new method like this:
```Go
func (el ExtentList) TrimEmptyExtents() ExtentList {
       // Move existing trimEmptyExtents logic here
}
```
Once moved,  `trimEmptyExtents` function in deltaproxycache.go should be removed and replaced with calls to this new method.

### Benefits
1. Cleaner architecture
2. Better separation of responsibilities between modules

shiyudesu:
## Description

Move the empty extent filtering logic from the delta proxy cache implementation into `timeseries.ExtentList`.

This change:

- adds `ExtentList.TrimEmptyExtents()`;
- updates `fetchExtents` to use the new method;
- removes the standalone `trimEmptyExtents` function from `deltaproxycache.go`;
- adds table-driven unit tests covering interspersed empty extents, all-empty lists, empty lists, and lists without empty extents.

This keeps the existing behavior while making `ExtentList` responsible for managing its own representation.

Closes #1126

## Type of Change

- - [ ] Bug fix
- - [ ] New feature
- - [x] Optimization
- - [x] Test coverage
- - [ ] Documentation
- - [ ] Infrastructure

## Testing

- `make build`
- `make lint`
- `make test`
- `make data-race-test`

All checks pass. The project test suite reports 91.02% statement coverage, and `TrimEmptyExtents` has 100% statement coverage.

## AI Disclosure

### DIFF
--- a/pkg/timeseries/extent_list.go
+++ b/pkg/timeseries/extent_list.go
@@ -348,6 +348,20 @@ func (el ExtentList) Clone() ExtentList {
 	return c
 }
 
+// TrimEmptyExtents returns a copy of the ExtentList with empty Extents removed.
+func (el ExtentList) TrimEmptyExtents() ExtentList {
+	trimmed := make(ExtentList, len(el))
+	var cursor int
+	for _, extent := range el {
+		if extent == (Extent{}) {
+			continue
+		}
+		trimmed[cursor] = extent
+		cursor++
+	}
+	return trimmed[:cursor]
+}
+
 // CloneRange returns a perfect copy of the ExtentList, cloning only the
 // Extents in the provided index range (upper-bound exclusive)
 func (el ExtentList) CloneRange(start, end int) ExtentList {
--- a/pkg/timeseries/extent_list.go
+++ b/pkg/timeseries/extent_list.go
@@ -350,16 +350,14 @@ func (el ExtentList) Clone() ExtentList {
 
 // TrimEmptyExtents returns a copy of the ExtentList with empty Extents removed.
 func (el ExtentList) TrimEmptyExtents() ExtentList {
-	trimmed := make(ExtentList, len(el))
-	var cursor int
+	trimmed := make(ExtentList, 0, len(el))
 	for _, extent := range el {
 		if extent == (Extent{}) {
 			continue
 		}
-		trimmed[cursor] = extent
-		cursor++
+		trimmed = append(trimmed, extent)
 	}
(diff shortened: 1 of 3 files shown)


pkg/timeseries/extent_list.go
```diff
@@ -348,6 +348,20 @@ func (el ExtentList) Clone() ExtentList {
 	return c
 }
 
+// TrimEmptyExtents returns a copy of the ExtentList with empty Extents removed.
+func (el ExtentList) TrimEmptyExtents() ExtentList {
+	trimmed := make(ExtentList, len(el))
```
jranson: You can use `trimmed := make(ExtentList, 0, len(el))` here and then use `append` in the loop to populate the trimmed slice without declaring `cursor`. That is simpler and the same performance.

pkg/timeseries/extent_list.go
```diff
@@ -348,6 +348,20 @@ func (el ExtentList) Clone() ExtentList {
...
+	var cursor int
+	for _, extent := range el {
+		if extent == (Extent{}) {
```
jranson: It's better to retain the `emptyExtent := timeseries.Extent{}` declaration outside of the loop and use `emptyExtent` in the comparison here. Using a bare `Extent{}` inside of the loop causes an unnecessary re-allocation on every loop iteration.

jranson (changes requested):

pkg/timeseries/extent_list.go
shiyudesu: Good suggestion, thanks. I’ve updated `TrimEmptyExtents` to preallocate with `make(ExtentList, 0, len(el))` and populate it using `append`, removing the cursor while retaining a single allocation. The relevant tests and lint checks pass.

jranson (approved):

PR #1134, opened 2026-09-11
fix: TSM / Prom - parse extraneous paren wrappings in promql

### LINKED ISSUE #1127: TSM merge strategy changes when a PromQL aggregation is wrapped in redundant parentheses
## Summary

ALB Time Series Merge selects different merge strategies for two semantically equivalent PromQL expressions depending only on redundant leading parentheses.

```promql
count(up) or vector(0)
```

selects `StrategySum`, while:

```promql
(count(up)) or vector(0)
```

selects `StrategyDedup`.

Redundant parentheses should not change either the result or the TSM execution plan.

## Environment

- Trickster v2.0.5 (`5e0c69198367544013c41fe2f2c7e67ae8afd4df`)
- Backend provider: Prometheus/Mimir
- ALB mechanism: `tsm`
- Observed with both instant and range queries

The same relevant behavior also exists in the earlier `1530013d90fdcf1020fc50855e0cb3230d1fa54d` revision.

## Minimal reproduction

Compare the plans generated for:

```promql
count(up) or vector(0)
```

and:

```promql
(count(up)) or vector(0)
```

`PlanTSMMerge` currently produces:

```text
count(up) or vector(0)       -> StrategySum
(count(up)) or vector(0)     -> StrategyDedup
```

An end-to-end reproduction can use two or more disjoint TSM member backends and a selector that matches series in only one member.

Each backend evaluates the complete expression locally:

- A backend with matching series returns an unlabeled `{}` count.
- A backend without matches returns the unlabeled `{}` value `0` because of `or vector(0)`.

With `StrategyDedup`, these shard-local results are deduplicated or exposed as conflicting series instead of being summed into one federated count.

## Observed impact

In one range-query example:

- A direct Mimir datasource returned one stable series with value `8`.
- TSM queried 65 usable logical shards.
- Grafana received `65 × 241` range rows instead of one 241-point merged series.
- The graph displayed overlapping `0` and `8` values at the same timestamps.
- The instant result could return `0` even though the directly queried member returned `8`.

This can silently produce misleading dashboard and alert values.

## Expected behavior

Semantically equivalent expressions should use the same TSM plan and return the same result.

For disjoint federated shards, the count-plus-zero expression should produce one series containing the sum of the shard-local counts.

If a top-level binary expression such as `or` cannot be safely distributed, both forms should instead receive the same documented fallback, warning, or global-finalization behavior. Query formatting should never determine correctness.

## Suspected cause

`PlanTSMMerge` defaults to `StrategyDedup` and changes the strategy when `OuterAggregator` recognizes an aggregation operator:

- https://github.com/trickstercache/trickster/blob/5e0c69198367544013c41fe2f2c7e67ae8afd4df/pkg/backends/prometheus/tsm_provider.go#L59-L89
- https://github.com/trickstercache/trickster/blob/5e0c69198367544013c41fe2f2c7e67ae8afd4df/pkg/backends/prometheus/promql/outer_aggregator.go#L46-L74

`OuterAggregator` only examines the trimmed textual prefix. It recognizes `count`, but it does not unwrap a leading parenthesized expression, so `(count(...))` is not detected.

Because the complete expression is actually a binary `or`, the general fix should ideally inspect the parsed PromQL expression rather than only stripping leading parentheses.

## Suggested fix

- Parse or normalize the complete PromQL expression before choosing a TSM merge strategy.
- Ensure redundant grouping parentheses cannot change the selected plan.
- If a binary expression cannot be distributed safely, return a warning or use a global finalizer consistently for every equivalent formatting.

## Current workarounds

Remove the redundant wrapper:

```promql
count(up) or vector(0)
```

Or make the intended cross-shard reduction an explicit complete outer expression:

```promql
sum(count(up) or vector(0))
```

jranson:
## Description

this fixes #1127 by swapping out trickster's many prom parsing conditionals for the official PromQL parser, which naturally resolves the extra parenthesis wrapper before selecting the merge strategy. Likely fixes other similar unreported issues about which we'd be playing whack-a-mole to fix ad infinitum.

## Type of Change

- - [x] Bug fix
- - [ ] New feature
- - [ ] Optimization
- - [x] Test coverage
- - [ ] Documentation
- - [ ] Infrastructure

## AI Disclosure


houyuwushang (commented):
I found one regression in the sort-wrapped zero-fallback case.

pkg/backends/prometheus/tsm_provider.go
```diff
@@ -488,23 +511,23 @@ func weightedAveragePlan(r *http.Request, originalQuery, fanoutQuery string,
...
-		return spec.InnerQuery, true
+func tsmInnerQuery(expr promql.Expr) (promql.Expr, bool) {
+	if spec, ok := promql.ParseRankAggregation(expr); ok {
+		return spec.Inner, true
 	}
-	if spec, ok := promql.ParseSortWrapper(query); ok {
-		if _, found := promql.OuterAggregator(spec.InnerQuery); found {
-			return spec.InnerQuery, true
+	if spec, ok := promql.ParseSortWrapper(expr); ok {
+		if _, _, found := promql.CompleteOuterAggregation(spec.Inner); found {
```
houyuwushang: Could we unwrap `sort`/`sort_desc` when the input is a supported zero fallback too? `CompleteOuterAggregation` rejects `count(up) or vector(0)` because its root is a binary expression, so `sort(count(up) or vector(0))` never reaches `zeroFallbackMergesBySum` and falls back to dedup instead of sum.

With shard counts 0, 3, and 5, a handler-level test returns 8 on the parent, but this head returns 0 for an instant query and 5 for a range query. The unsorted form still returns 8. The corresponding guard in `FinalizeTSMMerge` also needs to recognize this shape so grouped results retain global ordering.

houyuwushang (commented):
The sort-wrapped zero-fallback case I raised is fixed: it keeps the sum strategy, and grouped instant results retain global ordering. I rechecked `387c3ac` and have no further concerns with the PromQL changes.

LimitlessEarth (approved):

